PR #40, opened 2026-02-05
[0.x] Add Mistral AI support

jackbayliss:
This adds support for https://mistral.ai/ 

Which Prism supports 🖖🏻 

I think I've covered everything?


taylorotwell:
Doesn't Mistral support audio? Do we want to add that?

jackbayliss:
All sorted.. it supports STT (transcription) , not TTS afaik.  This was quite interesting to dig into!

I think I've got it covered, sorry If I've missed anything.

PR #48, opened 2026-02-06
Add Chutes AI provider

rezaulhreza:
Add Chutes AI provider with text, image, streaming, audio, transcription, and embedding support.

[Chutes](https://chutes.ai) ([docs](https://chutes.ai/docs), [pricing](https://chutes.ai/pricing)) gives access to open-source models like DeepSeek, Qwen, Gemma, Llama, and Kimi and more through an OpenAI-compatible API. This PR is not affiliated with or sponsored by Chutes.

## Configuration

Follows the same pattern as every other provider .  just `driver` + `key`:

```php
'chutes' => [
    'driver' => 'chutes',
    'key' => env('CHUTES_API_KEY'),
],
```

URLs are hardcoded in gateway classes, models in the provider class. Users override models at call time via the `model:` parameter.

## Text & Streaming

Routes through Prism's Groq handler since both use `/chat/completions`. Prism's OpenAI handler targets `/responses` which 404s against Chutes, and OpenRouter has a null-handling bug in `ToolCallMap`.

## Image Generation

Custom `ChutesImageGateway` since `image.chutes.ai` returns raw JPEG bytes rather than the OpenAI JSON format.

## Audio (TTS)

Kokoro via `ChutesAudioGateway`, POST to the TTS endpoint returns raw WAV. Maps `default-male` to `am_adam` and `default-female` to `af_heart` (same pattern as ElevenLabs/PrismGateway). Users can also pass any Kokoro voice ID directly.

## Transcription (STT)

Whisper large-v3 via the same `ChutesAudioGateway`, POST with base64-encoded audio, returns timestamped segments.

## Embeddings

Qwen3-Embedding-0.6B via `ChutesEmbeddingGateway`. The embedding endpoint lives on a separate subdomain, so it gets its own gateway.

## Bug Fix: Null usage in stream end event

Discovered while testing Chutes streaming - `PrismStreamEvent::toStreamEndEvent()` passes `$event->usage` directly to `PrismUsage::toLaravelUsage()`, which expects a non-null `Usage` object. Chutes doesn't include usage stats in streaming responses, so this throws a `TypeError`. The fix falls back to an empty `Usage` instance when `$event->usage` is null. This isn't Chutes-specific - any provider that omits usage from stream events would hit the same crash.

## Tests

24 unit tests + 7 integration tests covering text, ad-hoc agent, streaming, image, audio, transcription, and embeddings.

```
CHUTES_API_KEY=key vendor/bin/phpunit tests/Feature/ChutesProviderTest.php
CHUTES_API_KEY=key vendor/bin/phpunit tests/Feature/ChutesIntegrationTest.php
```

### DIFF
--- a/config/ai.php
+++ b/config/ai.php
@@ -60,21 +60,6 @@ return [
         'chutes' => [
             'driver' => 'chutes',
             'key' => env('CHUTES_API_KEY'),
-            'url' => env('CHUTES_BASE_URL', 'https://llm.chutes.ai/v1'),
-            'image_url' => env('CHUTES_IMAGE_URL', 'https://image.chutes.ai/generate'),
-            'tts_url' => env('CHUTES_TTS_URL', 'https://chutes-kokoro.chutes.ai/speak'),
-            'stt_url' => env('CHUTES_STT_URL', 'https://chutes-whisper-large-v3.chutes.ai/transcribe'),
-            'embedding_url' => env('CHUTES_EMBEDDING_URL', 'https://chutes-qwen-qwen3-embedding-0-6b.chutes.ai/v1/embeddings'),
-            'models' => [
-                'default' => env('CHUTES_MODEL', 'deepseek-ai/DeepSeek-V3'),
-                'cheapest' => env('CHUTES_CHEAPEST_MODEL', 'unsloth/gemma-3-4b-it'),
-                'smartest' => env('CHUTES_SMARTEST_MODEL', 'moonshotai/Kimi-K2.5'),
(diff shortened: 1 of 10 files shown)


taylorotwell:
There is way too much configuration compared to the other providers. It needs to be simpler. 👍

rezaulhreza:
@taylorotwell any luck with this?

rezaulhreza:
@taylorotwell is it something that can be merged? if not please close the PR. Understand this may not be the priority, however a decision would be nice to hear.

PR #50, opened 2026-02-06
 feat: add custom base URL support for AI providers

sadhakbj:


sadhakbj:
Sorry i was creating branch from main instead i should have done from 0.x i will fix this

PR #72, opened 2026-02-06
[0.x] Use `str_contains` instead of `Str::contains`

lucasmichot:
Use `str_contains` instead of `Str::contains`


gkmk:
Why?

lucasmichot:
> Why?

Why using `Str::contains`?

gkmk:
- Its Laravel flavor and it exists
- Its well documented
- Has opportunity to integrate better with Laravel specific future features

Personally i would use Laravel functions first even though they exist in vanila PHP because code style and readability matters the most i think.

PR #83, opened 2026-02-06
fix: remove invalid $event->delta access on ThinkingCompleteEvent

alanost:
The `ThinkingComplete` case in `PrismStreamEvent::toLaravelStreamEvent()` passes `$event->delta` to the `ReasoningEnd` constructor, but Prism's `ThinkingCompleteEvent` does not have a `$delta` property.

This causes an `Undefined property: Prism\Prism\Streaming\Events\ThinkingCompleteEvent::$delta` error when streaming models that emit thinking tokens (e.g. `qwen/qwen3-8b` via OpenRouter).

Additionally, `$event->delta` was passed as the third argument where `ReasoningEnd` expects `int $timestamp`, causing a type mismatch.

**Fix:** Remove `$event->delta` from the constructor call to match the `ReasoningEnd(id, reasoningId, timestamp, summary)` signature.


jkbennemann:
nice! Did not notice your PR, but created one with a similar fix! I've also tested a couple of different models to verify :) 
Hope this will be merged 🙏

https://github.com/laravel/ai/pull/88

PR #89, opened 2026-02-07
Add Agent Skills support (discovery, activation, resource reading)

truthanb:
Skills let you give your agent specialized knowledge using plain markdown files. A customer support agent could activate a skill that knows your escalation policies. A product agent could pull in feature documentation on demand. The format is flexible enough to describe anything your agent might need to know, without stuffing it all into the system prompt upfront.

Adds support for the [Agent Skills](https://agentskills.io) open standard, enabling agents to discover, activate, and read skill resources through a three-layer progressive disclosure model.

**Discovery**: `SkillLoader` parses `SKILL.md` files (YAML frontmatter + markdown body). `SkillRegistry` holds loaded skills and generates an `<available_skills>` XML block for the system prompt containing only names and descriptions.

**Activation**: `ActivateSkillTool` implements the `Tool` contract so the LLM can call `activate_skill` to retrieve a skill's full instructions on demand. If the skill has files in `scripts/`, `references/`, or `assets/`, the response includes an `<available_resources>` listing.

**Resource reading**: `ReadSkillResourceTool` lets the LLM read individual resource files from an activated skill by relative path, with path traversal protection.

All frontmatter fields from the spec are supported: `name`, `description`, `license`, `compatibility`, `allowed-tools`, and `metadata`.

**Note:** Test fixtures in `tests/Feature/Skills/Fixtures/` contain `SKILL.md` files. These are test data only.

image

```
<?php
$registry = SkillRegistry::fromDirectory(base_path('skills'));

// In your agent's instructions()
$registry->toPrompt();

// In your agent's tools()
new ActivateSkillTool($registry);
new ReadSkillResourceTool($registry);
```

### DIFF
--- a/src/Concerns/LoadsSkills.php
+++ b/src/Concerns/LoadsSkills.php
@@ -1,0 +1,111 @@
+<?php
+
+namespace Laravel\Ai\Concerns;
+
+use Illuminate\Support\Facades\Cache;
+use Laravel\Ai\Skills\Skill;
+use Laravel\Ai\Skills\SkillLoader;
+
+trait LoadsSkills
+{
+    /**
+     * The resolved skills (memoized per-instance).
+     *
+     * @var array<Skill>|null
+     */
+    protected ?array $resolvedSkills = null;
+
+    /**
+     * Get the skills available to the agent.
+     *
+     * By default, loads skills from `resource_path('ai/skills')` with
+     * automatic caching. Override this method to load from any source
+     * (database, S3, API, etc.).
+     *
+     * @return array<Skill>
+     */
+    public function skills(): array
+    {
+        if ($this->resolvedSkills !== null) {
+            return $this->resolvedSkills;
+        }
+
+        $paths = $this->skillPaths();
+
+        if (empty($paths)) {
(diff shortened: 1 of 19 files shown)


pushpak1300:
Thanks for sending PR. I'm not sure if @taylorotwell will like merge this but as some improvements you can do is. 
- create Interface on Agent called SupportSkills which will add skills method. 
- Ideally if interface has been added the skills should be automatically passes also there is very less chance user will interact with local skills it should be actually build on filesystem and skills method should load the skills form filesystem either local/s3 or DB (string) basically it should return array of skills which can be dynamically fetch. we could go one step and automatically cache if cache is setup. 

I'm just sure that a good dev ex can be created for this.

truthanb:
@pushpak1300 
> I'm just sure that a good dev ex can be created for this.

Thanks, I wanted to keep the first request pretty light and open, but I hear you. I've tried to make some changes that I hope are a bit closer to better for dev ex.

truthanb:
https://packagist.org/packages/truthanb/laravel-ai-skills

PR #102, opened 2026-02-08
Add cache response middleware for text generation

SanderMuller:
This PR adds a `CacheResponse` middleware that caches `AgentResponse` and `StructuredAgentResponse` 
  objects, avoiding redundant API calls for identical prompts. Similar to the existing embeddings     
  caching, but implemented as opt-in middleware per agent following the existing middleware pattern.  
                                                                                                      
  ```php                                                                                              
  class MyAgent implements Agent, HasMiddleware                                                       
  {
      use Promptable;

      public function middleware(): array
      {
          return [
              new CacheResponse(seconds: 3600, store: 'redis'),
          ];
      }
  }
  ```

  Cache keys are deterministic SHA256 hashes of the agent class, provider, model, instructions, and
  prompt text. Streaming responses pass through uncached. Cached responses are restored with a fresh
  invocation ID.

  This PR also includes the fix from #100 as it is required for short-circuiting middleware to work
  correctly.

### DIFF
--- a/src/Middleware/CacheResponse.php
+++ b/src/Middleware/CacheResponse.php
@@ -1,0 +1,124 @@
+<?php
+
+namespace Laravel\Ai\Middleware;
+
+use Closure;
+use Illuminate\Support\Facades\Cache;
+use Illuminate\Support\Str;
+use Laravel\Ai\Prompts\AgentPrompt;
+use Laravel\Ai\Responses\AgentResponse;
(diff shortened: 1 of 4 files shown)


taylorotwell:
Something that might be tricky here are attachments. You probably wouldn't want to cache those responses if the attachment content isn't also baked into the key.

I kinda think I may just leave this to be distributed as a package for now though. Thanks!

PR #107, opened 2026-02-09
Fix: Read timeout from provider config as fallback

### LINKED ISSUE #104: Provider config timeout key is silently ignored
Setting timeout in provider config has no effect. The SDK never reads it — no warning, no error, just silent ignore.
```
config(['ai.providers.anthropic' => [
    'driver' => 'anthropic',
    'key' => $apiKey,
    'url' => $url,
    'timeout' => 180, // ← does nothing
]]);

```

The actual timeout resolution in Promptable::getTimeout() is:

$timeout argument passed to prompt()
timeout() method on agent class
#[Timeout] attribute on agent class
Hardcoded return 60

Provider config is never checked.
Steps to Reproduce

Set timeout in config/ai.php provider config:

```
'anthropic' => [
    'driver' => 'anthropic',
    'key' => env('ANTHROPIC_API_KEY'),
    'timeout' => 180,
],

```
Create an agent without #[Timeout] attribute or timeout() method
Send a prompt that takes longer than 60 seconds
Request times out at 60s despite the config value

Expected Behavior
Either:

Provider config timeout should be read as a fallback (before the hardcoded 60), or
A deprecation warning / exception should be thrown if timeout key is present in provider config, so the developer knows it's not being used

Context
Discovered in production — batch processing 20 articles with translation needed ~180s. Got the default 60s timeout instead. The config key timeout sits right next to key and url which both work, so it's natural to assume it works too.

delabon:
## Problem
Setting `timeout` in provider config has no effect. Example:
```php
'anthropic' => [
    'driver' => 'anthropic',
    'key' => env('ANTHROPIC_API_KEY'),
    'timeout' => 180, // silently ignored
],
```

The timeout always defaults to 60 seconds regardless of this config value.

## Root Cause
The `getTimeout()` method never checked the provider config. It only checked:
1. Explicit parameter
2. `timeout()` method
3. `#[Timeout]` attribute
4. Then jumped straight to hardcoded `60`

## Solution
Added provider config as a fallback before the hardcoded default.

Now checks `config("ai.providers.{provider}.timeout")` and uses it if present.

## Testing
Added two tests confirming provider config timeout works for both `prompt()` and `stream()` methods.

Fixes #104

### DIFF
--- a/src/Promptable.php
+++ b/src/Promptable.php
@@ -224,7 +224,35 @@ trait Promptable
             return $attributes[0]->newInstance()->value;
         }
 
+        $providerName = $this->getDefaultProvider();
+
+        if ($providerName) {
+            $providerTimeout = config("ai.providers.{$providerName}.timeout");
+
+            if (! is_null($providerTimeout)) {
+                return (int) $providerTimeout;
+            }
+        }
+
         return 60;
+    }
+
+    /**
+     * Get the default provider name for this agent.
+     */
+    protected function getDefaultProvider(): ?string
+    {
+        if (method_exists($this, 'provider')) {
+            return $this->provider();
+        }
+
+        $attributes = (new ReflectionClass($this))->getAttributes(ProviderAttribute::class);
+
+        if (! empty($attributes)) {
+            return $attributes[0]->newInstance()->value;
+        }
+
+        return config('ai.default');
     }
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
I don't think we ever document this as an option?

PR #109, opened 2026-02-09
refactor: AiManager.php eliminate instanceof checks

mubbi:
Extract resolveProvider helper to eliminate repeated instanceof checks in AiManager.


taylorotwell:
Fine as is I think.

mubbi:
> Fine as is I think.

thats a code smell, DRY violation. :(

PR #111, opened 2026-02-09
[0.x] Add `agent()` method to the `\Laravel\Ai\Tools\Request` class

gdebrauwer:
I wanted to create a tool that returns data that is limited to what the user of the current conversation can see. Currently, there is no way to do that as the tool does not have access to the agent object.

This PR fixes that by adding an `agent()` method to the `\Laravel\Ai\Tools\Request` class. I chose to give the request access to the entire agent object and not just passing the value of the `conversationUser` property on the `RemembersConversations` trait, because an agent might have other properties that you want to access in a tool.

```php
class GetBookingsByStatus implements Tool
{
    {
		return Booking::query()
			->where('status', $request->string('status'))
			->where('user_id', $request->agent()->conversationParticipant()->id)
			->get()
			->toJson();
	}
}
```


taylorotwell:
I tried to avoid making the gateway knowledgable of any "agent" concept. I wonder if you could do:

```php
public function tools()
{
    return [new MyTool($this)];
}
```

PR #118, opened 2026-02-09
Fix streaming generators under Octane

csfh:
Hello,

This PR fixes generator-returning stream callbacks under Octane being ignored, which results in an empty response body. In Laravel’s `Illuminate\Routing\ResponseFactory::stream()`, when `$_SERVER['LARAVEL_OCTANE']` is set and the callback is a generator, the callback is attached directly to `StreamedResponse` without a wrapper that iterates it. Octane doesn’t iterate the generator, so no chunks are echoed. Our fix in `StreamableAgentResponse::toResponse()` and `CanStreamUsingVercelProtocol::toVercelProtocolResponse()` consumes any iterable returned by the stream callback and echoes each chunk (with flushing), so streaming works consistently under both Octane and non-Octane runtimes. It also exits early on `connection_aborted()` to avoid unnecessary work. This mirrors the sibling PR https://github.com/laravel/mcp/pull/148.

Here’s the Octane-specific branch in laravel/framework that triggers the issue (from `src/Illuminate/Routing/ResponseFactory.php`, method stream()):

```php
if (! is_null($callback) && (new ReflectionFunction($callback))->isGenerator()) {
    if (isset($_SERVER['LARAVEL_OCTANE'])) {
        return (new StreamedResponse(
            null, $status, array_merge($headers, ['X-Accel-Buffering' => 'no'])
        ))->setCallback($callback);
    }
    return new StreamedResponse(function () use ($callback) {
        foreach ($callback() as $chunk) {
            echo $chunk;
            when(ob_get_level() > 0, fn () => ob_flush());
            flush();
        }
    }, $status, array_merge($headers, ['X-Accel-Buffering' => 'no']));
}
```

I've also included my own trait I wrote to resolve this issue in my own project to resolve this issue.

**User benefits**

- Streaming responses work under Octane when callbacks return generators/iterables.
- Behavior matches non-Octane runtimes without changing user-facing APIs.
- Applies to both the default SSE stream and the Vercel protocol stream.

**Safety**
- Only affects stream callbacks that return an iterable/generator.
- Callbacks that echo directly are unchanged; they are not wrapped or double-emitted (unless a callback both echoes and returns an iterable, which is undefined usage).

**Tests**
- Added coverage for generator return values when `$_SERVER['LARAVEL_OCTANE']` is set (default SSE).
- Added coverage for generator return values when `$_SERVER['LARAVEL_OCTANE']` is set (Vercel protocol).
- `./vendor/bin/phpunit --filter StreamableAgentResponseStreamingTest` (filtered because I think a lot of test requires a token for a service provider?)

**Other options**

We could change laravel/octane to iterate generator callbacks in `ResponseFactory::stream()`, but that’s a broader framework behavior change. This PR keeps the fix local to this transport.

**AI**

I used GPT-Codex to triage this issue and work on a fix.

```php
<?php

declare(strict_types=1);

namespace App\Concerns;

use Laravel\Ai\Responses\StreamableAgentResponse;
use Symfony\Component\HttpFoundation\StreamedResponse;

/**
 * Converts a StreamableAgentResponse into an echo-based StreamedResponse.
 *
 * StreamableAgentResponse::toResponse() produces a generator-based callback
 * that is never iterated under Octane, resulting in an empty response body.
 * This trait manually iterates the stream and echoes each SSE event, which
 * works correctly under both Octane and non-Octane environments.
 */
trait StreamsAgentResponses
{
    protected function toStreamedResponse(StreamableAgentResponse $stream): StreamedResponse
    {
        return response()->stream(function () use ($stream): void {
            foreach ($stream as $event) {
                echo 'data: '.($event)."\n\n";
                if (ob_get_level() > 0) {
                    ob_flush();
                }
                flush();
            }
            echo "data: [DONE]\n\n";
            if (ob_get_level() > 0) {
                ob_flush();
            }
            flush();
        }, 200, [
            'Content-Type' => 'text/event-stream',
        ]);
    }
}
```

### DIFF
--- a/src/Responses/Concerns/CanStreamUsingVercelProtocol.php
+++ b/src/Responses/Concerns/CanStreamUsingVercelProtocol.php
@@ -26,7 +26,7 @@ trait CanStreamUsingVercelProtocol
             public ?array $lastStreamEndEvent = null;
         };
 
-        return response()->stream(function () use ($state) {
+        $stream = function () use ($state): iterable {
             $lastStreamEndEvent = null;
 
             foreach ($this as $event) {
@@ -69,6 +69,28 @@ trait CanStreamUsingVercelProtocol
             }
 
             yield "data: [DONE]\n\n";
+        };
+
+        return response()->stream(function () use ($stream): void {
+            $result = $stream();
+
+            if (! is_iterable($result)) {
+                return;
+            }
+
+            foreach ($result as $message) {
+                if (connection_aborted()) {
+                    return;
+                }
+
+                echo (string) $message;
+
+                if (ob_get_level() > 0) {
+                    ob_flush();
+                }
+
+                flush();
+            }
         }, headers: [
             'Cache-Control' => 'no-cache, no-transform',
             'Content-Type' => 'text/event-stream',
--- a/src/Responses/StreamableAgentResponse.php
+++ b/src/Responses/StreamableAgentResponse.php
@@ -103,12 +103,34 @@ class StreamableAgentResponse implements IteratorAggregate, Responsable
             return $this->toVercelProtocolResponse();
         }
 
-        return response()->stream(function () {
+        $stream = function (): iterable {
             foreach ($this as $event) {
                 yield 'data: '.((string) $event)."\n\n";
             }
 
             yield "data: [DONE]\n\n";
+        };
+
+        return response()->stream(function () use ($stream): void {
+            $result = $stream();
+
+            if (! is_iterable($result)) {
+                return;
+            }
+
+            foreach ($result as $message) {
+                if (connection_aborted()) {
+                    return;
+                }
+
+                echo (string) $message;
+
+                if (ob_get_level() > 0) {
+                    ob_flush();
+                }
+
+                flush();
+            }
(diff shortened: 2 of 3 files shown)


taylorotwell:
Broke streaming for me.

csfh:
@taylorotwell how are you returning the streaming response? I have this fix live, frankenphp, laravel cloud right now and working.

I haven't updated dependencies in a while, maybe something else recently fixed it, if this is now broken. Either way I'd love to reproduce it failing. I will chime back in here if I find out anything.

```
trait StreamsAgentResponses
{
    protected function toStreamedResponse(StreamableAgentResponse $stream): StreamedResponse
    {
        return response()->stream(function () use ($stream): void {
            foreach ($stream as $event) {
                echo 'data: '.($event)."\n\n";

                if (ob_get_level() > 0) {
                    ob_flush();
                }

                flush();
            }

            echo "data: [DONE]\n\n";

            if (ob_get_level() > 0) {
                ob_flush();
            }

            flush();
        }, 200, [
            'Content-Type' => 'text/event-stream',
        ]);
    }
}
```

Issue #76, opened 2026-02-06
[Feature] Anonymous conversations support


miktown:
From what we can see, a `user_id` is required and non-nullable for both `agent_conversations` and `agent_conversation_messages`.

This makes it hard to scale use cases with anonymous users, such as a public chatbot.

In my case, I’m currently creating an “anonymous user” and, once the user identifies themselves, I transfer the conversation to the real user. However, if the user never identifies, I still need to persist the conversation in order to keep long, multi-step context.

Is there a recommended approach for handling anonymous conversations, or plans to support conversations without a required user_id?

BoweFrankema:
I did the same with linking it the conversation to a session and then after login/registration linking it to a user. But it's not idea.

giuliodelmastro:
@BoweFrankema do you think it's not a good idea?
> I did the same with linking it the conversation to a session and then after login/registration linking it to a user. But it's not idea.

BoweFrankema:
Yes I think it is.. I just confirmed what you flagged as an issue... it
should be easier :-)

pushpak1300:
Closing this because of #122

PR #122, opened 2026-02-09
Add support for anonymous users

bensherred:
This adds support for anonymous users by allowing the `userId` to be null.

`latestConversationId` still requries a user ID as it stands, but if you feel this should also allow `null` as a value, then I'd be happy to make the change.

As it stands, I've also left the migration as requiring a `user_id`, but if you think it would be best to update/add a new migration, then I can make that change as well (we will potentially want to make this change as the `RemembersConversations` trait allows `conversationParticipant` to return `null`)

#76

### DIFF
--- a/src/Contracts/ConversationStore.php
+++ b/src/Contracts/ConversationStore.php
@@ -16,17 +16,17 @@ interface ConversationStore
     /**
      * Store a new conversation and return its ID.
      */
-    public function storeConversation(string|int $userId, string $title): string;
+    public function storeConversation(string|int|null $userId, string $title): string;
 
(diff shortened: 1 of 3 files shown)


giuliodelmastro:
@bensherred In my opinion, support for null user_id is a temporary solution, as it limits the conversation history. In my specific case, for example, I have a conversation-session association, and only if there is actually a logged-in user are these associated with the user. This way, I maintain support for the history even for users who are not logged in.
I suggest implementing separate interfaces to manage the conversation history and perhaps a polymorphic relationship between conversation, session, and user.

PR #126, opened 2026-02-10
Fix undefined array key error in queued agent assertions

mubbi:
Both assertAgentWasQueued and assertAgentNotQueued accessed $this->recordedQueuedPrompts[$agent] without checking if the key exists, causing an "Undefined array key" error when the agent was never queued. 

This PR adds the ?? [] fallback to match the pattern already used by the other assertion methods.

### DIFF
--- a/src/Concerns/InteractsWithFakeAgents.php
+++ b/src/Concerns/InteractsWithFakeAgents.php
@@ -105,7 +105,7 @@ trait InteractsWithFakeAgents
         return $this->assertAgentWasPrompted(
             $agent,
             $callback,
-            $this->recordedQueuedPrompts[$agent],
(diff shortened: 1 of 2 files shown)


src/Concerns/InteractsWithFakeAgents.php
```diff
@@ -105,7 +105,7 @@ public function assertAgentWasQueued(string $agent, Closure|string $callback): s
         return $this->assertAgentWasPrompted(
             $agent,
             $callback,
-            $this->recordedQueuedPrompts[$agent],
+            $this->recordedQueuedPrompts[$agent] ?? [],
```
BouchantoufZakaria: why not using the Arr::get() function ?

PR #137, opened 2026-02-11
Add Traceable trait with log and Langfuse tracing drivers

tomphilpotts:
I've been building with Laravel AI and one thing I kept running into is that there's no way to see what your agents are actually doing. When something goes wrong or even when things go right; I'm flying blind. There's no observability into the prompts being sent, the tools being called, or the tokens being used.

This PR adds a Traceable trait that gives you that visibility, with support for logging to a Laravel log channel or sending traces to Langfuse or others. 

Maybe you will find it useful.

## Summary

- Adds an opt-in `Traceable` trait that agents can use (like `RemembersConversations`) to send trace data to a configured backend during invocations
- Includes two drivers: **Log** (writes structured entries to a Laravel log channel) and **Langfuse** (dispatches queued jobs to the Langfuse batch ingestion API)
- Includes a **Null** driver for testing or explicitly disabling tracing
- Agents can override the default driver via `tracingDriver()` method

### How it works

When `GeneratesText::prompt()` or `StreamsText::stream()` detects the `Traceable` trait (via `class_uses_recursive`, same pattern as `RemembersConversations`), it registers scoped event listeners for that invocation's ID. These listeners collect data from `PromptingAgent`, `InvokingTool`, `ToolInvoked`, `AgentPrompted`, `AgentStreamed`, and `AgentFailedOver` events.

### New files

| File | Purpose |
|------|---------|
| `src/Concerns/Traceable.php` | Marker trait with optional `tracingDriver()` override |
| `src/Contracts/TracingDriver.php` | Driver interface |
| `src/Tracing/TracingManager.php` | Driver factory (singleton) |
| `src/Tracing/Drivers/LogDriver.php` | Log channel backend |
| `src/Tracing/Drivers/LangfuseDriver.php` | Langfuse backend |
| `src/Tracing/Drivers/NullDriver.php` | No-op driver |
| `src/Tracing/LangfusePayloadBuilder.php` | Builds Langfuse batch ingestion payload |
| `src/Tracing/Jobs/SendLangfuseTrace.php` | Queued job for Langfuse API |

### Usage

```php
class MyAgent implements Agent
{
    use Promptable, Traceable;

    public function instructions(): string { return '...'; }

    // Optional: override the default tracing driver
    public function tracingDriver(): ?string { return 'langfuse'; }
}
```

### Configuration

```php
// config/ai.php
'tracing' => [
    'default' => env('AI_TRACING_DRIVER', 'log'),
    'drivers' => [
        'log' => [
            'driver' => 'log',
            'channel' => env('AI_TRACING_LOG_CHANNEL'),
            'level' => 'info',
        ],
        'langfuse' => [
            'driver' => 'langfuse',
            'url' => env('LANGFUSE_URL', 'https://cloud.langfuse.com'),
            'public_key' => env('LANGFUSE_PUBLIC_KEY'),
            'secret_key' => env('LANGFUSE_SECRET_KEY'),
            'queue' => env('AI_TRACING_QUEUE', 'default'),
        ],
    ],
],
```

## Test plan

- [x] 16 unit tests covering TracingManager, all drivers, trait detection, and driver override
- [x] 3 integration tests with real Anthropic API calls verifying Langfuse payload structure and end-to-end trace delivery
- [x] Verified agents without `Traceable` trait are completely unaffected
- [x] All 112 existing tests continue to pass

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 /tmp
 /vendor
 CLAUDE.md
+.env
--- a/config/ai.php
+++ b/config/ai.php
@@ -40,6 +40,37 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Tracing
+    |--------------------------------------------------------------------------
+    |
+    | Below you may configure tracing for AI agent invocations. When an
+    | agent uses the Traceable trait, trace data will be sent to the
+    | configured driver. The Langfuse driver dispatches queued jobs.
+    |
+    */
+
+    'tracing' => [
+        'default' => env('AI_TRACING_DRIVER', 'log'),
+
+        'drivers' => [
+            'log' => [
+                'driver' => 'log',
+                'channel' => env('AI_TRACING_LOG_CHANNEL'),
+                'level' => 'info',
+            ],
+
+            'langfuse' => [
+                'driver' => 'langfuse',
+                'url' => env('LANGFUSE_URL', 'https://cloud.langfuse.com'),
+                'public_key' => env('LANGFUSE_PUBLIC_KEY'),
+                'secret_key' => env('LANGFUSE_SECRET_KEY'),
+                'queue' => env('AI_TRACING_QUEUE', 'default'),
+            ],
+        ],
+    ],
+
+    /*
+    |--------------------------------------------------------------------------
     | AI Providers
     |--------------------------------------------------------------------------
     |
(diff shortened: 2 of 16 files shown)


taylorotwell:
I would rather this be a package for now. We will be bringing better first-party observability very soon though.

bbredewold:
> I would rather this be a package for now. We will be bringing better first-party observability very soon though.

Soon™

PR #138, opened 2026-02-11
feat(prompts): compose with Blade

mattmcdev:
This PR proposes letting you compose your prompts with Blade templates for more complex prompting, particularly where you're blending data into the prompt. 

If this is wanted I'll flesh out the PR properly with tests, etc.

As written this would be breaking as it updates the `prompt()` signature to accept an array of `promptData` as the second argument.

### DIFF
--- a/src/Promptable.php
+++ b/src/Promptable.php
@@ -45,6 +45,7 @@ trait Promptable
      */
     public function prompt(
         string $prompt,
+        array $promptData = [],
         array $attachments = [],
         array|string|null $provider = null,
         ?string $model = null,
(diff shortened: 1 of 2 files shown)


mattmcdev:
Another option - to make it a non-breaking change - I could do it as a trait: `HasComposablePrompts`

taylorotwell:
You can already just return a view from `instructions`.

mattmcdev:
@taylorotwell You can, but the instructions are a system prompt, and this assumes all user messages are going to be simple. 

e.g. if you have an agent with instructions
```You are a marketing agent replying to support requests.```

And then you want to prompt it.
```
A new support request has come in for {businessName}.

This support request is from {customerName}.

[if (hasRecentSupportMessages)]
They have {recentSupportMessagesCount} recent messages that might be relevant.

Recent support message history:
[foreach (recentSupportMessages)]
    {recentSupportMessage}
[/endforeach]
[/endif]

{hasRecentSupportMessages ? 'Their most recent support request:' : 'Their support request:'}
{supportRequest}
```

PR #141, opened 2026-02-12
Allow customization of conversation title prompt

andi-xd:
The `RememberConversation` middleware has the title generation prompt hardcoded in English (line 75), so there's no way to customize it. Not through config, not through method override.

If the app runs in another language, the only option right now is to work around the SDK and handle title generation separately, which kind of defeats the purpose of `RemembersConversations`.

This PR adds a `conversationTitlePrompt()` method to the `RemembersConversations` trait so developers can override it per agent, the same way `maxConversationMessages()` already works. Since it lives in the agent, each one can have its own prompt independently.
```php
public function conversationTitlePrompt(): string
{
    return 'Your custom prompt in any language...';
}
```

The middleware checks for the method with `method_exists()` and falls back to the current default, so nothing breaks for existing implementations.

### DIFF
--- a/src/Concerns/RemembersConversations.php
+++ b/src/Concerns/RemembersConversations.php
@@ -69,6 +69,14 @@ trait RemembersConversations
     }
 
     /**
+     * Get the prompt used to generate conversation titles.
+     */
+    public function conversationTitlePrompt(): string
+    {
+        return 'Generate a concise 3-5 word title for a conversation that starts with the following message. Respond with only the title, no quotes or punctuation.';
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


taylorotwell:
I will be working on this directly over the next few weeks so I'm adding this to my list to revisit.

helitik:
Hi @taylorotwell, as of v0.10.1 and master, the title prompt is still hardcoded in English with no override hook. Still on your list? Happy to rework the PR if useful.

pushpak1300:
Hey @helitik i will take a look on this.

PR #142, opened 2026-02-12
Add string representation to Provider to prevent fatal error in unsupported provider path

### LINKED ISSUE #112: Provider attribute is case-sensitive and results in confusing error message
The `#[Provider(...)]` attribute currently appears to be case-sensitive.
If the provider name is written with different casing, for example `OpenAI` instead of `openai`, the SDK throws an exception that is difficult to understand and does not indicate that the problem is casing.

### Example
`#[Provider('OpenAI')]`

This results in the following error:
```
Error
[vendor/laravel/ai/src/Gateway/Prism/PrismGateway.php:359]
Object of class Laravel\Ai\Providers\OpenAiProvider could not be converted to string
```

### Expected behavior

One of the following would be much clearer:

1. Provider names are treated case-insensitively, so OpenAI, openai, or OPENAI all work.
2. A descriptive exception is thrown, e.g.: Unknown provider name [OpenAI]

boynet:
Fixes  https://github.com/laravel/ai/issues/112

Prevents a fatal error when an unsupported provider is encountered (for example `OpenAi` instead of `openai`).  
The provider instance is concatenated into the exception message, and without `__toString()` PHP throws “Object of class … could not be converted to string”, masking the intended `InvalidArgumentException`.

https://github.com/laravel/ai/blob/df3e8e0efb4ea9012723787032dbaccf047b8f59/src/Gateway/Prism/PrismGateway.php#L375

This change allows the provider to be safely rendered in that message.

### DIFF
--- a/src/Providers/Provider.php
+++ b/src/Providers/Provider.php
@@ -62,4 +62,12 @@ abstract class Provider
                 : [$key => $value]; // Provider name and model name...
         })->all();
     }
+    
+    /**
+     * Convert the provider to its string representation.
+     */
+    public function __toString(): string
+    {
+        return $this->driver();
+    }
 }


boynet:
we could change the error message to use $provider->driver() not sure which is preferred

PR #150, opened 2026-02-13
Fix invalid Message role and null response in HandlesRateLimiting

monayemislam:
## Changes

- **Message.php:** When a string is passed as `$role` that doesn't match any `MessageRole` enum case, `MessageRole::tryFrom($role)` returns null and was assigned to the non-nullable `$role` property, causing a TypeError. Now throws `InvalidArgumentException('Invalid message role.')` with a clear message.

- **HandlesRateLimiting.php:** `RequestException` can have a null `response` when the request never completed (e.g. connection timeout, DNS failure). Accessing `$e->response->status()` without a check could cause a null dereference. Now checks `$e->response !== null` before reading `->status()`.

### DIFF
--- a/src/Gateway/Concerns/HandlesRateLimiting.php
+++ b/src/Gateway/Concerns/HandlesRateLimiting.php
@@ -21,7 +21,7 @@ trait HandlesRateLimiting
         try {
             return $callback();
         } catch (RequestException $e) {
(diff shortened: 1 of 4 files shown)


monayemislam:
@taylorotwell can you please check this.

PR #153, opened 2026-02-13
Add Murf AI Provider support

prakhardubey2002:
Added Murf AI as a provider option for text-to-speech (TTS). Uses a dedicated gateway that calls Murf's `/v1/speech/generate` endpoint

**Supports:**

- Audio generation (text-to-speech)
- Default voice mapping 

**Configuration**

```env
MURF_API_KEY=
```

### DIFF
--- a/config/ai.php
+++ b/config/ai.php
@@ -94,6 +94,11 @@ return [
             'key' => env('JINA_API_KEY'),
         ],
 
+        'murf' => [
+            'driver' => 'murf',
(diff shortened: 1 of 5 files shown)


taylorotwell:
Thanks for your pull request to Laravel!

Unfortunately, I'm going to delay merging this code for now. To preserve our ability to adequately maintain the framework, we need to be very careful regarding the amount of code we include. 

If applicable, please consider releasing your code as a package so that the community can still take advantage of your contributions!

PR #156, opened 2026-02-13
Add Azure OpenAI vector store, file search & web search support

### LINKED ISSUE #155: [Azure OpenAI] Add file management support for Azure OpenAI Provider
I was really excited to see Azure OpenAI provider support added to the SDK! However, I noticed that file management is not yet supported. Attempting any file operation throws:
`Provider [Laravel\Ai\Providers\AzureOpenAiProvider] does not support file management.`

Azure OpenAI does support the [Files API](https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#files) and [File Search / Vector Stores](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/file-search), so it would be great to have parity with OpenAiProvider here. This is especially relevant for enterprise users who are required to use Azure OpenAI (compliance, data residency) and need FileSearch for RAG-based agents.

jesse-bos:
Added Azure OpenAI vector store, file search & web search support.
 
Note: Azure's /openai/v1/files endpoint requires purpose: `"assistants"` instead of purpose: `"user_data"`. The AzureOpenAiFileGateway overrides `filePurpose()` accordingly.

Fixes #155


pushpak1300:
Happy to accept PR based on new architecture.

PR #166, opened 2026-02-14
[0.x] Add support for configuring provider options on agents

### LINKED ISSUE #82: Support configuring provider options (e.g. OpenAI reasoning effort) on agents
## Problem

The Laravel AI SDK currently has no way for agents to specify provider-specific options that get passed through to Prism's `withProviderOptions()`.

This is a practical blocker for OpenAI's GPT-5 family (gpt-5, gpt-5-mini, gpt-5-nano), where the Responses API defaults `reasoning.effort` to `"medium"` when the parameter is omitted. For many agent use cases — simple classification, extraction, formatting — this burns unnecessary reasoning tokens, increasing both latency and cost. Setting reasoning effort to `"low"` or `"minimal"` isn't possible through the agent interface.

Prism already supports this on the OpenAI provider side:

```php
// vendor/prism-php/prism/src/Providers/OpenAI/Handlers/Text.php
'reasoning' => $request->providerOptions('reasoning'),
```

But there's no way to set this from an agent class. `TextGenerationOptions` only supports `maxSteps`, `maxTokens`, and `temperature` via attributes, and the `PrismGateway::generateText` method doesn't receive the agent — only the provider and options.

## Proposed Solution

Allow agents to define provider options. For example, a new interface:

```php

{
    // ...

    public function providerOptions(): array
    {
        return [
            'reasoning' => [
                'effort' => 'low',
            ],
        ];
    }
}
```

Or alternatively, dedicated attributes:

```php
use Laravel\Ai\Attributes\Reasoning;

#[Reasoning(effort: 'low')]
class MyAgent implements Agent
{
    // ...
}
```

Either approach would allow `CreatesPrismTextRequests` to forward these options to Prism's `withProviderOptions()`.

shafimsp:
This PR adds a HasProviderOptions interface that allows agents to specify provider-specific options (e.g., OpenAI's reasoning.effort, frequency_penalty, presence_penalty) that get passed through to Prism's withProviderOptions().

**Problem**

The SDK currently has no way for agents to configure provider-specific options. TextGenerationOptions only supports maxSteps, maxTokens, and temperature via attributes.

This is a practical blocker for OpenAI's reasoning models, where reasoning.effort defaults to "medium" when omitted — burning unnecessary tokens for simple tasks like classification or extraction. Other provider options like frequency_penalty and presence_penalty are also inaccessible.

**Solution**

A new HasProviderOptions interface that agents can implement:

```

{
    use Promptable;

    {
    }

    public function providerOptions(): array
    {
        return [
            'reasoning' => ['effort' => 'low'],
            'frequency_penalty' => 0.5,
            'presence_penalty' => 0.3,
        ];
    }
}
```

Why an interface over attributes?

• Provider options are diverse and provider-specific — attributes would proliferate quickly
• An interface allows dynamic options based on runtime context
• Aligns with existing patterns (HasTools, HasMiddleware, HasStructuredOutput)
• Community consensus in #82 favored this approach
Changes

• Add Laravel\Ai\Contracts\HasProviderOptions interface
• Modify TextGenerationOptions — extract provider options from agent via interface check
• Modify CreatesPrismTextRequests — merge agent's provider options with existing provider-specific defaults (agent options override)
• Add tests and test agent
Backward Compatibility

✅ Fully backward compatible. Existing agents without HasProviderOptions continue to work unchanged — providerOptions defaults to null and no additional options are passed.

Testing

• Provider options correctly extracted from agent implementing HasProviderOptions
• Options are null for agents without the interface
• Options passed through when prompting
• Options default to null when not provided
Fixes #82.

### DIFF
--- a/src/Contracts/HasProviderOptions.php
+++ b/src/Contracts/HasProviderOptions.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Laravel\Ai\Contracts;
+
+interface HasProviderOptions
+{
+    /**
+     * Get the provider-specific options to be passed to Prism.
+     *
+     * @return array<string, mixed>
+     */
+    public function providerOptions(): array;
+}
--- a/src/Contracts/HasProviderOptions.php
+++ b/src/Contracts/HasProviderOptions.php
@@ -2,6 +2,8 @@
 
 namespace Laravel\Ai\Contracts;
 
+use Laravel\Ai\Enums\Lab;
+
 interface HasProviderOptions
 {
     /**
@@ -9,5 +11,5 @@ interface HasProviderOptions
      *
      * @return array<string, mixed>
      */
-    public function providerOptions(): array;
+    public function providerOptions(Lab|string $provider): array;
 }
--- a/src/Contracts/HasProviderOptions.php
+++ b/src/Contracts/HasProviderOptions.php
@@ -7,7 +7,7 @@ use Laravel\Ai\Enums\Lab;
 interface HasProviderOptions
 {
     /**
-     * Get the provider-specific options to be passed to Prism.
+     * Get the provider-specific options to be passed to the provider.
      *
      * @return array<string, mixed>
      */
--- a/src/Gateway/Prism/Concerns/CreatesPrismTextRequests.php
+++ b/src/Gateway/Prism/Concerns/CreatesPrismTextRequests.php
@@ -60,11 +60,23 @@ trait CreatesPrismTextRequests
     protected function withProviderOptions($request, Provider $provider, ?array $schema, ?TextGenerationOptions $options)
     {
         if ($provider instanceof AnthropicProvider) {
+            $providerOptions = array_filter([
+                'use_tool_calling' => $schema ? true : null,
+            ]);
+
+            // Merge agent provider options (agent options can override defaults)
+            if (! is_null($options?->providerOptions)) {
+                $providerOptions = array_merge($providerOptions, $options->providerOptions);
+            }
+
             return $request
-                ->withProviderOptions(array_filter([
-                    'use_tool_calling' => $schema ? true : null,
-                ]))
+                ->withProviderOptions($providerOptions)
                 ->withMaxTokens($options?->maxTokens ?? 64_000);
+        }
+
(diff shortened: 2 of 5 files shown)


taylorotwell:
I think probably the method should receive the provider the agent is running a prompt for. That way you can return the correct options if you're using provider fallback.

shafimsp:
Updated! The method now receives the provider:

providerOptions(Lab|string $provider): array

fitzzz:
Nice work guys! This will be super helpful. I'm having major issues with some OpenRouter providers that time out, and I can't choose a specific provider I know has high uptime without extending PrismGateway.

grebe-vladin:
Is there a timeline of when this will be merged? it's a critical feature missing.

shafimsp:
@taylorotwell Any blockers on this one? I've addressed your feedback — providerOptions() now receives the provider name, supports per-provider config via match, and has full test coverage. 

Let me know if anything else needs changing!

taylorotwell:
Thanks!

CamilleScholtz:
Correct me if I'm wrong, but it seems like this does not work with Anthropic's cache control, as that is a provider option attached to the prompt right? Prism example:

```php
Prism::text()
    ->using(Provider::Anthropic, 'claude-3-5-sonnet-20241022')
    ->withSystemPrompt(
        (new SystemMessage('I am a long re-usable system message.'))
            ->withProviderOptions(['cacheType' => 'ephemeral', 'cacheTtl' => '1h'])
    )
 ```

shafimsp:
> Correct me if I'm wrong, but it seems like this does not work with Anthropic's cache control, as that is a provider option attached to the prompt right? Prism example:
> 
> 
> 
> ```php
> 
> Prism::text()
> 
>     ->using(Provider::Anthropic, 'claude-3-5-sonnet-20241022')
> 
>     ->withSystemPrompt(
> 
>         (new SystemMessage('I am a long re-usable system message.'))
> 
>             ->withProviderOptions(['cacheType' => 'ephemeral', 'cacheTtl' => '1h'])
> 
>     )
> 
>  ```


You're right, Anthropic's cache control is message-level, not request-level. This PR doesn't cover it.

PR #162, opened 2026-02-13
Add #[Thinking] attribute for extended thinking / reasoning support

anilcancakir:
This PR adds a `#[Thinking]` attribute that allows agents to opt into extended thinking (reasoning) capabilities offered by modern AI providers. The attribute follows the same pattern as the existing `#[MaxTokens]`, `#[Temperature]`, and `#[MaxSteps]` attributes.

## Motivation

Several major providers now support extended thinking / reasoning modes that significantly improve output quality for complex tasks:
- **Anthropic** — [Extended Thinking](https://docs.anthropic.com/en/docs/build-with-claude/extended-thinking) with `budget_tokens`
- **OpenAI** — [Reasoning Effort](https://platform.openai.com/docs/guides/reasoning#reasoning-effort) with `reasoning_effort` (low, medium, high) for o1/o3 models
- **Google Gemini** — [Thinking](https://ai.google.dev/gemini-api/docs/thinking) with `thinkingBudget` and `thinkingLevel`
- **xAI** — [Extended Thinking](https://docs.x.ai/docs/guides/extended-thinking) with `thinking.enabled`
- **Ollama** — [Thinking](https://ollama.com/blog/thinking) with boolean toggle
- **DeepSeek** — [Thinking Mode](https://api-docs.deepseek.com/guides/thinking_mode) with `thinking.type`
- **Groq** — [Reasoning](https://console.groq.com/docs/reasoning) with `reasoning_format` + `reasoning_effort`
- **Mistral** — [Reasoning](https://docs.mistral.ai/capabilities/reasoning) with `prompt_mode`
- **OpenRouter** — [Parameters](https://openrouter.ai/docs/api-reference/parameters) with `include_reasoning`
- **Azure OpenAI** — [Reasoning](https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/reasoning) with `reasoning_effort` (identical to OpenAI)

Currently there is no way to configure this through agent attributes.

## Usage

```php
use Laravel\Ai\Attributes\Thinking;

// Anthropic — enable with token budget
#[Thinking(budgetTokens: 10000)]
class DeepAnalysisAgent implements Agent { ... }

// OpenAI / Azure OpenAI — set reasoning effort for o1/o3 models
#[Thinking(effort: 'high')]
class ReasoningAgent implements Agent { ... }

// Gemini — both budget and effort level
#[Thinking(budgetTokens: 8000, effort: 'medium')]
class ResearchAgent implements Agent { ... }

[...]

PR #171, opened 2026-02-15
Add `#ReasoningEffort[]` Attribute

JonPurvis:
I recently updated my app to use `gpt-5-nano` in some places and then immediately realised how slow it was.

I forgot that they're reasoning models, which is the reason for the slowness. 

This PR adds a way to set the reasoning effort, in my case `#[ReasoningEffort('minimal')]`. With this in place, the responses are speedy and exactly what I need.


anilcancakir:
#162

taylorotwell:
Will work on PR commented above.

PR #180, opened 2026-02-18
fix: use null coalescing assignment operator in XaiProvider imageGate…

wuwx:


taylorotwell:
What does this fix exactly?

PR #190, opened 2026-02-19
Add Ollama dimensions option to embeddings

WoutervdBrink:
Ollama has [an option](https://docs.ollama.com/api/embed#body-dimensions) to specify dimensions when calculating embeddings. This pull request adds that option to the Prism gateway when the Ollama driver is used.

Upstream, Prism _also_ doesn't use the `dimensions` option yet for Ollama. I've [created a PR there](https://github.com/prism-php/prism/pull/913). Assuming that is merged, this PR adds support for it in the Laravel AI package.


WoutervdBrink:
The feature was added upstream, so this PR is ready to be reviewed.

PR #192, opened 2026-02-19
Include Unit testsuite in PHPUnit configuration

monayemislam:
Unit tests in tests/Unit/ were not part of the default PHPUnit config, so they weren’t run by ./vendor/bin/phpunit. This adds a Unit testsuite so all tests (Unit + Feature) run by default and in CI. All 9 Unit tests pass.


taylorotwell:
Unrelated chat changes here.

PR #198, opened 2026-02-20
Adding agent methods for text generation options

MaestroError:
This PR adds method-based resolution for `maxSteps`, `maxTokens`, and `temperature` on agents, following the same pattern already used by `provider()`, `model()`, and `timeout()`.

Currently these options can only be set via PHP 8 attributes, which are compile-time constants. This makes it impossible to configure them dynamically from a database, config, or any runtime source.

### Resolution priority (method → attribute → null)

```php
maxSteps: method_exists($agent, 'maxSteps')
    ? $agent->maxSteps()
    : (! empty($maxSteps) ? $maxSteps[0]->newInstance()->value : null),
```

### Usage

```php
{
    use Promptable;

    public function maxSteps(): int
    {
        return config('agents.support.max_steps', 5);
    }

    public function maxTokens(): int
    {
        return $this->user->plan->maxTokens();
    }

    public function temperature(): float
    {
        return Setting::get('support_agent_temperature', 0.7);
    }
}
```

Attributes continue to work as before — methods are checked first, then attributes, then `null`.

### Changes

- `TextGenerationOptions::forAgent()` — added `method_exists()` checks before attribute reflection
- Added `MethodOptionsAgent` and `MethodOverridesAttributeAgent` test agents
- Added `TextGenerationOptionsTest` with 3 tests covering method resolution, method-over-attribute priority, and the null fallback

### DIFF
--- a/src/Gateway/TextGenerationOptions.php
+++ b/src/Gateway/TextGenerationOptions.php
@@ -47,9 +47,15 @@ class TextGenerationOptions
         $temperature = $reflection->getAttributes(Temperature::class);
 
         return new self(
-            maxSteps: ! empty($maxSteps) ? $maxSteps[0]->newInstance()->value : null,
-            maxTokens: ! empty($maxTokens) ? $maxTokens[0]->newInstance()->value : null,
-            temperature: ! empty($temperature) ? $temperature[0]->newInstance()->value : null,
+            maxSteps: method_exists($agent, 'maxSteps')
+                ? $agent->maxSteps()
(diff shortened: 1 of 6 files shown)


hipig:
I urgently need the merger of this PR!!!

seankndy:
I am running into this limitation as well.  I am creating the agents dynamically and need a way to set these parameters at runtime.

In the meantime, I created 3 different dynamic agent classes each with varying temperatures, but being able to set at runtime along with max tokens would be huge.

pushpak1300 (approved):

PR #222, opened 2026-02-27
Handle audio file attachments in message conversion

sulimanbenhalim:
## Summary

- `Base64Audio`, `LocalAudio`, `RemoteAudio`, and `StoredAudio` were missing from the `match` in `PrismMessages::fromLaravelAttachments()`, causing `UnhandledMatchError` at runtime when any audio `File` type is used as a chat message attachment


sulimanbenhalim:
Closing to reopen as a clean PR with only this fix

PR #223, opened 2026-02-27
Handle audio file attachments in message conversion

sulimanbenhalim:
## Summary

- `Base64Audio`, `LocalAudio`, `RemoteAudio`, and `StoredAudio` were missing from the `match` in `PrismMessages::fromLaravelAttachments()`, causing `UnhandledMatchError` at runtime when any audio `File` type is used as a chat message attachment


taylorotwell:
Can you pass audio into conversations?

sulimanbenhalim:
GPT-4o (gpt-4o-audio-preview) and Gemini both support audio as conversation input natively so yes it works with those, Claude doesn't support audio in conversation context via the API yet so it would only apply to the models that do

PR #224, opened 2026-02-27
Pass StreamedAgentResponse to queued then() callbacks in BroadcastAgent

sulimanbenhalim:
BroadcastAgent::handle() was passing StreamableAgentResponse (returned by each()) to withCallbacks(), but then() callbacks expect StreamedAgentResponse (which extends AgentResponse). This captures the StreamedAgentResponse after stream consumption and passes that to callbacks instead, consistent with how InvokeAgent passes AgentResponse.

### DIFF
--- a/src/Jobs/BroadcastAgent.php
+++ b/src/Jobs/BroadcastAgent.php
@@ -29,11 +29,17 @@ class BroadcastAgent implements ShouldQueue
      */
     public function handle(): void
     {
-        $this->withCallbacks(fn () => $this->agent->stream($this->prompt, $this->attachments, $this->provider, $this->model)
+        $streamedResponse = null;
+
+        $this->agent->stream($this->prompt, $this->attachments, $this->provider, $this->model)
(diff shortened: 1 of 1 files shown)


taylorotwell:
I am closing this pull request because it lacks sufficient explanation, tests, or both. It is difficult for us to merge pull requests without these things because the change may introduce breaking changes to the framework.

Feel free to re-submit your change with a **thorough** explanation of the feature and tests - integration tests are preferred over unit tests. Please include it's benefit to end users; the reasons it does not break any existing features; how it makes building web applications easier, etc.

Thanks!

sulimanbenhalim:
Hey @taylorotwell the issue here is that BroadcastAgent was passing StreamableAgentResponse to withCallbacks() but then() callbacks expect StreamedAgentResponse, so anyone using a then() callback on a broadcast agent would get the wrong type which is inconsistent with how InvokeAgent handles it where it correctly passes AgentResponse after the response is fully resolved

PR #226, opened 2026-02-27
Improve base URL logic to prevent double path segments

Danni2901:
handle direct OpenAI V1 URLs in Azure provider: If the URL already contains '/openai/v1', return it immediately to prevent redundant path appending.


Danni2901:
As shown in the screenshot from Azure Portal, the provided Endpoint already includes the /openai/v1 segment. Without this PR, the SDK would append it again, resulting in an invalid URL.
image

PR #228, opened 2026-02-27
Add content moderation support via OpenAI moderations endpoint

ahmadmayahi:
- Add content moderation support via OpenAI's /moderations endpoint
- Support for all 13 moderation categories with flagged status and confidence scores

```php
use Laravel\Ai\Moderation;

$response = Moderation::check('I want to hurt someone');

 // true/false
$response->flagged;                      

// all categories
$response->categories;

// only flagged ones
$response->flaggedCategories(); 

// specific category
$response->category('violence');

```


taylorotwell:
Let me think about this one. I may bake it into Str class.

PR #236, opened 2026-03-02
Retrieve conversation list from ConversationStore and agent trait

### LINKED ISSUE #381: latestConversationId() can return the wrong conversation because parent updated_at is never refreshed
I found what looks like a conversation-store bug in `laravel/ai` `v0.1.5`.

`DatabaseConversationStore::latestConversationId()` selects the latest conversation by ordering `agent_conversations.updated_at DESC`.

However, `agent_conversations.updated_at` appears to only be written when the conversation row is created. When later user/assistant messages are stored, the parent conversation row is not updated.

## Why this matters

If a user has multiple remembered conversations, `continueLastConversation()` can resume the wrong one:

1. create conversation A
2. create conversation B
3. add more messages to conversation A
4. call `continueLastConversation()`

At that point, conversation A is the most recently active conversation, but conversation B may still win because its parent row has the newer `updated_at`.

## Expected

The “last conversation” lookup should reflect the most recently active conversation.

## Actual

It reflects the most recently created conversation row, unless something else updates the parent row externally.

## Relevant code

- `src/Storage/DatabaseConversationStore.php`
- `latestConversationId()`
- `storeConversation()`
- `storeUserMessage()`
- `storeAssistantMessage()`

Possible fixes could be either:

- updating `agent_conversations.updated_at` whenever a message is stored, or
- deriving “latest” from message timestamps instead of the parent row timestamp.

barryvdh:
Adds a way to list a user's past conversations so apps can render a "recent chats" sidebar without dropping down to raw SQL.

### Usage

```php
class User extends Authenticatable
{
    use HasConversations;
}

$conversations = $request->user()
    ->conversations()
    ->latest('updated_at')
    ->paginate(20);
```

### Why

`ConversationStore` only exposed `latestConversationId()`, so building a chat UI with a conversation list required querying the `agent_conversations` table directly. This gives users a first-class relationship to read from, and ensures `conversations.updated_at` advances as new messages arrive so ordering by recency works.


barryvdh:
Not sure if we want to return it as array like messages, or collection (from the trait).

And maybe we should include the created_at / updated_at?

taylorotwell:
I wonder if this should be traits you can add on to your `User` model for example that allows you to access the conversations via Eloquent? For both conversations and conversation messages.

barryvdh:
I guess that would make sense and make it easier to handle the conversations/messages

barryvdh:
Maybe we should provide base models for the conversations also?

pushpak1300:
```php
use Laravel\Ai\Concerns\HasConversations;

class User extends Authenticatable
{
    use HasConversations;
}

// List recent conversations
$conversations = $user->conversations();       // default limit 25
$conversations = $user->conversations(limit: 10);

// Each item is a Conversation DTO
foreach ($conversations as $conversation) {
    $conversation->id;
    $conversation->title;
    $conversation->userId;
    $conversation->createdAt;
    $conversation->updatedAt;
}

```

This is how `HasConversations` usage looks like.

pushpak1300 (approved):

Issue #204, opened 2026-02-23
Can't access all Prism meta properties


mattmcdev:
The Prism `Meta` class contains additional data about the request (`rate_limits` and `service_tier`).

Rather than mapping all possible values that could be included now or in the future, should there instead be a way to access the full metadata from Prism ?

delabon:
Hey @mattmcdev ,
I added a [PR](https://github.com/laravel/ai/pull/241) for that. After the PR is merged, you can access all additional data using $meta->extra (collection).
But I'm not sure if @taylorotwell will accept something like this.

PR #241, opened 2026-03-04
feat: add `extra` property to Meta class for full Prism metadata

delabon:
This PR adds an `extra` property to the `Meta` class that exposes all metadata from Prism responses, including fields like `rate_limits`, `service_tier`, and `id`.

The Prism Meta class contains additional data about the request (rate_limits, service_tier, id) that wasn't being exposed to Laravel AI SDK users. Manually mapping each field would:
- Require continuous updates as Prism adds new fields
- Bloat the Meta class with provider-specific fields
- Miss undocumented or experimental fields

Added an `extra` Collection property to the `Meta` class that captures all Prism metadata. The `PrismMeta` converter filters out the `model` field (which is already a top-level property) and includes everything else in `extra`.

### Usage
$extra = SalesCoach::make()->prompt('Test')->meta->extra
$extra->get('id'); // 'req_abc123'
$extra->get('rate_limits'); // [...]
$extra->get('service_tier'); // 'auto'

Requested by https://github.com/laravel/ai/issues/204

### DIFF
--- a/src/Gateway/Prism/PrismGateway.php
+++ b/src/Gateway/Prism/PrismGateway.php
@@ -91,12 +91,14 @@ class PrismGateway implements Gateway
             new Collection($response->additionalContent['citations'] ?? [])
         );
 
+        $meta = PrismMeta::toLaravelMeta($provider->name(), $response->meta, $citations);
+
         return $structured
             ? (new StructuredTextResponse(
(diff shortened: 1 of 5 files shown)


taylorotwell:
I can't guarantee everything will always be Prism under the hood.

PR #251, opened 2026-03-07
Allow instructions() to return an array for prompt cache optimization

jlcara:
Providers like Anthropic and Gemini support prompt caching, but only when static content is isolated in its own system message block. Today, `instructions()` returns a single string, so any per-request context (variables, user-specific data) mixed into the prompt invalidates the entire cache on every call. That means you're paying full token costs repeatedly for content that never changes.

By allowing `instructions()` to return an array, each element becomes a separate `withSystemPrompt()` call. This lets you split long, static instructions into one block (cached across requests) and volatile context into another. On prompt-heavy agents, this can significantly reduce API costs.

Fully backward compatible — returning a string works identically. New test verifies both array elements reach the Prism request as separate `SystemMessage` objects.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -19,7 +19,7 @@ use Laravel\Ai\Contracts\Agent;
  * Get an ad-hoc agent instance.
  */
 function agent(
-    string $instructions = '',
+    string|array $instructions = '',
     iterable $messages = [],
     iterable $tools = [],
     ?Closure $schema = null,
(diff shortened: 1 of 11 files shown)


taylorotwell:
This isn't how OpenAI describes prompt caching as working. Static prefixes, even within a single system prompt, are still cached: https://developers.openai.com/api/docs/guides/prompt-caching/

PR #258, opened 2026-03-10
[0.x] Create a CI test workflow

jackbayliss:
Feel free to close if this is too early, or you'd prefer to handle it internally.

This adds a workflow for the unit tests (integration doesnt work because it expects real API tokens, so left it out for now)                                                     
                                                                                                                                                                                 
Without a CI workflow, reviewers / contributors have no feedback on whether a PR breaks. 

I've left code styling out as wasnt sure if you'd  use Style CI etc. 

so TLDR is this just gives at least some feedback rather than none, and helps spotting any regression unit test wise.

left fairly basic so it can be improved but follows all the other laravel package tests.

Here's the action output: https://github.com/jackbayliss/ai/actions/runs/22927878150
🫡

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -26,7 +26,7 @@
     "require-dev": {
         "laravel/pint": "^1.26",
         "mockery/mockery": "^1.6.12",
-        "orchestra/testbench": "^10.6"
+        "orchestra/testbench": "^10.6|^11.0"
     },
     "minimum-stability": "dev",
     "prefer-stable": true,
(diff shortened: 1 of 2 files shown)


composer.json
```diff
@@ -26,7 +26,7 @@
     "require-dev": {
         "laravel/pint": "^1.26",
         "mockery/mockery": "^1.6.12",
-        "orchestra/testbench": "^10.6"
+        "orchestra/testbench": "^10.6|^11.0"
```
jackbayliss, line 29: Needed for laravel 13 tests

PR #230, opened 2026-02-27
CLI providers for subscription based AI usage

sulimanbenhalim:
## Summary

So basically right now if you wanna use Laravel AI you need API keys for every provider which means you gotta set up billing, get credits, manage keys etc

But most devs already have subscriptions to Claude, Codex, Gemini through their CLI tools. Like if you have a Claude Pro/Max subscription you can literally use the `claude` command in headless mode as an API alternative

This PR adds CLI gateway providers that let you use your existing AI subscriptions instead of paying for API access separately. Works with Claude Code, OpenAI Codex CLI and Google Gemini CLI

Reference on how Claude Code headless mode works as an API alternatie: https://code.claude.com/docs/en/headless

### Whats included

CLI gateway abstraction that handles process executon, session management for multi turn conversations, streaming support and environment isolation so nested CLI tools dont break each other

Three concrete gateways
- **ClaudeCliGateway** with full session resumption via `--resume`, structured output with json schema, and stream json parsing
- **CodexCliGateway** with JSONL output parsing, session continuity, and structured output support  
- **GeminiCliGateway** lightweight wrapper since gemini cli doesnt support sessions in headless mode

Plus a `CliProvider` that plugs right into the existing provider system so you can use it like any other provider with `Ai::provider('claude-cli')` and configure models, failovers etc

### Config

```php
'claude-cli' => [
    'driver' => 'claude-cli',
    'timeout' => 300,
    'binary' => 'claude',     // optional custom path
    'default_model' => 'sonnet',
    'models' => [
        'cheapest' => 'haiku',
        'smartest' => 'opus',
    ],
],
```

### Tests

197 tests covering basically evrything, unit tests for all gateway methods, edge cases like unicode handling, falsy PHP strings, command injection saftey, session isolation between instances, streaming timeout behavior etc. Also feature tests for the provider and manager factory integration

## Test plan

- [x] All 197 new tests passing
- [x] Existing unit tests unaffected (196 pass)
- [x] Existing feature tests unaffected (failures are pre existing API key issues)
- [x] Manual test with real Claude CLI subscription
- [x] Manual test with Codex CLI subscription
- [x] Manual test with Gemini CLI subscription

PR #266, opened 2026-03-13
Add CLI providers to support AI subscriptions as an alternative to API keys

sulimanbenhalim:
Supersedes #230

I know this was closed for being too much code to maintain but I want to make the case for why this one is worth it

Right now Laravel AI requires API keys and billing setup for every provider, most devs and companies already pay monthly for Claude Pro/Max, ChatGPT Plus or Gemini Advanced subscriptions and will never set up separate API billing just to use Laravel AI, that's a real adoption wall

This PR removes that wall completley by letting you use the CLI tools that come with those subscriptions as providers, no API key, no billing setup, just point Laravel AI at the binary thats already on the machine

**Why this belongs in core and not a package**

This is a first class feature not a plugin, the whole value proposition of Laravel AI is that it works with the AI tools devs already have, CLI subscriptions are how a huge chunk of the dev community actually uses AI today, making them a second class citizen in a separate package sends the wrong message and fragments the ecosystem

Laravel AI would be the first major framework to support bring your own subscription natively, that's a real differentiator and a marketing story that writes itself, no other framework does this

**Whats included**

CLI gateway abstraction handling process execution, session management for multi turn conversations, streaming support and environment isolation

Three gateways:
- ClaudeCliGateway with full session resumption via --resume, structured output with json schema and stream json parsing
- CodexCliGateway with JSONL output parsing, session continuity and structured output support
- GeminiCliGateway lightweight wrapper for Gemini CLI headless mode

CliProvider that plugs into the existing provider system so you use it exactly like any other provider

**Config**

```php
'claude-cli' => [
    'driver' => 'claude-cli',
    'timeout' => 300,
    'binary' => 'claude',
    'default_model' => 'sonnet',
],
```

**Tests**

197 tests covering all gateway methods, edge cases, unicode handling, command injection safety, session isolation, streaming timeout behavior and feature tests for provider and manager integration

### DIFF
--- a/config/ai.php
+++ b/config/ai.php
@@ -124,6 +124,24 @@ return [
             'driver' => 'xai',
             'key' => env('XAI_API_KEY'),
         ],
+
+        // 'claude-cli' => [
+        //     'driver' => 'claude-cli',
+        //     'default_model' => 'sonnet',
+        //     'timeout' => 300,
+        // ],
+
+        // 'codex-cli' => [
+        //     'driver' => 'codex-cli',
+        //     'default_model' => 'gpt-4.1',
+        //     'timeout' => 300,
+        // ],
+
+        // 'gemini-cli' => [
+        //     'driver' => 'gemini-cli',
+        //     'default_model' => 'gemini-2.5-flash',
+        //     'timeout' => 120,
+        // ],
     ],
 
 ];
--- a/src/AiManager.php
+++ b/src/AiManager.php
@@ -15,8 +15,12 @@ use Laravel\Ai\Contracts\Providers\TextProvider;
 use Laravel\Ai\Contracts\Providers\TranscriptionProvider;
 use Laravel\Ai\Enums\Lab;
 use Laravel\Ai\Gateway\Prism\PrismGateway;
+use Laravel\Ai\Gateway\Cli\ClaudeCliGateway;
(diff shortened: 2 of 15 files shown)


fosron:
May i propose adding a generic CLI gateway and then having the specific CLI adapters under a separate package? That way theres less code to manage on the AI package and i feel like that would have way bigger chance of getting merged.

pushpak1300:
Yes I agree with @fosron here.

pushpak1300:
I like the idea in general but I'm pretty sure we don't want to maintain is as first party all the CLI adapters as they change all the time. one adapter Laravel can support out of the box is `opencode` and then all can be shipped as individual package. Thats my opinion. 

will still need to get taylors thought on this.

sulimanbenhalim:
Hey @pushpak1300 @fosron just wanted to share that I personaly been running Claude Code in headless mode on a real business project for almost 3 months now and literaly never touched the implementation after the initial setup, zero maintenance, same commands same output the entire time so from real usage the stability concern is lower than it might seem, whatever approach you guys decide to go with I fully support it, just wanted to put that out there because I think the "bring your own subscription" angle is genuinely huge, a lot of devs including myself already pay for Claude/ChatGPT subscriptions and would never setup separate API billing just to use a framework

pushpak1300:
Thanks for your pull request!

PR #272, opened 2026-03-15
[0.x] Remove redundant Collection wrapping in StreamEnd::combineUsage().

Button99:
This PR removes redundant Collection wrapping in `StreamEnd::combineUsage()`

The method accepts `array|Collection` and already normalizes `$events` into a `Collection`. The extra `new Collection($events)` wrap on line 26 was therefore redundant.

### DIFF
--- a/src/Streaming/Events/StreamEnd.php
+++ b/src/Streaming/Events/StreamEnd.php
@@ -23,7 +23,7 @@ class StreamEnd extends StreamEvent
     {
         $events = is_array($events) ? new Collection($events) : $events;
 
-        return (new Collection($events))->whereInstanceOf(StreamEnd::class)
+        return $events->whereInstanceOf(StreamEnd::class)
             ->values()
             ->map
             ->usage


src/Streaming/Events/StreamEnd.php
```diff
@@ -23,7 +23,7 @@ public static function combineUsage(Collection|array $events): Usage
     {
         $events = is_array($events) ? new Collection($events) : $events;
```
siarheipashkevich, line 24: @Button99 maybe we can use `$events = Collection::wrap($events)` here?
https://laravel.com/docs/12.x/collections#method-wrap

Button99: Hello @siarheipashkevich , if `$events` has already been normalized to a collection, using `Collection::wrap($events)` here would be redundant.

siarheipashkevich: @Button99 This method exists precisely for this purpose, so as not to make checks.

Issue #189, opened 2026-02-19
Structured output validation


dnyg:
Hi

I am a bit new to `laravel/ai` but I have a question regarding agents that implements `HasStructuredOutput`

If the structured format of the `StructuredAgentResponse` doesn't follow the format defined in the schema - shouldn't an exception be thrown? Should this happen on the Prism level, the Laravel/Ai level - or am I supposed to do this myself?

I don't really see any point in structured output if I have no guarantee that the result will conform to my schema.

mikescott:
Responses aren't checked against the schema and there seem to be issues getting Anthropic to provide a valid schema, especially with Haiku. 

You can construct the schema from your agent and validate a response against it with [opis/json-schema](https://github.com/opis/json-schema)

dwightwatson:
I've noticed this specifically when upgrading from v0.3.1 to v0.4.0 with OpenAI, and I believe it's related to #275 which replaced the Prism gateway with a custom OpenAI gateway.

It seems that the Prism gateway (which was just removed in #414) may have handled schema validation and now that's missing.

I have this issue in production, but given that it's AI it's hard to reproduce consistently. However, I'm getting `Undefined array key "phone_numbers"` exceptions for responses that are meant to contain that key.

```php
/**
 * Get the agent's structured output schema definition.
 */
public function schema(JsonSchema $schema): array
{
    return [
        'phone_numbers' => $schema->array()
            ->items($schema->string()->required())
            ->required(),
    ];
}
```

pushpak1300:
@dwightwatson I am not sure what exactly the issue. schema validation was not at all supported in prism also if you try latest version to see if the issue you are facing is still valid.

dwightwatson:
I'm on the latest version and it's still an issue. It didn't appear to be a problem on v0.3.1.

I'm confused then - what's the point of [structured output](https://laravel.com/docs/13.x/ai-sdk#structured-output) if it isn't guaranteeing that the response matches the structure? I would have expected that if I have marked a key as `required` then it should be there when I reference it.

para1992:
I can work on this using `opis/json-schema` for validation. Invalid responses would throw a dedicated `InvalidStructuredOutputException`; no automatic retry initially. Would adding this dependency and using fail-fast behavior be acceptable?

PR #275, opened 2026-03-16
Add OpenAI gateway for Responses API

pushpak1300:
The OpenAI provider previously went through Prism PHP for text generation. This adds a dedicated gateway that talks directly to the OpenAI Responses API, giving full access to provider-specific features while other providers continue using Prism.

### Why a direct gateway instead of staying on Prism?

Prism's abstraction forces OpenAI-specific features like `previous_response_id`, provider tool options (`web_search_preview`, `file_search` with filters), and `text.format` structured output through `withProviderOptions()`, an escape hatch that defeats the purpose of the abstraction. A direct gateway lets us adopt new Responses API features immediately without waiting for upstream releases, removes one translation layer for easier debugging, and gives direct control over rate limiting, streaming, and error handling. 

**Other providers (Anthropic, Gemini, Groq, etc.) still use Prism.**

https://developers.openai.com/api/docs/guides/text?lang=curl

> [!WARNING]
> This changes the OpenAI provider's gateway from `PrismGateway` to `OpenAiGateway`. The public API (`Ai` facade, agents, tools) is unchanged, but any code that directly references or extends `PrismGateway` for OpenAI will need updating.

### DIFF
--- a/src/AiManager.php
+++ b/src/AiManager.php
@@ -14,6 +14,7 @@ use Laravel\Ai\Contracts\Providers\StoreProvider;
 use Laravel\Ai\Contracts\Providers\TextProvider;
 use Laravel\Ai\Contracts\Providers\TranscriptionProvider;
 use Laravel\Ai\Enums\Lab;
+use Laravel\Ai\Gateway\OpenAi\OpenAiGateway;
 use Laravel\Ai\Gateway\Prism\PrismGateway;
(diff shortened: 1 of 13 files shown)


dwightwatson:
I have a concern that this change may have affected structured schema validation - I'm wondering if these custom gateways implement validation the same way Prism did?

More info here: #189

PR #278, opened 2026-03-17
[0.x] Handle rate limiting in Cohere, Voyage AI, and Jina gateways

Button99:
This PR fixes missing rate limit handling in the Cohere, VoyageAI, and Jina gateways.

It wraps the affected HTTP calls with `withRateLimitHandling()` so 429 responses throw `RateLimitedException`.


pushpak1300:
Fixed in the prism migrations.

PR #282, opened 2026-03-18
Pass through provider specific Gemini image aspect ratios

morcken:
Gemini supports image aspect ratios such as `3:4`, but the SDK currently drops any ratio outside `1:1`, `2:3`, and `3:2` when building Gemini image options.

This PR makes Gemini consistent with the existing `size(string $size)` API by passing through non-null size values as `aspect_ratio`, while preserving the existing mappings for `1:1`, `2:3`, and `3:2`.

## Reproduction

```php
use Laravel\Ai\Enums\Lab;
use Laravel\Ai\Files;
use Laravel\Ai\Image;

return Image::of($prompt)
    ->quality('medium')
    ->size('3:4')
    ->timeout(60)
    ->generate(provider: Lab::Gemini);
```

## Before
- `GeminiProvider::defaultImageOptions()` converted unknown aspect ratios to `null`, so `size('3:4')` was not forwarded to Gemini.

## After
- Non-null `size` values are passed through to Gemini as `aspect_ratio`.

## Why this change
- Gemini supports `3:4` and other aspect ratios (https://ai.google.dev/gemini-api/docs/image-generation#3.1-flash-image-preview).
- `size()` already accepts a string at the public API level.
- OpenAI already forwards provider-specific size values instead of narrowing them to the three built-in ratios.
- This makes provider behavior more consistent without changing the convenience helpers.

## Tests
Added a focused unit test that verifies:
- Gemini preserves `3:4` as `aspect_ratio`
- An existing mapped ratio (`1:1`) still behaves as before

## Backward Compatibility
This change is backward compatible:
- No method signatures changed
- `square()`, `portrait()`, and `landscape()` are unchanged
- Existing known Gemini mappings are unchanged


saqueib:
@morcken thanks man, I wasn't able to pass aspect ratio like `16:9` I hope it get merged to fix this

pushpak1300:
hey can you please fix the conflicts accordingly to new changes we did to Gemini Gateway.

pushpak1300 (approved):

pushpak1300:
Thanks.

PR #284, opened 2026-03-18
Allow external Prism providers in PrismGateway

### LINKED ISSUE #283: Support external Prism providers in PrismGateway (bedrock, workers-ai, etc.)
## Problem

`PrismGateway::toPrismProvider()` uses a hardcoded `match` statement to map AI SDK driver names to `PrismProvider` enum values. External Prism providers registered via `PrismManager::extend()` — like [`prism-php/bedrock`](https://github.com/prism-php/bedrock) or custom Workers AI providers — cannot be used through the `agent()` system because there is no corresponding enum case or fallback.

```php
// PrismGateway.php line 382
protected static function toPrismProvider(Provider $provider): PrismProvider
{
    return match ($provider->driver()) {
        'anthropic' => PrismProvider::Anthropic,
        // ... hardcoded list ...
        'xai' => PrismProvider::XAI,
        default => throw new InvalidArgumentException('Gateway does not support provider ['.$provider.'].'),
    };
}
```

Meanwhile, Prism's `using()` method already accepts `string|ProviderEnum`:

```php
// ConfiguresProviders.php
public function using(string|ProviderEnum $provider, ...): self
```

And `PrismManager::resolve()` already supports custom providers via `extend()`:

```php
// PrismManager.php
if (isset($this->customCreators[$name])) {
    return $this->callCustomCreator($name, $config);
}
```

So the Prism layer is ready — the gap is only in the Laravel AI SDK gateway layer.

## Who this affects

- **[prism-php/bedrock](https://github.com/prism-php/bedrock)** — AWS Bedrock provider, published on Packagist, 100k+ downloads. Works with `Prism::text()->using('bedrock', ...)` but not `agent()->prompt(...)`.
- **Any external Prism provider** registered via `PrismManager::extend()`.
- Related: #132 and #261 — users wanting to use custom/self-hosted backends through the agent system.

## Proposed Fix

Change `toPrismProvider()` to fall back to the driver name as a string for unrecognized drivers, instead of throwing. Since `Prism::using()` already accepts strings and `PrismManager` resolves them through `customCreators`, this is a one-line change:

```php
protected static function toPrismProvider(Provider $provider): PrismProvider|string
{
    return match ($provider->driver()) {
        'anthropic' => PrismProvider::Anthropic,
        'azure' => PrismProvider::OpenAI,
        'deepseek' => PrismProvider::DeepSeek,
        'gemini' => PrismProvider::Gemini,
        'groq' => PrismProvider::Groq,
        'mistral' => PrismProvider::Mistral,
        'ollama' => PrismProvider::Ollama,
        'openai' => PrismProvider::OpenAI,
        'openrouter' => PrismProvider::OpenRouter,
        'voyageai' => PrismProvider::VoyageAI,
        'xai' => PrismProvider::XAI,
        default => $provider->driver(), // Fall back to string for custom Prism providers
    };
}
```

The `configure()` method in `CreatesPrismTextRequests` already passes the result to `$prism->using()`, which accepts `string|ProviderEnum` — so no other changes are needed.

For users, this means any external Prism provider that registers via `PrismManager::extend()` would automatically work through `agent()`:

```php
// config/ai.php
'workers-ai' => [
    'driver' => 'workers-ai',  // matches PrismManager custom creator key
    'key' => env('CLOUDFLARE_AI_API_KEY'),
    'url' => env('WORKERS_AI_URL'),
],
```

```php
// Just works
agent(instructions: 'You are helpful.')
    ->prompt('Hello!', provider: 'workers-ai');
```

## Why string fallback is safe

1. If the driver name matches a registered `PrismManager::extend()` creator → it resolves correctly
2. If the driver name matches a built-in `create{Name}Provider()` method → it resolves correctly
3. If neither exists → `PrismManager::resolve()` throws `InvalidArgumentException("Provider [{$name}] is not supported.")` — same error class, better message

The current behavior throws immediately at the gateway layer before Prism even gets a chance to resolve. The proposed behavior defers to Prism's own resolution, which already handles the error case.

meirdick:
## Summary

- `toPrismProvider()` now returns the driver name as a string for unrecognized drivers instead of throwing `InvalidArgumentException`
- Return type changed from `PrismProvider` to `PrismProvider|string` (matches what `Prism::using()` already accepts)
- All existing built-in drivers continue to map to their `PrismProvider` enum values — zero behavior change for current users

## Problem

External Prism providers registered via `PrismManager::extend()` — like [prism-php/bedrock](https://github.com/prism-php/bedrock) (100k+ Packagist downloads) — cannot be used through the `agent()` system because `toPrismProvider()` throws for any driver not in its hardcoded match.

The Prism layer already supports this: `Prism::using()` accepts `string|ProviderEnum`, and `PrismManager::resolve()` checks `customCreators` first. The gap is only in the Laravel AI gateway layer.

## Fix

```diff
- default => throw new InvalidArgumentException('Gateway does not support provider ['.$provider.'].'),
+ default => $provider->driver(),
```

The string driver name flows through to `Prism::using()` → `PrismManager::resolve()` which resolves custom providers via `extend()`, or throws its own `InvalidArgumentException("Provider [{$name}] is not supported.")` if truly unknown.

## Test plan

- [x] All 11 built-in drivers still return correct `PrismProvider` enum values
- [x] Unknown drivers (e.g. `'bedrock'`, `'workers-ai'`) return the driver name as a string
- [x] All 21 existing unit tests pass (91 assertions)

Closes #283

### DIFF
--- a/src/Gateway/Prism/PrismGateway.php
+++ b/src/Gateway/Prism/PrismGateway.php
@@ -378,8 +378,13 @@ class PrismGateway implements Gateway
 
     /**
      * Map the given Laravel AI provider to a Prism provider.
+     *
+     * Built-in drivers are mapped to their corresponding Prism provider enum.
+     * Unknown drivers fall through as strings, allowing external Prism providers
+     * registered via PrismManager::extend() (e.g. prism-php/bedrock) to be
+     * resolved by Prism's own provider resolution.
      */
-    protected static function toPrismProvider(Provider $provider): PrismProvider
+    protected static function toPrismProvider(Provider $provider): PrismProvider|string
     {
         return match ($provider->driver()) {
             'anthropic' => PrismProvider::Anthropic,
@@ -393,7 +398,7 @@ class PrismGateway implements Gateway
             'openrouter' => PrismProvider::OpenRouter,
             'voyageai' => PrismProvider::VoyageAI,
             'xai' => PrismProvider::XAI,
-            default => throw new InvalidArgumentException('Gateway does not support provider ['.$provider.'].'),
+            default => $provider->driver(),
         };
     }
 
--- a/tests/Unit/Gateway/Prism/ToPrismProviderTest.php
+++ b/tests/Unit/Gateway/Prism/ToPrismProviderTest.php
@@ -1,0 +1,81 @@
+<?php
+
+namespace Tests\Unit\Gateway\Prism;
+
+use InvalidArgumentException;
+use Laravel\Ai\Gateway\Prism\PrismGateway;
+use Laravel\Ai\Providers\Provider;
+use Prism\Prism\Enums\Provider as PrismProvider;
+use ReflectionMethod;
+use Tests\TestCase;
+
+class ToPrismProviderTest extends TestCase
+{
+    /**
+     * Call the protected static toPrismProvider method via reflection.
+     */
+    protected function callToPrismProvider(Provider $provider): PrismProvider|string
+    {
+        $method = new ReflectionMethod(PrismGateway::class, 'toPrismProvider');
+
+        return $method->invoke(null, $provider);
+    }
+
+    /**
+     * Create a minimal Provider stub with the given driver name.
+     */
+    protected function makeProvider(string $driver): Provider
+    {
+        return new class($driver) extends Provider
+        {
+            public function __construct(string $driver)
+            {
+                parent::__construct(
+                    new \Laravel\Ai\Gateway\Prism\PrismGateway(app('events')),
+                    ['driver' => $driver, 'key' => 'test-key', 'name' => $driver],
+                    app('events'),
+                );
+            }
+        };
+    }
+
+    public function test_built_in_drivers_return_prism_provider_enum(): void
+    {
+        $mappings = [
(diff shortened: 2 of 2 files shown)


andrew-alchemy:
Why was this closed?
There is some ongoing inexplicable antipathy to supporting Bedrock (numerous PRs and request all ignored for many months across both Prism and AI repos) - and changes like this that let us extend AI to add our own providers is shut down.
I don't get it. 
I'd like some statement from the Laravel gatekeepers on what exactly the plan is with regards to both Bedrock and extensibility. 
I think the strategy of binding in specific support to the AI package is a mess. Each provider should have been a separate maintained plug in. All I see is an endless sea of tickets complaining some new LLM-specific edge case doesn't work and then the whole AI package needs updating.

meirdick:
@andrew-alchemy My guess is the team has a roadmap and this is probably coming down the pipe, and the team has it covered. Now we wait!

Most likely well see Laravel AI moving toward provider-specific gateways that bypass Prism for direct API access (Rationale from PR #275: "Prism's abstraction forces OpenAI-specific features through withProviderOptions(), an escape hatch that defeats the purpose of the abstraction")

PR #285, opened 2026-03-18
Add cache control support for system prompts, messages, and tools

CamilleScholtz:
## Summary

There's currently no way to configure cache control on system prompts, messages, or tools. Providers like Anthropic support [prompt caching](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching) via cache breakpoints, and Prism already supports this via `withProviderOptions()` on messages and tools — but the Laravel AI SDK doesn't expose it.

This PR adds cache control support at three levels:

### 1. System prompt caching (via agent provider options)

```php
{
    {
        return match ($provider) {
            Lab::Anthropic => [
                'system_prompt_caching' => 'ephemeral',
                // or with TTL:
                // 'system_prompt_caching' => ['type' => 'ephemeral', 'ttl' => '5m'],
            ],
            default => [],
        };
    }
}
```

Anthropic supports `'ephemeral'` as the cache type, with TTL values of `'5m'` (default) or `'1h'`.

### 2. Message-level cache control

Adds `withProviderOptions()` to the `Message` base class. Options flow through to Prism messages during conversion:

```php
use Laravel\Ai\Messages\UserMessage;

$message = (new UserMessage('...'))->withProviderOptions([
    'cacheType' => 'ephemeral',
]);
```

Works on `UserMessage`, `AssistantMessage`, and `ToolResultMessage`.

### 3. Tool cache control

When a tool implements `HasProviderOptions`, its provider options are passed through to the Prism tool:

```php
use Laravel\Ai\Enums\Lab;

class SearchTool implements Tool, HasProviderOptions
{
    {
        return match ($provider) {
            Lab::Anthropic => ['cacheType' => 'ephemeral'],
            default => [],
        };
    }
}
```

### Changes

- **`PrismGateway`**: Extracted `applySystemPrompt()` that reads `system_prompt_caching` from agent provider options and wraps instructions in a `SystemMessage` with cache control. Updated `addTools()` to pass the provider through.
- **`CreatesPrismTextRequests`**: Strips `system_prompt_caching` from agent provider options before passing to Prism (gateway-level concern, not a request-level Prism option).
- **`AddsToolsToPrismRequests`**: `createPrismTool()` now checks if the tool implements `HasProviderOptions` and passes options to the Prism tool via `withProviderOptions()`.
- **`Message`**: Added `$providerOptions` property and `withProviderOptions()` method.
- **`PrismMessages`**: Updated `fromLaravelMessages()` to transfer provider options from Laravel AI messages to Prism messages for all message types.

### DIFF
--- a/src/Gateway/Prism/Concerns/AddsToolsToPrismRequests.php
+++ b/src/Gateway/Prism/Concerns/AddsToolsToPrismRequests.php
@@ -4,10 +4,12 @@ namespace Laravel\Ai\Gateway\Prism\Concerns;
 
 use Illuminate\JsonSchema\JsonSchemaTypeFactory;
 use Illuminate\Support\Collection;
+use Laravel\Ai\Contracts\HasProviderOptions;
 use Laravel\Ai\Contracts\Providers\SupportsFileSearch;
 use Laravel\Ai\Contracts\Providers\SupportsWebFetch;
 use Laravel\Ai\Contracts\Providers\SupportsWebSearch;
 use Laravel\Ai\Contracts\Tool;
+use Laravel\Ai\Enums\Lab;
 use Laravel\Ai\Gateway\Prism\PrismTool;
 use Laravel\Ai\Gateway\TextGenerationOptions;
 use Laravel\Ai\ObjectSchema;
@@ -27,12 +29,12 @@ trait AddsToolsToPrismRequests
     /**
      * Add the given tools to the Prism request.
      */
-    protected function addTools($request, array $tools, ?TextGenerationOptions $options = null)
+    protected function addTools($request, Provider $provider, array $tools, ?TextGenerationOptions $options = null)
     {
         return $request
             ->withTools(
-                (new Collection($tools))->map(function ($tool) {
-                    return ! $tool instanceof ProviderTool ? $this->createPrismTool($tool) : null;
+                (new Collection($tools))->map(function ($tool) use ($provider) {
+                    return ! $tool instanceof ProviderTool ? $this->createPrismTool($tool, $provider) : null;
                 })->filter()->values()->all()
             )
             ->withToolChoice(ToolChoice::Auto)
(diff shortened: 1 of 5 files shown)


alexpgates:
Tested against a tool-heavy agent and seeing ~8,200 prompt tokens of fixed overhead on every request... even for a one-word message. That's almost entirely system prompt and tool definitions being resent each turn. Caching this would be a meaningful cost reduction for agentic use cases. Hoping to see this merged!

ipalaus:
Great add! I was looking into how to do with Laravel's AI SDK as we also have heavy input tokens.

FYI 5m cache is 90% cost reduction, 1hr 75% savings

khalidmaquilang:
Looking forward with this PR!

adamthehutt:
Thanks for putting this together, @CamilleScholtz — the design is exactly the right shape, and `system_prompt_caching` inside `providerOptions()` slots cleanly into the `HasProviderOptions` contract.

I wanted to flag a timing wrinkle for the maintainers that I think makes this PR harder to merge as-is: PR #309 (the direct Anthropic gateway) was opened a week after this PR and shipped in v0.4.0. As of v0.4, the default Anthropic provider no longer routes through `PrismGateway` — it uses `Laravel\Ai\Gateway\Anthropic\AnthropicGateway` directly. The `BuildsTextRequests` concern there builds the system prompt as a plain string and has no per-content-block annotation hook, so the changes in this PR (which all live in the Prism path) wouldn't affect Anthropic users on v0.4+.

I just hit this concretely on a real project: even with `HasProviderOptions` returning `cache_control` keys, no caching happens because the direct gateway just `array_merge`s providerOptions into the top-level body, where `cache_control` isn't a valid Anthropic field. We're paying ~\$0.50 per agent invocation that should be ~\$0.20 with caching enabled, and there's currently no way to fix this without forking the gateway.

Two ways forward and I'm happy to help with either:

1. **Expand this PR** to also annotate the system prompt and tool result content blocks in `AnthropicGateway::buildBody` (and the equivalent message-level path), keeping the same `system_prompt_caching` API shape you already proposed. I can contribute the direct-gateway changes onto your branch if you'd like.

2. **Open a complementary PR** that targets only the direct gateway, mirroring this PR's API verbatim and explicitly referencing it for design. This keeps the two PRs independently mergeable and lets Anthropic users get caching even if the Prism-path changes in this PR are deferred.

Either path is fine with me. Just don't want to step on your work without coordinating. Whatever shape lands, I'll happily port our internal use case onto it and report results.

pushpak1300:
Closing because of #410

PR #286, opened 2026-03-18
Add Concurrent marker interface for tools

CamilleScholtz:
## Summary

Prism PHP supports [concurrent tool execution](https://prismphp.com/docs/core-concepts/tools-function-calling#concurrent-tool-execution) via `->concurrent()` on tools, allowing multiple I/O-bound tool calls to run in parallel using Laravel's Concurrency facade. However, the Laravel AI SDK has no way to expose this — `createPrismTool()` never calls `->concurrent()`.

This PR adds a `Concurrent` marker interface that tools can implement to opt-in to parallel execution.

### Changes

- **New `Laravel\Ai\Contracts\Concurrent` interface** — empty marker interface following the existing pattern (`HasTools`, `HasStructuredOutput`, `HasMiddleware`, etc.)
- **Modified `AddsToolsToPrismRequests::createPrismTool()`** — checks if the tool implements `Concurrent` and calls `->concurrent()` on the Prism tool
- **Added tests** — verifies concurrent tools get marked, and non-concurrent tools don't

### Usage

```php
use Laravel\Ai\Contracts\Concurrent;

class SearchPlacesTool implements Concurrent, Tool
{
    // Tool implementation...
}
```

When the AI model requests multiple tool calls in a single step, tools implementing `Concurrent` will execute in parallel rather than sequentially — reducing total wait time for I/O-bound operations like API calls, database queries, or file reads.

### How it works

Prism groups tool calls by their concurrency flag. Concurrent tools are batched and executed via `Concurrency::run()` (Laravel's Concurrency facade), while sequential tools run one at a time. Results are returned in the original order regardless of execution timing. Error handling is identical for both — if one concurrent tool fails, others still complete.

### DIFF
--- a/src/Contracts/Concurrent.php
+++ b/src/Contracts/Concurrent.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace Laravel\Ai\Contracts;
+
+interface Concurrent
+{
+    //
+}
--- a/src/Gateway/Prism/Concerns/AddsToolsToPrismRequests.php
+++ b/src/Gateway/Prism/Concerns/AddsToolsToPrismRequests.php
@@ -4,6 +4,7 @@ namespace Laravel\Ai\Gateway\Prism\Concerns;
 
 use Illuminate\JsonSchema\JsonSchemaTypeFactory;
 use Illuminate\Support\Collection;
+use Laravel\Ai\Contracts\Concurrent;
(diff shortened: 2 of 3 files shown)


pushpak1300:
As we have removed the prism implementation. this MR no longer applies. feel free to submit changes with new gateway support.

PR #301, opened 2026-03-23
Fix reasoning data lost when restoring conversation history

### LINKED ISSUE #271: OpenAI 400 error when continuing conversation with reasoning model (gpt-5-mini)
## Description

When continuing a conversation that was previously streamed with a reasoning model (e.g. `gpt-5-mini`), the OpenAI API returns a 400 error:

```
OpenAI Error [400]: invalid_request_error - Item 'rs_...' of type 'reasoning' was provided without its required following item.
```

## Root Cause

The issue is in how Prism's `MessageMap` (`vendor/prism-php/prism/src/Providers/OpenAI/Maps/MessageMap.php`) maps assistant messages containing both text content and tool calls with reasoning IDs.

In `mapAssistantMessage()` (lines 118-162), when an assistant message has **both** text content and tool calls:

1. The text content is pushed first as `{ role: 'assistant', content: [{ type: 'output_text', ... }] }` (line 135)
2. Then reasoning blocks as `{ type: 'reasoning', id: ..., summary: ... }` (line 152)
3. Then function calls as `{ type: 'function_call', ... }` (lines 153-159)

This produces the sequence: `output_text` → `reasoning` → `function_call`

OpenAI's Responses API requires that a `reasoning` item must be **immediately followed** by its associated output (`function_call` or `output`). The `output_text` item being placed **before** the reasoning block breaks this contract.

## Steps to Reproduce

1. Create an agent that uses conversation memory (`RemembersConversations`)
2. Stream a response using `gpt-5-mini` (or any OpenAI reasoning model)
3. The model responds with text + tool calls (with reasoning)
4. Continue the conversation with a second message
5. The SDK replays conversation history → OpenAI returns 400

## Expected Behavior

The reasoning blocks should be ordered correctly when replaying conversation history, respecting OpenAI's requirement that reasoning items are immediately followed by their associated output items.

## Environment

- laravel/ai: v0
- prism-php/prism: latest
- PHP: 8.4
- Model: gpt-5-mini (OpenAI Responses API)

## Stack Trace

```
Laravel\Ai\Exceptions\AiException: OpenAI Error [400]: invalid_request_error - Item 'rs_...' of type 'reasoning' was provided without its required following item.
  at vendor/laravel/ai/src/Gateway/Prism/PrismException.php:63
  → PrismGateway->streamText()
  → StreamsText->stream()
  → StreamableAgentResponse->getIterator()
```

Cbrad24:
Originally introduced in PR #203.

This fixes #271, by ensuring that reasoning_id and reasoning_summary are replayed from database conversations to OpenAI reasoning models.


mohali-id:
Its works very well for me

joelclermont:
I can also confirm that this fixes the issue. We have run into this issue with reasoning models and this solves it.

Any timeline on when it could be merged?

pushpak1300 (approved):

mohali-id:
> I can also confirm that this fixes the issue. We have run into this issue with reasoning models and this solves it.
> 
> 
> 
> Any timeline on when it could be merged?

It still error if function call more then one in one message. Fixed in PR 461

nikitaku11:
This doesn’t fix the issue, because all tool calls are still merged into a single message, while every provider requires two messages per tool call: https://github.com/laravel/ai/issues/402

mohali-id:
> This doesn’t fix the issue, because all tool calls are still merged into a single message, while every provider requires two messages per tool call: #402

This PR #461 actually fix the issue

Cbrad24:
> This doesn’t fix the issue, because all tool calls are still merged into a single message, while every provider requires two messages per tool call: #402

Yes this PR is a month old now, and since then this package has pulled in its own gateway functionality into the package natively, but at the time I also lodged https://github.com/prism-php/prism/pull/971 upstream to fix the other issue in the Prism gateway.

Been running both fixes using `cweagans/composer-patches` without issue since then, but the other PR might be the go now.

PR #312, opened 2026-03-25
Add Gemini gateway for direct API integration

pushpak1300:
The AI SDK currently routes Gemini text generation, embeddings, and image generation through Prism PHP, creating a bottleneck where bug fixes and new features must first be implemented in Prism before the SDK can adopt them. This adds a direct Gemini gateway — following the same pattern as the OpenAI migration in #275 — giving the SDK full ownership of its Gemini integration.

> [!WARNING]
> **Breaking Change:** Gemini now uses the direct AI SDK gateway instead of Prism PHP. Any Prism-specific provider options, middleware, or customizations applied to Gemini text generation will no longer take effect. If you were relying on Prism's Gemini provider behavior (e.g., `withProviderOptions` passed through Prism, Prism-specific tool mapping, or Prism event listeners for Gemini), you will need to migrate to the AI SDK's provider options interface (`HasProviderOptions`) and the direct gateway's configuration.

### Approach

- New `GeminiGateway` implementing the full `Gateway` interface, with text-specific logic in 6 concern traits (same structure as `OpenAiGateway`)
- Talks directly to `generativelanguage.googleapis.com/v1beta/` — `generateContent`, `streamGenerateContent`, and `batchEmbedContents`
- Tool continuations resend full conversation history (Gemini has no `previous_response_id`)
- Uses Gemini 3+ `functionCall.id` when available, with synthetic UUID fallback for older models
- Thinking/reasoning parts (`thought: true`) mapped to reasoning stream events
- Image generation uses `imageConfig` nesting per the Gemini API spec
- No public API changes — backend-only swap in `AiManager::createGeminiDriver()`


TonsiTT:
Great work on this.
Would love too see prompt caching and batch embedding support as well. Seems to be in prism as well

PR #322, opened 2026-03-28
Update prism-php/prism dependency version

rspahni:
Prism has advanced to version 0.100.x.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,7 +21,7 @@
         "illuminate/support": "^12.0|^13.0",
         "laravel/prompts": "^0.3.6",
         "laravel/serializable-closure": "^2.0",
-        "prism-php/prism": "^0.99.0"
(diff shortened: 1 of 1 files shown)


pushpak1300:
There are some breaking changes to this. https://github.com/prism-php/prism/releases/tag/v0.100.0 

We are also diverging away form prism very soon so I will let @taylorotwell decide what to do here.

rspahni:
@pushpak1300 - fair enough. Actually, I had not even been aware that Prism was included in Laravel AI up until Composer didn't let me install AI "over" Prism. I figure the best path forward is to remove Prism from AI's dependencies and let people install it separately if they want to use it. Either way, thx for addressing this.

pushpak1300 (approved):
There are some breaking changes in the api behaviour but we already have migrated off prism for those providers so we have to tacked them separately in different pr.

PR #328, opened 2026-03-29
Fix Mistral transcription with diarize()

NoelDeMartin:
This PR fixes diarization when using the Mistral provider.

Some things to note:

- The Prism dependency also has a bug, I've also opened a PR to fix the issue: https://github.com/prism-php/prism/pull/995
- The "Unknown Speaker" fallback is necessary because until the aforementioned PR is merged in Prism, the `speaker_id` will be null.
- Mistral speaker segments use `speaker_id`, whereas OpenAI use `speaker`. That's why we still need both in the segments mapping.

### DIFF
--- a/src/Gateway/Prism/PrismGateway.php
+++ b/src/Gateway/Prism/PrismGateway.php
@@ -307,12 +307,21 @@ class PrismGateway implements Gateway
                     ),
                 });
 
-            if ($provider->driver() === 'openai') {
-                $request->withProviderOptions(array_filter([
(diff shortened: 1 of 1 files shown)


pushpak1300:
fix it after https://github.com/laravel/ai/pull/368 is merged so we no longer dependent on prism.

NoelDeMartin:
Ok, thanks for the heads up! Opened a new PR #385, so we can close this one.

PR #344, opened 2026-04-01
[0.x] Add UI metadata support for tool responses

chrisnetonline:
When building chat UIs on top of streaming tool calls, there's currently no way to attach UI-only information (e.g. "thinking" labels, status indicators, or display hints) alongside tool results. Everything returned from `Tool::handle()` is collapsed to a single string that goes straight to the model. If a frontend wants to show a "Searching return policies…" label while a tool runs, there's nowhere clean to put it - you end up encoding it into the result string itself, which leaks UI concerns into model context.

The Vercel AI SDK solves this with `toModelOutput` on tools, separating what the model sees from what the UI sees. This PR brings that same concept to the Laravel AI SDK in a minimal, backwards-compatible way.

### Changes

- **New `Laravel\Ai\Tools\ToolResponse` value object** - tools can optionally return `ToolResponse::make($result)->meta([...])` instead of a plain string. The `meta` array is UI-only and never sent to the provider.
- **New `Laravel\Ai\Contracts\HasToolMeta` interface** - optional marker interface for tools that want to declare meta _before_ execution (shows up on the `tool-input-available` event, e.g. as a "thinking" label while the tool runs).
- **`Tool::handle()` return type widened** to `Stringable|string|ToolResponse` - existing tools returning strings are unaffected.
- **`ToolResult` and `ToolCall` DTOs** gain an optional `?array $meta` parameter (defaults to `null`). Serialization shape is unchanged when meta is absent.
- **Vercel streaming protocol** - both `tool-input-available` and `tool-output-available` events include a `meta` key when present, omit it entirely when null:

```json
{"type":"tool-input-available","toolCallId":"fc_abc","toolName":"ReadPolicies","input":{},"meta":{"thinking":"Reading policies…"}}
```

```json
{"type":"tool-output-available","toolCallId":"fc_abc","output":"{...}","meta":{"thinking":"Reading policies…"}}
```

- **`DatabaseConversationStore`** round-trips `meta` through persistence.
- **Provider serialization is untouched** - `buildToolResultsInput`, `serializeToolResultOutput`, and `mapToolResultMessage` only use `->result`, never `->meta`.

### Usage

```php
use Laravel\Ai\Contracts\HasToolMeta;
use Laravel\Ai\Tools\ToolResponse;

class ReadReturnPolicies implements Tool, HasToolMeta
{
    public function toolMeta(): array
    {
        return ['thinking' => 'Reading return policies…'];
    }

    // Or...

    public function handle(Request $request): ToolResponse
    {
        $policies = $this->fetchPolicies($request->get('policy_identifier'));

        return ToolResponse::make(json_encode($policies))
            ->meta(['thinking' => 'Reading return policies…']);
    }

    // ...
}
```

Tools that don't opt in see zero behavioral change — `handle()` can still return a plain string.

### Backwards compatibility

- All constructor changes use trailing optional parameters with defaults.
- `toArray()` / `jsonSerialize()` shapes are identical when `meta` is `null` - existing keys like `result_id` and `reasoning_id` are always present regardless.
- The only internal change is `InvokesTools::executeTool()` returning `array{string, ?array}` instead of `string` (protected method on a trait, not part of the public API).


pushpak1300:
Marking this draft as merge conflicts are there.

chrisnetonline:
> Marking this draft as merge conflicts are there.

fixed the conflict and tweaked some naming to better align with package conventions.

pushpak1300:
I will tackle this feature once we refactor the LoopControl.

Thanks for the PR.

PR #346, opened 2026-04-01
Fix Files::put mimeType and name overrides being dropped

RanaTuhin:
`Files::put($file, mimeType: ..., name: ...)` silently dropped both overrides. The values were forwarded as extra arguments to the file provider but never applied to the underlying file, so gateways still read the original values and uploaded files with the wrong MIME type or filename.

### Usage

```php
Files::put(
    Document::fromStorage('uploads/report', 'local'),
    mimeType: 'application/pdf',
    name: 'quarterly-report.pdf',
);
```

### Approach

- Wrap the file in a lightweight decorator when a `mimeType` override is passed, so the override is transparently returned in place of the detected value.
- Reassign the result of the rename call so immutable `StorableFile` implementations also pick up the new name.
- Existing and third-party `StorableFile` classes require no changes.

### DIFF
--- a/src/Gateway/Gemini/Concerns/BuildsTextRequests.php
+++ b/src/Gateway/Gemini/Concerns/BuildsTextRequests.php
@@ -83,7 +83,16 @@ trait BuildsTextRequests
             'topP' => $options?->topP,
         ]));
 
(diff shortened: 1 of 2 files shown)


taylorotwell:
There is no way to solve this without a wrapper class? Feels odd.

Issue #347, opened 2026-04-02
Refactor: Centralize multi-step tool loop in the SDK core, make gateways single-turn LLM adapters


vinitkadam03:
### Problem

Today, the multi-step tool execution loop (call LLM → detect tool calls → execute tools → call LLM again → repeat) lives **inside each gateway implementation**:

- **PrismGateway** delegates the entire loop to Prism, which runs it internally via `withMaxSteps`. Tool execution, step counting, and loop termination all happen inside Prism.
- **OpenAiGateway** implements its own recursive tool loop in `ParsesTextResponses::processResponse()` (non-streaming) and `HandlesTextStreaming::handleStreamingToolCalls()` (streaming).

This means:

1. **Every new gateway must reimplement the tool loop.** Adding a new direct-API gateway (e.g., for Anthropic's Messages API, or any future provider) requires duplicating the step counting, tool execution, tool result assembly, and loop termination logic.

2. **Cross-cutting tool behaviors must be wired into every gateway separately.** Features like client-executed tools, human-in-the-loop tool approval, tool call filtering, concurrent tool execution, error recovery, and max-step enforcement all need per-gateway implementations. This is the immediate pain point -- adding client-executed tools and tool approval support (see [Prism PR](https://github.com/prism-php/prism/pull/932)) requires touching all gateways (providers in prism) with similar but slightly different code.

3. **The `TextGateway` contract conflates "talk to the LLM" with "orchestrate multi-turn tool loops."** The `generateText()` method accepts `tools` and `maxSteps` and is expected to return a fully-resolved `TextResponse` after N turns. This makes it impossible to control the loop from the SDK core.

4. **PrismGateway is a double-abstraction.** Prism already has its own tool loop, message assembly, step tracking, and response building. The SDK then re-maps all of that (steps, messages, tool calls, tool results) back into its own types. Two loops exist conceptually, but only Prism's actually runs.

### Proposed Architecture

Adopt a provider/core split similar to the Vercel AI SDK:

**Gateways (Providers) = single-turn LLM adapters.** Their only job:
- Accept messages, tools (as schemas), and model config
- Make one HTTP call to the LLM
- Return the raw response mapped to SDK types (text, tool calls, usage, finish reason)
- Handle streaming for a single turn

**SDK Core = orchestration.** A single, central place that:
- Runs the multi-step tool loop (call gateway → check for tool calls → execute tools → append messages → call gateway again)
- Enforces `maxSteps`
[...]

PR #270, opened 2026-03-14
Bedrock provider

tott:
## Add AWS Bedrock Provider

 This PR adds basic support for AWS Bedrock as an AI provider, enabling access to models from Anthropic, Amazon, Meta, Mistral, and more through a unified interface. This is likely not in a place to be fully merged as it requires further testing but it should help people looking for this in reference to #39, and potentially can serve as a starting point for a full-featured implementation

  ### Features

  **Text Generation**
  - Support for Claude, Titan, Llama, Mistral, and other models
  - Streaming via Bedrock's `converseStream` API
  - Tool/function calling
  - Structured output support with JSON validation

  **Embeddings**
  - Amazon Titan Embed Text
  - Configurable dimensions

  **Image Generation**
  - Amazon Nova Canvas
  - Amazon Titan Image Generator
  
  **Speech Transcription**
  - Amazon Nova Sonic
  
 ### Implementation Details

  **Authentication**
  Supports three authentication methods:
  1. IAM credentials (access key, secret key, session token)
  2. Bearer token (`AWS_BEARER_TOKEN_BEDROCK`)
  3. Default AWS credential chain (environment, `~/.aws/credentials`, IAM roles)

  **Tool Calling**
  - Full multi-turn conversation loop with configurable `maxSteps`
  - Automatic tool execution with invocation callbacks
  - Token usage tracking across all tool calls

  **Message Formatting**
  Proper support for:
  - `AssistantMessage` with tool calls
  - `ToolResultMessage` for tool outputs
  - `UserMessage` with document attachments
  - Conversation history maintenance across tool loops

  ### Configuration Example

  ```php
  'bedrock' => [
      'driver' => 'bedrock',
      'region' => env('AWS_BEDROCK_REGION', 'us-east-1'),
      'access_key_id' => env('AWS_ACCESS_KEY_ID'),
      'secret_access_key' => env('AWS_SECRET_ACCESS_KEY'),
      'session_token' => env('AWS_SESSION_TOKEN'),
      'use_default_credential_provider' => env('AWS_USE_DEFAULT_CREDENTIALS', true),
      'models' => [
          'text' => [
              'default' => 'us.anthropic.claude-sonnet-4-5-20250929-v1:0',
              'cheapest' => 'us.anthropic.claude-haiku-4-5-20250929-v1:0',
              'smartest' => 'us.anthropic.claude-opus-4-6-20250929-v1:0',
          ],
          'embeddings' => [
              'default' => 'amazon.titan-embed-text-v2:0',
[...]

PR #349, opened 2026-04-03
Refactor: 🔄 Centralize multi-step tool loop into StepLoop; gateways become single-turn adapters

vinitkadam03:
## Summary

This PR extracts the multi-step tool loop from individual gateways into a central `StepLoop` class, making gateways thin single-turn adapters. The net result is **thousands of lines of duplicated orchestration code** replaced by **shared loop logic** and leaner gateways.

- **`StepLoop`** (`src/Gateway/StepLoop.php`) — new central class that owns the multi-step tool loop (LLM call → tool execution → re-prompt) for both `generate` and `stream` flows. Named after the `Step` DTO it produces on each iteration.
- **`SingleTurnResponse`** (`src/Gateway/SingleTurnResponse.php`) — lean DTO representing one LLM turn, returned by all gateways. Lives under `Laravel\Ai\Gateway` (not `Responses`) because it's an internal gateway↔loop contract, never exposed to userland.
- **`StepContext`** (`src/Gateway/StepContext.php`) — small DTO passed from `StepLoop` to each gateway call, carrying per-turn orchestration hints (currently `stepNumber`, `isFinalStep`). Lets providers consume the orchestrator's terminal-step verdict without making them recompute it themselves. Today only Bedrock reads it; everyone else accepts-and-ignores.
- **All eleven gateways** (OpenAI, AzureOpenAi, Anthropic, Groq, Gemini, Mistral, xAI, DeepSeek, Ollama, OpenRouter, Bedrock) simplified to single-turn adapters — each gateway's `ParsesTextResponses` and `HandlesTextStreaming` traits drop recursive multi-step logic entirely, and Bedrock's internal `while ($step < $maxSteps)` Converse loop / `executeToolCalls` are gone too.
- **`previous_response_id`** threading for OpenAI / AzureOpenAi / xAI — `StepLoop` passes an opaque provider response ID between turns so providers that support stateful continuation can use it instead of replaying the full conversation.
- **`providerExecuted` flag on `ToolCall`** — Anthropic's `server_tool_use` blocks (web_search, web_fetch, code_execution, advisor) are now tagged when parsed and re-emitted as `server_tool_use` on replay. The step loop skips local execution for them via the existing `findTool()` → null path.
- **`providerOptions`** correctly persisted in continuation requests (addresses the fix from #340).

## A note on scope

This is a large architectural change and my first PR to this repo, so understandably there may not be trust built up yet. I'm completely fine with this serving as a **reference implementation** that maintainers can review, cherry-pick from, or rewrite as they see fit.

I raised the proposal in #347 after noticing the same orchestration logic being duplicated across new gateway PRs (#309, #311). 

I've also been contributing to [Prism](https://github.com/echolabsdev/prism) and working on client-executed tools and tool approval there ([prism-php/prism#932](https://github.com/prism-php/prism/pull/932)), which is what originally motivated this refactor - those features would only need to be implemented once in the step loop rather than in every gateway.

Happy to iterate on any feedback.

> **Note:** I have not been able to run the full integration test suite as I don't have API keys for all providers. The unit tests and feature tests with HTTP fakes all pass. Would appreciate the maintainers running the integration suite against this branch.

## Motivation

Before this change, every gateway independently implemented the same recursive multi-step tool orchestration — `ParsesTextResponses` had `processResponse` → `continueWithToolResults` → `processResponse` loops, `HandlesTextStreaming` had parallel `handleStreamingToolCalls` → `processTextStream` recursion, and Bedrock had its own bespoke `while ($step < $maxSteps)` Converse loop with internal `executeToolCalls`. This meant:

- Every new gateway (Anthropic #309, Groq #311, DeepSeek #399, Ollama, OpenRouter, Bedrock #270, AzureOpenAi #404) had to reimplement ~200–400 lines of identical orchestration logic.
- Cross-cutting concerns (telemetry, tool approval, rate limiting around tool calls) would need to be added in every gateway.
- Bug fixes to the tool loop required changes in every gateway.
- Provider-specific behaviors like Anthropic's `server_tool_use` replay, OpenAI's `previous_response_id` continuation, or Bedrock's "force structured-output tool on the final step" trick had to be hand-rolled inside each gateway's bespoke loop.

With orchestration centralized, a new gateway only needs to implement single-turn LLM communication - one `generateText()` and one `streamText()` method. The surface area becomes small enough that providers could realistically be maintained as community packages: anyone could add support for a new LLM by implementing the single-turn contract without understanding or replicating the orchestration logic. Telemetry / observability (e.g. OpenTelemetry, Arize Phoenix) can also be added in a single place - `StepLoop` - without touching any gateway.

## Changes

| Area | What changed |
|------|-------------|
| `TextGateway` contract | `generateText()` returns `SingleTurnResponse`, accepts `?string $previousResponseId` and `?StepContext $context`; `onToolInvocation()` removed |
| `StepLoop` (new) | Owns multi-step tool loop, usage aggregation, structured output handling, `previousResponseId` threading, `StepContext` construction, streaming path |
| `SingleTurnResponse` (new) | DTO returned by every gateway — text, toolCalls, finishReason, usage, meta, structured, responseId |
| `StepContext` (new) | Per-turn orchestration hints (`stepNumber`, `isFinalStep`) passed to gateways. Used by Bedrock; ignored by others |
| `ToolCall` | New `providerExecuted: bool` flag for server-side tool invocations (Anthropic server_tool_use) |
| `OpenAiGateway` | Single-turn; `ParsesTextResponses` drops ~270 lines; `BuildsTextRequests` handles `previous_response_id` |
| `AzureOpenAiGateway` | Single-turn; mirrors the OpenAI port (same trait stack, same `previousResponseId` flow). No behavior change |
| `AnthropicGateway` | Single-turn; `ParsesTextResponses` drops ~260 lines; `HandlesTextStreaming` drops ~220 lines; `providerExecuted` re-emits `server_tool_use` blocks on replay |
| `GroqGateway` | Single-turn; `ParsesTextResponses` drops ~280 lines; `HandlesTextStreaming` drops ~190 lines |
| `GeminiGateway` | Single-turn; `ParsesTextResponses` drops ~210 lines; `HandlesTextStreaming` drops ~150 lines |
| `MistralGateway` | Single-turn; `ParsesTextResponses` drops ~250 lines; `HandlesTextStreaming` drops ~160 lines |
| `XaiGateway` | Single-turn; `ParsesTextResponses` drops ~250 lines; `BuildsTextRequests` handles `previous_response_id` |
| `DeepSeekGateway` | Ported to single-turn contract |
| `OllamaGateway` | Ported to single-turn. Behavioral quirk preserved: force `FinishReason::ToolCalls` when `tool_calls` are populated regardless of `done_reason`, since Ollama can return `"stop"` alongside tool calls |
| `OpenRouterGateway` | Ported to single-turn. Behavioral quirk preserved: `finish_reason: "error"` with inline error payload → emit stream `Error` event |
| `BedrockTextGateway` | Single-turn — internal `while ($step < $maxSteps)` Converse loop, `executeToolCalls`, in-loop conversation builders all removed (~400 lines). Bedrock's Converse API has no native structured-output mode, so the existing synthetic `structured_output` tool injection is preserved. The original "force `toolChoice` to the synthetic tool *only* on the final step of the loop" optimization (which lets the model use real tools on intermediate steps) is preserved by reading `StepContext::$isFinalStep` — the only behavior the single-turn split couldn't infer locally |
| `FakeTextGateway` | Updated to match new contract |
| `GeneratesText` / `StreamsText` | Instantiate `StepLoop` and delegate; tool-invocation events wired via callbacks |
| `StreamEnd` | Gains `?string $responseId` for provider ID propagation through the streaming path |

### DIFF
--- a/src/Gateway/Anthropic/Concerns/HandlesTextStreaming.php
+++ b/src/Gateway/Anthropic/Concerns/HandlesTextStreaming.php
@@ -8,7 +8,6 @@ use Laravel\Ai\Gateway\TextGenerationOptions;
 use Laravel\Ai\Providers\Provider;
 use Laravel\Ai\Responses\Data\FinishReason;
 use Laravel\Ai\Responses\Data\ToolCall;
-use Laravel\Ai\Responses\Data\ToolResult;
 use Laravel\Ai\Responses\Data\UrlCitation;
 use Laravel\Ai\Responses\Data\Usage;
 use Laravel\Ai\Streaming\Events\Citation as CitationEvent;
@@ -23,44 +22,83 @@ use Laravel\Ai\Streaming\Events\TextDelta;
 use Laravel\Ai\Streaming\Events\TextEnd;
 use Laravel\Ai\Streaming\Events\TextStart;
 use Laravel\Ai\Streaming\Events\ToolCall as ToolCallEvent;
-use Laravel\Ai\Streaming\Events\ToolResult as ToolResultEvent;
 
 trait HandlesTextStreaming
 {
     /**
-     * Process an Anthropic streaming response and yield Laravel stream events.
+     * Stream with automatic pause_turn resume — a provider-internal server-tool
+     * continuation that is transparent to the caller / StepLoop.
+     */
+    protected function processTextStreamWithPauseTurnResume(
+        string $invocationId,
+        Provider $provider,
+        string $model,
+        array $requestBody,
+        ?TextGenerationOptions $options,
+        ?int $timeout,
+    ): Generator {
+        $maxResumes = $options?->maxSteps ?? 5;
+
+        for ($depth = 0; $depth < $maxResumes; $depth++) {
+            $response = $this->withErrorHandling(
+                $provider->name(),
+                fn () => $this->client($provider, $timeout)
+                    ->withOptions(['stream' => true])
+                    ->post('messages', $requestBody),
+            );
+
+            $responseContent = [];
+            $stopReason = '';
+            $bufferedStreamEnd = null;
+
+            foreach ($this->processTextStream($invocationId, $provider, $model, $response->getBody(), $responseContent, $stopReason) as $event) {
+                if ($event instanceof StreamEnd) {
+                    $bufferedStreamEnd = $event;
+                } else {
+                    yield $event;
+                }
+            }
+
+            if ($stopReason !== 'pause_turn') {
+                if ($bufferedStreamEnd) {
+                    yield $bufferedStreamEnd;
+                }
+
+                return;
+            }
+
+            $requestBody['messages'][] = [
+                'role' => 'assistant',
+                'content' => $this->ensureToolInputIsObject(array_values($responseContent)),
+            ];
+        }
+    }
+
+    /**
+     * Process an Anthropic streaming response for a single turn and yield Laravel stream events.
      */
     protected function processTextStream(
         string $invocationId,
         Provider $provider,
         string $model,
-        array $tools,
-        ?array $schema,
-        ?TextGenerationOptions $options,
         $streamBody,
-        array $requestBody = [],
-        int $depth = 0,
-        ?int $maxSteps = null,
-        ?int $timeout = null,
+        array &$outResponseContent = [],
+        string &$outStopReason = '',
     ): Generator {
-        $maxSteps ??= $options?->maxSteps;
-
         $messageId = $this->generateEventId();
         $reasoningId = '';
         $streamStartEmitted = false;
         $textStartEmitted = false;
         $reasoningStartEmitted = false;
 
-        $currentText = '';
         $currentBlockType = '';
-        $currentBlockIndex = -1;
-        $currentBlockText = '';
         $currentThinkingText = '';
         $currentSignature = '';
         $currentToolIndex = -1;
         $currentServerToolInput = '';
         $pendingToolCalls = [];
         $responseContent = [];
+        $currentServerToolBlock = [];
 
         $inputTokens = 0;
         $cacheCreationTokens = 0;
@@ -133,11 +171,10 @@ trait HandlesTextStreaming
             if ($type === 'content_block_start') {
                 $blockType = $data['content_block']['type'] ?? '';
                 $currentBlockType = $blockType;
-                $currentBlockIndex = $data['index'] ?? count($responseContent);
+                $index = $data['index'] ?? count($responseContent);
+                $responseContent[$index] = $data['content_block'] ?? [];
 
                 if ($blockType === 'text') {
-                    $currentBlockText = '';
-
                     if ($event = $emitTextStart()) {
                         yield $event;
                     }
@@ -158,12 +195,13 @@ trait HandlesTextStreaming
                     ];
                 } elseif ($blockType === 'server_tool_use') {
                     $currentServerToolInput = '';
+                    $currentServerToolBlock = $data['content_block'] ?? [];
 
                     yield (new ProviderToolEvent(
                         $this->generateEventId(),
-                        $data['content_block']['id'] ?? '',
+                        $currentServerToolBlock['id'] ?? '',
                         $blockType,
-                        $data['content_block'] ?? [],
+                        $currentServerToolBlock,
                         'started',
                         time(),
                     ))->withInvocationId($invocationId);
@@ -178,11 +216,6 @@ trait HandlesTextStreaming
                     ))->withInvocationId($invocationId);
                 }
 
-                // Track content blocks for tool loop replay...
-                if (isset($data['content_block'])) {
-                    $responseContent[$data['index'] ?? count($responseContent)] = $data['content_block'];
-                }
-
                 continue;
             }
 
@@ -193,12 +226,14 @@ trait HandlesTextStreaming
                     $textDelta = (string) ($data['delta']['text'] ?? '');
 
                     if ($textDelta !== '') {
+                        $blockIndex = $data['index'] ?? array_key_last($responseContent);
+                        if ($blockIndex !== null && isset($responseContent[$blockIndex])) {
+                            $responseContent[$blockIndex]['text'] = ($responseContent[$blockIndex]['text'] ?? '').$textDelta;
+                        }
+
                         if ($event = $emitTextStart()) {
                             yield $event;
                         }
-
-                        $currentText .= $textDelta;
-                        $currentBlockText .= $textDelta;
 
                         yield (new TextDelta(
                             $this->generateEventId(),
@@ -252,10 +287,6 @@ trait HandlesTextStreaming
 
             if ($type === 'content_block_stop') {
                 if ($currentBlockType === 'text' && $textStartEmitted) {
-                    if (isset($responseContent[$currentBlockIndex])) {
-                        $responseContent[$currentBlockIndex]['text'] = $currentBlockText;
-                    }
-
                     yield (new TextEnd(
                         $this->generateEventId(),
                         $messageId,
@@ -264,11 +295,6 @@ trait HandlesTextStreaming
 
                     $textStartEmitted = false;
                 } elseif ($currentBlockType === 'thinking' && $reasoningStartEmitted) {
-                    if (isset($responseContent[$currentBlockIndex])) {
-                        $responseContent[$currentBlockIndex]['thinking'] = $currentThinkingText;
-                        $responseContent[$currentBlockIndex]['signature'] = $currentSignature;
-                    }
-
                     yield (new ReasoningEnd(
                         $this->generateEventId(),
                         $reasoningId,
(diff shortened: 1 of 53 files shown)


vinitkadam03:
Anthropic's thinking blocks include a cryptographic signature that must be replayed verbatim in tool use continuations. The old gateway preserved this by replaying raw API content. The orchestrator round-trips through typed DTOs, so the signature needs to be explicitly modeled.

Added reasoningSignature to ToolCall, following the existing reasoningId/reasoningSummary pattern. The Anthropic gateway now extracts thinking text + signature and attaches them to tool calls. MapsMessages emits the full thinking block with signature on replay.

Also fixed a pre-existing gap in DatabaseConversationStore — reasoningId, reasoningSummary, and the new reasoningSignature were being written to the DB but never read back on load.

Longer term, reasoning could be modeled as standalone content parts on AssistantMessage with a provider metadata bag - keeping core types generic while letting each provider carry opaque data like signatures through the round-trip. Happy to take that on as a follow-up.

vinitkadam03:
Just rebased over #409. Thought it was a nice example of what this refactor helps with: that fix needed changes in BuildsTextRequests, continueWithToolResults, and handleStreamingToolCalls within each affected gateway. With the centralized step loop, only BuildsTextRequests needs the fix per gateway since continuation just calls the same request builder again. Fewer places to patch within each gateway for future fixes like this one 🙂

vinitkadam03:
Rebased the branch 

- Dropped prism gateway changes. 

- AzureOpenAi Port
  Simple port - mirrors the OpenAI gateway exactly (same trait stack, previousResponseId flow, parseTextResponse -> SingleTurnResponse). No behavior change.

- Bedrock Port
  What - Single-turn port + new StepContext DTO on the gateway interface.
  
  Why - Bedrock's Converse API has no native structured-output mode, so the gateway injects a synthetic structured_output tool. The original code forced toolChoice to it only on the final step of its internal loop, letting the model use real tools on intermediate steps. With StepLoop now owning the loop, the gateway can no longer infer "is this the final step?" on its own.
  
  How - StepLoop constructs a StepContext { stepNumber, isFinalStep } per call and passes it to the gateway. Bedrock reads isFinalStep to restore the original final-step toolChoice forcing. All other gateways accept-and-ignore — zero behavior change for them.

src/Gateway/OpenAi/Concerns/HandlesTextStreaming.php
```diff
@@ -70,7 +57,7 @@ protected function processTextStream(
 
             if ($type === 'response.created' && ! $streamStartEmitted) {
                 $streamStartEmitted = true;
-                $responseId = $data['response']['id'] ?? '';
+                $responseId = $data['response']['id'] ?? null;
```
vinitkadam03, line 60: switched the default and ?? fallback from '' to null to mirror the non-streaming path's responseId: $data['id'] ?? null in parseTextResponse() and keep "no continuation token" unambiguous.

src/Gateway/OpenAi/Concerns/MapsMessages.php
```diff
@@ -116,4 +116,16 @@ protected function mapToolResultMessage(ToolResultMessage|Message $message, arra
             ];
         }
     }
+
+    /**
+     * Serialize a tool result output value to a string suitable for the API.
+     */
+    protected function serializeToolResultOutput(mixed $output): string
```
vinitkadam03, line 135: Moved here from ParsesTextResponses (was already called by mapToolResultMessage upstream - relied on trait composition); ParsesTextResponses no longer needs it after the loop logic moved to StepLoop, and the new BuildsTextRequests::extractToolResultsInput also calls it, so it now sits in the trait that actually owns outbound-message shaping.

src/Gateway/OpenAi/Concerns/ParsesTextResponses.php
```diff
@@ -422,15 +186,4 @@ protected function mapToolCallsWithReasoning(array $toolCalls, array $reasonings
             $firstReasoning ? $firstReasoning['summary'] ?? null : null,
         ), $toolCalls);
     }
-
-    /**
-     * Combine usage across all steps.
-     */
-    protected function combineUsage(Collection $steps): Usage
```
vinitkadam03, line 427: Moved into StepLoop::buildFinalResponse() — gateways only see one turn now, so usage aggregation belongs with the orchestrator that owns the $steps collection. Same Usage::add reduce, just hoisted up.

src/Gateway/OpenAi/Concerns/ParsesTextResponses.php
```diff
@@ -49,48 +41,13 @@ protected function validateTextResponse(array $data): void
...
-        ?array $schema,
-        Collection $steps,
-        Collection $messages,
-        int $depth = 0,
-        ?int $maxSteps = null,
-        ?TextGenerationOptions $options = null,
-        ?int $timeout = null,
-    ): TextResponse {
-        $responseId = $data['id'] ?? '';
+    ): SingleTurnResponse {
```
vinitkadam03, line 50: Single-turn parsing is pure (array $data → SingleTurnResponse); $tools/$schema/$options/$timeout were only needed for the old recursive continuation path, hence removed. Adding back is a one-line signature change if a parse-time use ever surfaces

src/Gateway/Anthropic/Concerns/HandlesTextStreaming.php
```diff
@@ -22,43 +20,30 @@
...
-        $currentText = '';
         $currentBlockType = '';
-        $currentBlockIndex = -1;
-        $currentBlockText = '';
         $currentThinkingText = '';
         $currentSignature = '';
         $currentToolIndex = -1;
         $pendingToolCalls = [];
-        $responseContent = [];
+        $currentServerToolBlock = [];
```
vinitkadam03: $responseContent was load-bearing for replay — handleStreamingToolCalls sent it back verbatim as the next-turn assistant content. Replay now lives in StepLoop + MapsMessages (rebuilds from AssistantMessage DTOs via reasoningSignature / providerExecuted), so the accumulator + its patching are dead. The only surviving use — pairing content_block_start ↔ content_block_stop for a server_tool_use block — collapses to a scalar ($currentServerToolBlock) since blocks arrive sequentially.

vinitkadam03:
## Review pass — sequence

I wanted to make reviewing this PR as streamlined as possible. Inline comments are left on every diff hunk that needed justification - load-bearing changes are explained on the line they originate from; stylistic drifts (whitespace, import order, comment rewording, helper extraction, identifier renames, default-value churn) were either trimmed in commits on this branch or flagged for follow-up.

**Bedrock and Anthropic are the two that need careful review** - they carry the load-bearing additions:

- **Anthropic** — new `reasoningSignature` / `providerExecuted` fields on `ToolCall`, signed-thinking replay path in `MapsMessages::mapAssistantMessage`, server_tool_use round-tripping in `extractToolCalls`, and the synthetic `output_structured_data` filter that replaces upstream's raw-content replay.
- **Bedrock** — motivated `StepContext.isFinalStep` (the orchestrator hint that lets Bedrock force `toolChoice` on the final turn), `mapFinishReason` is new (upstream surfaced none), and the streaming path was largely rewritten around the single-turn contract.

The other gateways are mostly mechanical mirrors of the OpenAi / Chat-Completions patterns - comments inline if anything stands out.

Self reviewed each gateway against `upstream/0.x` end-to-end, in this order:

1. **Shared scaffolding** — `TextGateway` contract · `StepLoop` · `SingleTurnResponse` · `StepContext` · `FakeTextGateway` · `Providers/Concerns/{GeneratesText,StreamsText}` · `ToolCall` (new fields) · `StreamEnd.responseId` · `DatabaseConversationStore` rehydration · `InvokesTools`
2. **OpenAi** — canonical Responses API gateway
3. **Xai** — Responses API twin of OpenAi
4. **Anthropic** — Messages API; load-bearing `reasoningSignature` / `providerExecuted` plumbing
5. **Groq** — Chat Completions
6. **Gemini** — Google Generative Language API; thinking-parts handling
7. **Mistral** — Chat Completions
8. **DeepSeek** — Chat Completions
9. **Ollama** — NDJSON streaming
10. **OpenRouter** — Chat Completions, meta-provider
11. **AzureOpenAi** — composes OpenAi concerns
12. **Bedrock** — Converse API; motivated `StepContext.isFinalStep`

vinitkadam03:
rebased.

vinitkadam03:
@pushpak1300 Squashed commits and rebased, since there were too many commits with conflicts. But I wanted opinion / confirmation on one change. 

pause_turn handling: gateway-internal loop vs. FinishReason::Continue in StepLoop

pause_turn support was added in #493. In the current upstream implementation, pause_turn resumes share the same $depth counter as tool calls inside the gateway's recursive generateText/streamText. With the StepLoop refactor, there are two ways to handle this:

Option A: Gateway-internal loop (not counted against maxSteps) The Anthropic gateway handles pause_turn internally with its own loop, transparent to StepLoop. Simpler StepLoop, but the user has no visibility or control over these extra API calls — each of which is a real billed request.

Option B: FinishReason::Continue surfaced to StepLoop (counted against maxSteps) Add a FinishReason::Continue case. The gateway returns immediately on pause_turn, and StepLoop appends the assistant message (with providerContentBlocks for verbatim replay) and continues the loop — counting it as a step. This matches the current upstream behavior where pause_turn competes with tool calls for the same $depth budget. The tradeoff is StepLoop needs to know about Continue, but the concept is provider-agnostic — any provider that needs server-side continuation could use it.

I have both implementations working with all tests passing. Since upstream currently counts pause_turn against maxSteps/$depth, I've gone with Option B, but happy to switch if you'd prefer keeping it gateway-internal.

vinitkadam03:
rebased

pushpak1300:
Thanks for contributing to the repo! @vinitkadam03 

This is potentially a large refactor, and we are very interested in doing it soon. However, rather than merging everything in one go, I’ll be cherry-picking the commits, reviewing the implementation carefully, and making sure it aligns with the direction we want to take. We’ll likely merge the changes individually instead of as a single large PR.

Thank you again for your work on this. I’ll make sure to add you as a GitHub co-author for all the commits I make related to this specific issue.

vinitkadam03:
@pushpak1300 Thanks a lot for the response and for considering the refactor seriously.

I completely understand the hesitation around merging a large architectural change at once. I just wanted to mention that many parts of the refactor are fairly interconnected, which is why I opened it as a larger PR, a lot of the benefits only really emerge once the orchestration is centralized in `StepLoop` and gateways become thin single-turn adapters.

This structure also made features like concurrent tool execution very straightforward to add, which I implemented as a small POC:
https://github.com/vinitkadam03/ai/pull/2

I think it also creates a strong foundation for future capabilities like:

- tool approval / human-in-the-loop flows
- UI/client-executed tools (for example interactive forms called by ai as tool calls the UI provides user selection as tool output)
- OpenTelemetry instrumentation e.g. Arize Phoenix (one of the main features that helps us monitor conversations and tool calls and refine things by understanding how users interact with our agents)
- more advanced execution strategies without duplicating gateway logic

Currently, I use a custom fork of Prism with client-executed tools, concurrency, and telemetry implemented, but I’d love to eventually switch fully to the AI SDK. That was one of the main motivations behind raising this PR, since it would serve as the foundation for follow-up contributions in those areas as well.

Happy to help split things up further or adapt the implementation in whatever way makes review and merging easier if that would help. Really looking forward to this refactor and the capabilities it could unlock.

PR #354, opened 2026-04-05
[0.x] Throw early when HasStructuredOutput returns empty schema

### LINKED ISSUE #294: HasStructuredOutput agent with empty schema() crashes with undefined property TextResponse::$structured
### Description

When an agent implements `HasStructuredOutput` but its `schema()` method returns an empty array, the SDK crashes with an `Undefined property` error. The gateway layer and the response-wrapping layer disagree on whether to use structured output mode.

### Steps to Reproduce

```php
use Laravel\Ai\Promptable;

class MyAgent implements Agent, HasStructuredOutput
{
    use Promptable;

    {
    }

    {
        return []; // Empty schema
    }
}

// Trigger the bug:
$agent = new MyAgent;
$response = $agent->prompt('Hello');
```

### Expected Behavior

An `InvalidArgumentException` thrown early with a clear message explaining that agents implementing `HasStructuredOutput` must return a non-empty schema.

### Actual Behavior

```
Undefined property: Laravel\Ai\Responses\TextResponse::$structured
```

### Root Cause

There is a mismatch between two independent checks for structured output mode:

1. **Gateway layer** (`PrismGateway::generateText` and `CreatesPrismTextRequests`) uses `! empty($schema)` to decide whether to create a structured or text request and return a `StructuredTextResponse` or `TextResponse`.

2. **Response-wrapping layer** (`GeneratesText::prompt`) uses `$agent instanceof HasStructuredOutput` to decide whether to wrap the response in `StructuredAgentResponse` (accessing `$response->structured`) or `AgentResponse`.

When `schema()` returns `[]`:
- `empty([])` is `true`, so the gateway creates a text request and returns a `TextResponse`
- `$agent instanceof HasStructuredOutput` is still `true`, so the wrapper tries to access `$response->structured` on a `TextResponse` that does not have that property

### Suggested Fix

Throw early in `GeneratesText::prompt()` after resolving the schema, before calling the gateway:

```php
$schema = $agent instanceof HasStructuredOutput
    ? $agent->schema(new JsonSchemaTypeFactory)
    : null;

if ($agent instanceof HasStructuredOutput && empty($schema)) {
    throw new \InvalidArgumentException(
        'Agent ['.get_class($agent).'] implements HasStructuredOutput but returned an empty schema. '
        .'Either define a schema or remove the HasStructuredOutput interface.'
    );
}
```

This is consistent with how `StreamsText::stream()` already rejects `HasStructuredOutput` agents early with an `InvalidArgumentException`.

### Versions

- `laravel/ai`: v0.3.2
- PHP: 8.5
- Laravel: 12.x

sumaiazaman:
Fixes #294

## Summary

When an agent implements `HasStructuredOutput` but `schema()` returns an empty array, the SDK crashes with `Undefined property: TextResponse::$structured`.

### Root Cause

The gateway layer checks `! empty($schema)` to decide whether to return a `StructuredTextResponse` or plain `TextResponse`. But the response-wrapping layer checks `$agent instanceof HasStructuredOutput` — which is `true` regardless of schema contents. When the schema is empty:

1. Gateway returns `TextResponse` (no `->structured` property)
2. Response layer assumes `StructuredTextResponse` and accesses `->structured`
3. Crash

### The Fix

Validate the schema before calling the gateway. If an agent implements `HasStructuredOutput` but returns an empty schema, throw a clear `InvalidArgumentException`:

```php
if ($agent instanceof HasStructuredOutput && empty($schema)) {
    throw new InvalidArgumentException(
        'Agents implementing HasStructuredOutput must return a non-empty schema from the schema() method.'
    );
}
```

This turns an obscure `Undefined property` crash into a clear, actionable error message.

### Changes

- `src/Providers/Concerns/GeneratesText.php` — Validate schema before gateway call

### DIFF
--- a/src/Providers/Concerns/GeneratesText.php
+++ b/src/Providers/Concerns/GeneratesText.php
@@ -56,13 +56,23 @@ trait GeneratesText
 
                 $this->listenForToolInvocations($invocationId, $agent);
 
+                $schema = $agent instanceof HasStructuredOutput
+                    ? $agent->schema(new JsonSchemaTypeFactory)
+                    : null;
+
+                if ($agent instanceof HasStructuredOutput && empty($schema)) {
+                    throw new \InvalidArgumentException(
+                        'Agents implementing HasStructuredOutput must return a non-empty schema from the schema() method.'
+                    );
+                }
+
                 $response = $this->textGateway()->generateText(
                     $this,
                     $prompt->model,
                     (string) $agent->instructions(),
                     $messages,
                     $agent instanceof HasTools ? $agent->tools() : [],
-                    $agent instanceof HasStructuredOutput ? $agent->schema(new JsonSchemaTypeFactory) : null,
+                    $schema,
                     TextGenerationOptions::forAgent($agent),
                     $prompt->timeout,
                 );


pushpak1300:
Closing in favour of https://github.com/laravel/ai/pull/391/. 

I don't think it should throw error on this.

PR #355, opened 2026-04-05
[0.x] Fix PrismTool dropping arguments when provider uses 'input' key

### LINKED ISSUE #291: PrismTool: tool call arguments not mapped when provider uses 'input' key instead of 'arguments'
## Description

`PrismTool::toLaravelToolCall()` assumes tool call arguments are always under the `arguments` key. Some tool calls arrive with `input` as the key instead. When `arguments` is absent and `input` is present, the call to `normalizeArguments()` falls through to an empty array `[]`, silently dropping all arguments.

## Current behaviour

```php
static::normalizeArguments($toolCall['arguments']['schema_definition'] ?? $toolCall['arguments'] ?? []),
```

When the tool call array has `input` instead of `arguments`, this evaluates to `normalizeArguments([])` and all arguments are lost.

## Expected behaviour

The mapping should check `input` as a fallback when `arguments` is not present.

## Suggested fix

```php
static::normalizeArguments(is_array($arguments = ($toolCall['arguments'] ?? $toolCall['input'] ?? [])) ? ($arguments['schema_definition'] ?? $arguments) : $arguments),
```

## Regression test

```php
$toolCall = PrismTool::toLaravelToolCall([
    'id' => 'call_123',
    'name' => 'parallel_subagents',
    'input' => '{"tasks":[{"sub_agent_id":7,"prompt":"Assess risk"}]}',
]);

// Before fix: $toolCall->arguments === []
// After fix:  $toolCall->arguments === ['tasks' => [...]]
```

## Environment

- `laravel/ai` v0.3.0
- PHP 8.4

sumaiazaman:
Fixes #291

## Summary

Some AI providers (notably Anthropic) send tool call arguments under the \`input\` key instead of \`arguments\`. \`PrismTool::toLaravelToolCall()\` and \`toLaravelToolResult()\` only check for \`arguments\`, silently dropping all arguments when \`input\` is used.

### The Bug

```php
// Tool call from Anthropic uses 'input' key
$toolCall = [
    'id' => 'call_123',
    'name' => 'search',
    'input' => '{"query": "Laravel AI"}',  // 'input' not 'arguments'
];

$result = PrismTool::toLaravelToolCall($toolCall);
$result->arguments; // [] — all arguments silently lost
```

### The Fix

Check \`input\` as a fallback when \`arguments\` is not present:

```php
// Before
$toolCall['arguments']['schema_definition'] ?? $toolCall['arguments'] ?? []

// After
$arguments = $toolCall['arguments'] ?? $toolCall['input'] ?? [];
```

Applied to both \`toLaravelToolCall()\` and \`toLaravelToolResult()\`.

### Consistency

This follows the same pattern used in PR #326 which fixed the identical issue in \`DatabaseConversationStore\`.

### Changes

- \`src/Gateway/Prism/PrismTool.php\` — Add \`input\` key fallback in both methods

### DIFF
--- a/src/Gateway/Prism/PrismTool.php
+++ b/src/Gateway/Prism/PrismTool.php
@@ -40,10 +40,12 @@ class PrismTool extends Tool
     public static function toLaravelToolCall(PrismToolCall|array $toolCall): ToolCall
     {
         if (is_array($toolCall)) {
+            $arguments = $toolCall['arguments'] ?? $toolCall['input'] ?? [];
+
             return new ToolCall(
                 $toolCall['id'] ?? '',
                 $toolCall['name'] ?? '',
-                static::normalizeArguments($toolCall['arguments']['schema_definition'] ?? $toolCall['arguments'] ?? []),
+                static::normalizeArguments(is_array($arguments) ? ($arguments['schema_definition'] ?? $arguments) : $arguments),
                 $toolCall['resultId'] ?? null,
                 $toolCall['reasoningId'] ?? null,
                 $toolCall['reasoningSummary'] ?? null,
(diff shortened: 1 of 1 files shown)


pushpak1300:
No longer valid.

rezaulhreza:
> No longer valid.

What version fixes it? Please leave the  reference

pushpak1300:
https://github.com/laravel/ai/releases/tag/v0.6.3 we removed prism form core.

PR #358, opened 2026-04-07
Add truncate option to embeddings

### LINKED ISSUE #71: [Feature] Handling long texts in embeddings
First up - amazing SDK. Love using it. Beautiful!

I noticed that when embedding text that exceeds the model's context window (e.g. 8k tokens for Nomic/Jina via Ollama), the SDK currently throws a raw provider error (for example, `api_error - the input length exceeds the context length`).

It would be helpful to have:
- Option to auto-truncate input text with a warning
- Or: built-in chunking support that returns multiple embeddings

For reference, I built an embeddings package a while ago which used a dedicated embeddings table with a polymorphic model relation that supported multiple embeddings per model: https://github.com/5am-code/ada-laravel
I actually want to archive it now in favor of the official SDK - which is why I'd love to see this feature here! 

Let me know what you think, and if I can help.

WoutervdBrink:
Most embeddings provider allow setting whether the provider truncates documents to the model's context window (default) or to raise an error when the document size exceeds the model's context window.

This commit adds a `truncate` argument to embeddings, which, if supported by the provider, sets whether the input should be silently truncated when it is to large (when set to `true`), or if the provider should raise an error in that case (when set to `false`).

All providers offering this option truncate by default, i.e. the behaviour of embeddings without the truncate option set does not change with this commit.

To the best of my knowledge, the embeddings providers for which I didn't implement setting the truncate option also do not offer this option in their API -- they silently truncate by default, which fortunately still aligns with the default value of the argument (`true`).

Related: #71 (first requested this option).


pushpak1300:
can you please fix merge conflicts ?

WoutervdBrink:
@pushpak1300 Apologies, I'm just now seeing your reply. I think there are no merge conflicts anymore with your changes?

taylorotwell:
Has conflicts.

pushpak1300:
Hey @WoutervdBrink, thanks a lot for all the work on this MR!

Since #555 landed, embeddings now support a generic `providerOptions` parameter that already covers this use case without adding a dedicated `truncate` argument. You can pass the truncation knob per provider:

**Enable truncation (silently truncate oversized inputs):**

```php
// Cohere — 'END' / 'START' / 'RIGHT' (for embed-v4)
Embeddings::for($chunks)
    ->providerOptions(['truncate' => 'END'])
    ->generate(provider: 'cohere', model: 'embed-v4.0');

// Voyage AI
Embeddings::for($chunks)
    ->providerOptions(['truncation' => true])
    ->generate(provider: 'voyageai', model: 'voyage-3');

// Jina
Embeddings::for($chunks)
    ->providerOptions(['truncate' => true])
    ->generate(provider: 'jina');
```

**Disable truncation (raise an error on oversized inputs):**

```php
// Cohere
Embeddings::for($chunks)
    ->providerOptions(['truncate' => 'NONE'])
    ->generate(provider: 'cohere', model: 'embed-v4.0');

// Voyage AI
Embeddings::for($chunks)
    ->providerOptions(['truncation' => false])
    ->generate(provider: 'voyageai', model: 'voyage-3');

// Jina
Embeddings::for($chunks)
    ->providerOptions(['truncate' => false])
    ->generate(provider: 'jina');
```

This keeps the public API surface small and lets each provider's native semantics through (e.g. Cohere's `END` / `START` / `RIGHT` / `NONE`) instead of flattening them to a boolean.

Given that, I think we don't need a separate `truncate` argument here. Closing this in favor of `providerOptions` but really appreciate the effort and the detailed write-up! 🙏

PR #364, opened 2026-04-09
Switch Gemini structured output to response_json_schema

### LINKED ISSUE #139: Gemini structured output fails with nested `additionalProperties` in schema
## Description

When using `HasStructuredOutput` agents with a Gemini provider, structured output requests fail with:

```
Gemini Error [400]: INVALID_ARGUMENT - Invalid JSON payload received. 
Unknown name "additionalProperties" at 'generation_config.response_schema.properties[0].value.items': Cannot find field.
```

Gemini does not support `additionalProperties` in its response schema. Prism's `SchemaMap` strips it at the top level and recursively for Prism-native schema objects, but the Laravel AI SDK bypasses this recursion.

## Root Cause

In `CreatesPrismTextRequests::withStructuredOutputOptions()`, the agent's schema array is wrapped in `new ObjectSchema($schema)`, which serializes the entire schema tree to a **flat array** via `Illuminate\JsonSchema\Types\ObjectType::toArray()`.

When Prism's Gemini `SchemaMap` receives this, it can only strip `additionalProperties` from the top-level array (line 28). The recursive stripping (lines 60-73) relies on traversing Prism-native `Schema` objects via `property_exists($this->schema, 'items')` and `$this->schema instanceof ObjectSchema` — but since the Laravel AI `Schema` serializes to a plain array, these checks fail and nested `additionalProperties` passes through to Gemini.

## Reproduction

Any agent using `HasStructuredOutput` with nested objects in the schema and targeting a Gemini model:

```php
class MyAgent implements Agent, HasStructuredOutput
{
    {
        return [
            'items' => $schema->array()->items(
                $schema->object([
                    'title' => $schema->string()->required(),
                    'score' => $schema->number()->required(),
                ])->withoutAdditionalProperties()
            )->required(),
        ];
    }
}
```

Works with OpenAI (which requires `additionalProperties: false`), fails with Gemini (which rejects it).

## Suggested Fix

In `CreatesPrismTextRequests::withStructuredOutputOptions()` or in `ObjectSchema::toArray()`, recursively strip `additionalProperties` from the serialized schema array when the target provider is Gemini. The provider is already available in `withStructuredOutputOptions()` since it receives `$provider`.

## Versions

- `laravel/ai`: v0.1.3
- `prism-php/prism`: v0.99.19
- PHP: 8.4

ipalaus:
## Summary

Switches Gemini structured output from `response_schema` to `response_json_schema`, fixing 400 errors caused by unsupported JSON Schema features in nested schemas.

`response_schema` uses Gemini's OpenAPI subset, which does not support keys like `additionalProperties` or union types (`["string", "null"]`). `response_json_schema` accepts standard JSON Schema natively, so no stripping or workarounds are needed.

## What this fixes

**Nested objects** — `ObjectSchema::toSchema()` recursively adds `additionalProperties: false` to every nested object (required by OpenAI). With `response_schema`, `Arr::except()` only stripped it from the top level, so any nested objects caused Gemini to reject the request with HTTP 400: `Unknown name "additionalProperties"`. For example:

```php
{
    return [
        'elements' => $schema->array()->required()->items(
            $schema->object([
                'atomicNumber' => $schema->integer()->required(),
                'symbol' => $schema->string()->required(),
            ])
        ),
    ];
}
```

**Nullable types** — `->nullable()` serializes as `"type": ["number", "null"]` (JSON Schema union type). `response_schema` rejects this because it expects a single string type. This forced workarounds like sentinel values (`-1`) to represent null. For example:

```php
'meltingPoint' => $schema->number()->nullable()->required(),
```

Both issues are resolved by switching to `response_json_schema`, which accepts standard JSON Schema as-is.

## Verification

Confirmed working in production by pointing `composer.json` at this branch — nested objects and nullable types both work without workarounds.

All existing and new tests pass.

## Test plan

- [x] `test_structured_output_uses_response_json_schema` — flat schema uses new key
- [x] `test_nested_structured_output_uses_response_json_schema` — array of objects uses new key
- [x] `test_nullable_schema_types_are_preserved_in_response_json_schema` — union types preserved
- [x] Full Gemini test suite passes (21 tests)

### DIFF
--- a/src/Gateway/Gemini/Concerns/BuildsTextRequests.php
+++ b/src/Gateway/Gemini/Concerns/BuildsTextRequests.php
@@ -2,7 +2,6 @@
 
 namespace Laravel\Ai\Gateway\Gemini\Concerns;
 
-use Illuminate\Support\Arr;
 use Laravel\Ai\Enums\Lab;
 use Laravel\Ai\Gateway\TextGenerationOptions;
 use Laravel\Ai\ObjectSchema;
@@ -74,7 +73,7 @@ trait BuildsTextRequests
 
         if (filled($schema)) {
             $generationConfig['response_mime_type'] = 'application/json';
-            $generationConfig['response_schema'] = $this->buildResponseSchema($schema);
+            $generationConfig['response_json_schema'] = $this->buildResponseSchema($schema);
         }
 
         if (! is_null($options?->maxTokens)) {
@@ -127,8 +126,6 @@ trait BuildsTextRequests
      */
     protected function buildResponseSchema(array $schema): array
     {
-        $objectSchema = new ObjectSchema($schema);
-
-        return Arr::except($objectSchema->toSchema(), ['additionalProperties', 'name']);
+        return (new ObjectSchema($schema))->toSchema();
     }
 }
--- a/tests/Feature/Agents/NestedStructuredAgent.php
+++ b/tests/Feature/Agents/NestedStructuredAgent.php
@@ -1,0 +1,36 @@
+<?php
+
+namespace Tests\Feature\Agents;
+
+use Illuminate\Contracts\JsonSchema\JsonSchema;
+use Laravel\Ai\Contracts\Agent;
+use Laravel\Ai\Contracts\HasStructuredOutput;
+use Laravel\Ai\Promptable;
+
+class NestedStructuredAgent implements Agent, HasStructuredOutput
+{
+    use Promptable;
+
+    /**
+     * Get the instructions that the agent should follow.
+     */
+    public function instructions(): string
+    {
+        return 'You are a helpful assistant that knows about periodic table elements.';
+    }
+
+    /**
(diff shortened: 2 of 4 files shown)


ipalaus:
@pushpak1300 don't merge it yet, I think it can be fixed on a more elegant way and also fix a Gemini bug when using `nullable()` in the schema definition. Testing a solution with my code

ipalaus:
I've updated the approach, instead of recursively stripping additionalProperties and name, I switched from response_schema to response_json_schema. This accepts standard JSON Schema natively, so no stripping is needed at all.

This also fixes a second bug: `->nullable()` serializes as `"type": ["number", "null"]` which response_schema rejects. With response_json_schema, nullable types work out of the box.

Tested in our production code, both nested objects and nullable types work without workarounds. Ready for review :)

pushpak1300 (approved):

PR #365, opened 2026-04-09
Fix Anthropic system prompt caching

### LINKED ISSUE #119: Support Anthropic prompt caching (cache_control on system messages)

```
if (! empty($instructions)) {
}
```

Impact

Suggestion

```

if (! empty($instructions)) {
}

#[CacheSystemPrompt]

```

pushpak1300:
Anthropic prompt caching needs the system prompt to be sent with cache metadata on the system block so agents can opt in to cached prompts.

Fixes #119

### Approach

- Format Anthropic system prompts as cached text blocks when provider options request `cache_control`
- Remove `cache_control` from the top-level Anthropic provider options so it only lands on the system prompt block
- Add coverage for the initial request and tool-loop follow-up requests

### DIFF
--- a/src/Gateway/Anthropic/Concerns/BuildsTextRequests.php
+++ b/src/Gateway/Anthropic/Concerns/BuildsTextRequests.php
@@ -27,13 +27,22 @@ trait BuildsTextRequests
             'max_tokens' => $options?->maxTokens ?? 64_000,
         ];
 
+        $providerOptions = $options?->providerOptions(Lab::Anthropic) ?? [];
+        $cacheControl = $providerOptions['cache_control'] ?? null;
+
+        unset($providerOptions['cache_control']);
+
         if (filled($instructions)) {
-            $body['system'] = $instructions;
+            $body['system'] = filled($cacheControl)
+                ? [[
+                    'type' => 'text',
+                    'text' => $instructions,
+                    'cache_control' => $cacheControl,
+                ]]
+                : $instructions;
         }
 
         $mappedTools = filled($tools) ? $this->mapTools($tools, $provider) : [];
-
-        $providerOptions = $options?->providerOptions(Lab::Anthropic) ?? [];
 
         if (filled($schema) && $this->supportsNativeStructuredOutput($provider)) {
             $body['output_config'] = [
--- a/tests/Feature/Providers/Anthropic/ProviderOptionsTest.php
+++ b/tests/Feature/Providers/Anthropic/ProviderOptionsTest.php
@@ -3,9 +3,15 @@
 namespace Tests\Feature\Providers\Anthropic;
 
 use Illuminate\Support\Facades\Http;
+use Laravel\Ai\Contracts\Agent;
+use Laravel\Ai\Contracts\HasProviderOptions;
+use Laravel\Ai\Contracts\HasTools;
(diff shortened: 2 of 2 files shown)


pushpak1300:
## Behavior difference from Prism's approach

Worth noting this PR takes a different approach to `cache_control` than Prism PHP does:

**Prism** passes `cache_control` as a **top-level key** in the Anthropic request payload (automatic caching — Anthropic decides what to cache across system, tools, and messages). It does not strip or move it to the system block. See `prism-php/prism` → `Providers/Anthropic/Handlers/Text.php:91`.

**This PR** strips `cache_control` from provider options and applies it as an **explicit cache breakpoint** on the system prompt block (`[{type: text, text: ..., cache_control: ...}]`).

These are two different Anthropic features:
- **Top-level `cache_control`** = automatic caching (API auto-decides what to cache)
- **Per-block `cache_control`** = explicit cache breakpoints (developer controls what gets cached)

Implications:
1. Users cannot opt into Anthropic's automatic caching through this gateway — `cache_control` is always consumed and applied to the system block only.
2. Only the system prompt gets the cache breakpoint. Automatic caching would also cover tools and message history.

This is likely the right call for most users (explicit system prompt caching is the common use case), but wanted to document the difference in case it matters down the line.

pushpak1300:
## Usage: Automatic vs Explicit Caching

### What this PR enables (explicit cache breakpoints)

The user sets `cache_control` in provider options, and the gateway applies it as a breakpoint on the system prompt block:

```php
// Agent definition
{
    use Promptable;

    {
        return 'You are a helpful assistant with a very long system prompt...';
    }

    {
        return match ($provider) {
            Lab::Anthropic => [
                'cache_control' => ['type' => 'ephemeral'],
            ],
            default => [],
        };
    }
}
```

This sends to Anthropic:

```json
{
  "model": "claude-sonnet-4-6",
  "system": [
    {
      "type": "text",
      "text": "You are a helpful assistant with a very long system prompt...",
      "cache_control": { "type": "ephemeral" }
    }
  ],
  "messages": [...]
}
```

Only the system prompt is cached. You control exactly what gets the breakpoint.

---

### What Anthropic's top-level automatic caching looks like (not available through this gateway)

If `cache_control` were passed as a top-level key instead (like Prism does), the request would be:

```json
{
  "model": "claude-sonnet-4-6",
  "cache_control": { "type": "ephemeral" },
  "system": "You are a helpful assistant with a very long system prompt...",
  "messages": [...],
  "tools": [...]
}
```

With automatic caching, Anthropic's API decides what to cache — it can cache the system prompt, tools, and message history together. The developer doesn't place breakpoints manually.

---

### Summary

| | Explicit (this PR) | Automatic (top-level) |
|---|---|---|
| What gets cached | Only the system prompt | System + tools + messages (API decides) |
| Developer control | Full — you choose the breakpoint | None — API optimizes automatically |
| How it's sent | `cache_control` inside the system content block | `cache_control` as a top-level request key |
| Available via this gateway | ✅ Yes | ❌ No (stripped and redirected to system block) |

If we want to support both in the future, we could use a separate key (e.g., `auto_cache_control`) for the top-level behavior, or only apply the per-block transform when the value includes a signal like `'apply_to' => 'system'`.

PR #368, opened 2026-04-09
Add Mistral gateway for direct API integration

pushpak1300:
The AI SDK currently depends on Prism PHP as a middleware layer for Mistral text generation, embeddings, and transcription. This creates a bottleneck where any bug fix or new feature must first land in Prism before the SDK can adopt it. This replaces the Prism dependency for Mistral with a direct gateway, giving the SDK full ownership of its Mistral integration.

### Approach

- Implements a new `MistralGateway` that calls the Mistral Chat Completions API directly, following the same trait-based composition pattern established by the OpenAI and Groq gateway migrations
- Handles synchronous text generation, streaming with SSE parsing, multi-round tool call loops with full conversation history replay, structured output via JSON schema, image and document attachments, and provider options forwarding
- Migrates embeddings (`/embeddings`) and audio transcription (`/audio/transcriptions`) inline in the gateway, removing the last remaining Prism dependencies for Mistral
- Fixes a bug where `max_tokens`, `temperature`, and `timeout` were dropped from follow-up requests during tool call loops (same class of issue as #363)
- Updates `MistralProvider` to lazily initialize its own gateway, removing the `PrismGateway` dependency from `AiManager`
- No public API changes — the `Ai` facade, agents, tools, and all user-facing interfaces remain identical

https://docs.mistral.ai/api

### DIFF
--- a/config/ai.php
+++ b/config/ai.php
@@ -99,6 +99,7 @@ return [
         'mistral' => [
             'driver' => 'mistral',
             'key' => env('MISTRAL_API_KEY'),
+            'url' => env('MISTRAL_URL', 'https://api.mistral.ai/v1'),
         ],
 
         'ollama' => [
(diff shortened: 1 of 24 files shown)


MikeElghali:
Merci!

PR #371, opened 2026-04-10
Support system prompt caching via system_prompt_cache_type provider option

### LINKED ISSUE #119: Support Anthropic prompt caching (cache_control on system messages)

GioChocolateBro:
Fixes #119

### Problem

Anthropic's prompt caching can dramatically reduce input token costs for agents with large system prompts. However, `GeneratesText` casts `instructions()` to `(string)` before passing to the gateway, so there's no way to attach `cache_control` metadata through the agent interface — even though Prism fully supports it via `SystemMessage::withProviderOptions()`.

### Solution

Add a `system_prompt_cache_type` provider option that mirrors the existing `tool_result_cache_type` pattern. When set, the gateway wraps the instructions string in a `SystemMessage` with the appropriate cache control before passing to Prism.

This is a single new method on `PrismGateway` — `toSystemPrompt()` — called from both `generateText()` and `streamText()`. When no `system_prompt_cache_type` is configured, the plain string is passed through unchanged (zero breaking changes).

### Usage

Agents opt in by implementing `HasProviderOptions`:

```php
use Laravel\Ai\Enums\Lab;

class MyAgent implements Agent, HasProviderOptions, HasTools
{
    {
        return match ($provider) {
            Lab::Anthropic, 'anthropic' => [
                'system_prompt_cache_type' => 'ephemeral',
                'tool_result_cache_type' => 'ephemeral', // existing, unchanged
            ],
            default => [],
        };
    }
}
```

### Why not strip `cache_control` from provider options?

PR #365 took that approach and was closed. Stripping `cache_control` from provider options would break the existing `tool_result_cache_type` mechanism — Prism's `MessageMap::map()` reads that from provider options to apply cache breakpoints on tool results between agent steps. This is critical for multi-step agents where large tool outputs (SQL results, API responses) need caching between iterations.

This PR keeps both mechanisms independent:
- **System prompt caching** — new `system_prompt_cache_type` key, applied in the gateway
- **Tool result caching** — existing `tool_result_cache_type` key, flows through Prism's `MessageMap` untouched

### Results

On a multi-step analytics agent with a ~15k token system prompt:

| | Prompt tokens (full price) | Cache read | Cache write |
|---|---|---|---|
| Before | 37,685 | 0 | 0 |
| After (warm cache) | 334 | 35,316 | 496 |

~90% reduction in effective input token cost.

### Changes

- `PrismGateway.php` — added `toSystemPrompt()` method, updated `generateText()` and `streamText()` to use it
- 2 new imports (`Lab`, `SystemMessage`)
- No interface changes, no new files, no breaking changes

### DIFF
--- a/src/Gateway/Prism/PrismGateway.php
+++ b/src/Gateway/Prism/PrismGateway.php
@@ -15,6 +15,7 @@ use Laravel\Ai\Contracts\Providers\EmbeddingProvider;
 use Laravel\Ai\Contracts\Providers\ImageProvider;
 use Laravel\Ai\Contracts\Providers\TextProvider;
 use Laravel\Ai\Contracts\Providers\TranscriptionProvider;
+use Laravel\Ai\Enums\Lab;
 use Laravel\Ai\Files\File;
 use Laravel\Ai\Files\Image as ImageFile;
 use Laravel\Ai\Files\LocalImage;
@@ -36,6 +37,7 @@ use Prism\Prism\Exceptions\PrismException as PrismVendorException;
 use Prism\Prism\Facades\Prism;
 use Prism\Prism\ValueObjects\Media\Audio;
 use Prism\Prism\ValueObjects\Media\Image as PrismImage;
+use Prism\Prism\ValueObjects\Messages\SystemMessage;
 
 class PrismGateway implements Gateway
 {
@@ -71,7 +73,9 @@ class PrismGateway implements Gateway
         ];
 
         if (! empty($instructions)) {
-            $request->withSystemPrompt($instructions);
+            $request->withSystemPrompt(
+                $this->toSystemPrompt($instructions, $provider, $options)
+            );
         }
 
         if (count($tools) > 0) {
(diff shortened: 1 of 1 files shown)


Hesesses:
Would really appreciate getting this merged 🙏

vinitkadam03:
The prism gateway is going away soon it seems. so I doubt this will get merged.

ahoiroman:
That would reduce my bill by such a huge factor - at the moment, this is a blocker for me, so I'd be super happy if this gets merged.

@vinitkadam03: Where can I read more about the plans of getting rid of the prism gateway?

pushpak1300:
Hey closing this as we have migrated form prism for most of the provider. can you send new pull request, please ?

GioChocolateBro:
@pushpak1300 tool caching is lost with the new anthropic gateway right? Or missing something? It's small potatoes compared to system prompt caching, but worth asking.

Will have a new PR soon.

PR #379, opened 2026-04-10
Refactor tests with rector-pest

MrPunyapal:
This PR has a refactor of multiple expect calls into one chain.


pushpak1300:
Thanks.

PR #383, opened 2026-04-10
Add tool call assertions for agent responses

yousefkadah:
When building agents that use tools, there's no way to assert in tests that the right tools were called with the right arguments. This PR adds three assertion methods that follow the same pattern as the existing prompt assertions.

### Usage

```php
SalesCoach::fake([
    (new TextResponse('Here are the results...', new Usage, new Meta))
        ->withToolCallsAndResults(collect([
            new ToolCall('1', 'search_web', ['query' => 'Laravel pricing']),
        ]), collect()),
]);

(new SalesCoach)->prompt('Research Laravel pricing');

SalesCoach::assertCalledTool('search_web');
SalesCoach::assertCalledTool('search_web', ['query' => 'Laravel pricing']);
SalesCoach::assertCalledTool('search_web', fn (array $args) => str_contains($args['query'], 'Laravel'));
SalesCoach::assertDidNotCalledTool('delete_records');
SalesCoach::assertCalledNoTools();

// Also available via the Ai facade
Ai::assertAgentCalledTool(SalesCoach::class, 'search_web', ['query' => 'Laravel pricing']);
```

### Approach

- Tool calls are recorded in the fake infrastructure when a faked agent prompt completes, using the same middleware that already records prompts. This means assertions don't require the response object — they work the same way as `assertPrompted`, `assertNeverPrompted`, etc.
- `assertCalledTool` accepts an optional second argument: an array for subset matching (including nested), or a closure for custom logic.
- All three methods are available as static helpers on any agent class using `Promptable`, and via `Ai::` for cases where the agent class isn't directly accessible.


pushpak1300:
Thanks for your pull request!

Right now, you can assert tool usage by listening for the `InvokingTool` and `ToolInvoked` events in your tests

Issue #380, opened 2026-04-10
DatabaseConversationStore rehydrates every stored message as plain text and drops attachments/tool context


ruttydm:
I hit this on `laravel/ai` `v0.1.5` while testing remembered conversations.

`DatabaseConversationStore::getLatestConversationMessages()` currently maps every stored row back to a plain `Laravel\Ai\Messages\Message`:

```php
->map(fn ($m) => new Message($m->role, $m->content));
```

But the store writes richer data than that:

- user message attachments are stored in `attachments`
- assistant tool calls are stored in `tool_calls`
- tool results are stored in `tool_results`

Because that data is ignored on rehydration, continuing a remembered conversation loses non-text context.

## Impact

This breaks continued conversations in at least these cases:

- previous user messages had attachments
- previous assistant turns included tool calls
- previous tool results need to remain in context for later turns

In practice, the conversation history becomes behaviorally different after retrieval than it was when originally stored.

## Expected

Rehydration should restore the original message types and associated structured data, for example:

- `UserMessage` with attachments
- `AssistantMessage` with tool calls
- `ToolResultMessage` where applicable

## Actual

Every row becomes a plain text `Message`, so structured context is discarded.

## Relevant code

- `src/Storage/DatabaseConversationStore.php`
- `storeUserMessage()`
- `storeAssistantMessage()`
- `getLatestConversationMessages()`

If useful, I can open a PR with a proposed reconstruction path.

PR #392, opened 2026-04-13
fix(anthropic): preserve provider content blocks on assistant replay (refs #298)

### LINKED ISSUE #298: Results of provider tool calls are forgotten by agents
The results of provider tool calls (e.g. web search) are available in Prism as `additionalContent` but this seems to be dropped by laravel/ai even for agents with `RemembersConversations`.

As a result, agents can end up performing unecessary duplicate tool calls on every conversation turn, which can be quite costly.

CodeWrap:
Hit this integrating advisor — every multi-turn advisor conversation 400s because the `advisor_tool_result` block drops on replay.

`MapsMessages::mapAssistantMessage()` rebuilds assistant content from `text` + `tool_calls` only. On replay that drops Anthropic provider blocks like `server_tool_use`, `server_tool_result`, `thinking`, and `redacted_thinking`. In multi-turn conversations that used server tools, the next request can fail with `400 invalid_request_error` because the assistant history no longer matches what Anthropic expects. This is the in-memory replay side of #298.

This change adds a `contentBlocks` property to `AssistantMessage` and uses it verbatim in `mapAssistantMessage()` when present, instead of rebuilding from `text` + `tool_calls`. The non-streaming parser now passes the raw ordered content array into `AssistantMessage`. If `contentBlocks` is empty, the existing path is unchanged, so current behavior and older cached/unserialized messages still fall back to the old rebuild logic.

Scope is limited to in-memory `AssistantMessage` replay. `DatabaseConversationStore` still rehydrates assistant messages from `text` + `tool_calls`, so persisted conversations are unchanged here; that needs schema work and is already tracked in #380. Streaming is also unchanged: same-request replay already uses the raw response content directly and does not go through `mapAssistantMessage()`.

Tests cover the verbatim mapping path, parser population of `contentBlocks`, round-tripping a parsed assistant message back through `mapMessages()`, and the fallback behavior when `contentBlocks` is absent. Anthropic tests are green. Related: #380 for persistence, #301 for adjacent reasoning-data loss, and #349 for the TextOrchestrator refactor this avoids touching.

### DIFF
--- a/src/Gateway/Anthropic/Concerns/MapsMessages.php
+++ b/src/Gateway/Anthropic/Concerns/MapsMessages.php
@@ -54,9 +54,6 @@ trait MapsMessages
      */
     protected function mapAssistantMessage(AssistantMessage|Message $message, array &$mapped): void
     {
-        // In-memory replay only: persisted assistant messages rehydrated by
-        // DatabaseConversationStore have no provider blocks and fall through
-        // to the rebuild path. See #380 for the persistence story.
         if ($message instanceof AssistantMessage && filled($message->providerContentBlocks)) {
             $mapped[] = [
                 'role' => 'assistant',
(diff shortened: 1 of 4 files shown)


pushpak1300:
can you fix the merge conflict ?

CodeWrap:
done

PR #399, opened 2026-04-13
Migrate DeepSeek provider to a native gateway

pushpak1300:
DeepSeek exposes an OpenAI Chat Completions-compatible API — the same shape Groq and OpenAI already run natively in this repo — so routing it through Prism adds a third-party layer with no benefit. This PR brings DeepSeek onto the native gateway pattern for direct control over the wire protocol, tool loops, streaming, and error handling.

### DIFF
--- a/config/ai.php
+++ b/config/ai.php
@@ -73,6 +73,7 @@ return [
         'deepseek' => [
             'driver' => 'deepseek',
             'key' => env('DEEPSEEK_API_KEY'),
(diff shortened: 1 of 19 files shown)


vinitkadam03:
Waiting for all providers to get migrated so that we can ship client executed tools & tool approval flow along with telemetry.

Telemetry is something that I need a lot when optimizing agents. Currently, I use a custom fork of prism which supports both client executed and phoenix telemetry.

PR #410, opened 2026-04-15
Add Anthropic prompt caching for system prompt, tools, and messages

### LINKED ISSUE #119: Support Anthropic prompt caching (cache_control on system messages)

GioChocolateBro:
Fixes #119

### Problem

Anthropic prompt caching cuts input-token cost ~90% on warm hits, but the SDK has no path to enable it on the native `AnthropicGateway`. Three issues, not one:

1. **No system-prompt caching.** The `system` field is sent as a plain string with no way to attach `cache_control` metadata.
2. **`tool_result_cache_type` regressed silently.** It worked through the Prism gateway via `MessageMap`, but disappeared during the Prism → native migration. `grep -r tool_result_cache_type src/` returned zero matches before this PR.
3. **No way to cache tool definitions or arbitrary messages.** Anthropic supports up to 4 cache breakpoints (tools, system, messages); the SDK exposed none of them.

### Solution

A single canonical `cache_control` provider option, matching [Anthropic's documented format](https://docs.anthropic.com/en/docs/build-with-claude/prompt-caching) verbatim, applied at three layers:

#### 1. System prompt

```php
class AnalyticsAgent implements Agent, HasProviderOptions
{
    use Promptable;

    {
        return file_get_contents(resource_path('agents/analytics.md')); // ~10k tokens
    }

    {
        return $provider === Lab::Anthropic ? [
            'cache_control' => ['type' => 'ephemeral'],              // 5m default
            // or: 'cache_control' => 'ephemeral'                    // string shorthand
            // or: 'cache_control' => ['type' => 'ephemeral', 'ttl' => '1h']
        ] : [];
    }
}
```

Wraps `system` as a content-block array with `cache_control`. The SDK strips the key from the top-level body so it doesn't leak.

#### 2. Tool definitions

Any tool implementing `HasProviderOptions` becomes a breakpoint candidate. Anthropic caches every tool up to and including the marked tool as a single prefix, so a single breakpoint at the end is the canonical pattern:

```php
class ExecuteSqlTool implements Tool, HasProviderOptions
{
    public function description(): string { return '...'; }
    public function handle(Request $r): string { return '...'; }
    public function schema(JsonSchema $s): array { return [...]; }

    {
        return $provider === Lab::Anthropic
            ? ['cache_control' => ['type' => 'ephemeral']]
            : [];
    }
}
```

If multiple tools opt in, the last one wins (per Anthropic's prefix-caching semantics).

#### 3. Messages (incl. tool results)

`Message::withProviderOptions()` is new — it attaches options to any message, including `ToolResultMessage`. The breakpoint lands on the last content block of that message:

```php
$toolResultMessage = (new ToolResultMessage($results))
    ->withProviderOptions(['cache_control' => ['type' => 'ephemeral']]);
```

Plus the agent-level shorthand for the common case (cache the most recent tool result automatically across the tool-call loop):

```php
{
    return ['tool_result_cache_type' => 'ephemeral'];
}
```


Tlapi:
This would be really nice! +1

stuartbrameld:
we could also *really* do with prompt caching in the sdk to reduce costs

GioChocolateBro:
@stuartbrameld that's exactly what this PR adds! Results in ~90% reduction in input tokens.

Was there a different caching scenario you had in mind? Once this is accepted, I'll open a follow-up PR for tool result caching (which was previously supported via Prism but got lost in the move to the new Anthropic gateway).

stuartbrameld:
@GioChocolateBro no it's perfect as is, we're already using your approach while we wait for this to merge 🙏

Great news re tool result caching too, I'm seeing some customers tools being rather inefficient and eating huge amounts of context.

p.s. interesting OpenAI prompt caching is fully automatic, I hadn't realised this

GioChocolateBro:
Hey @pushpak1300 could you have a look? 

It'll make a lot of impact for people in terms of costs, thank you!

pushpak1300:
https://github.com/laravel/ai/issues/119#issuecomment-4387404568

pushpak1300:
Closing this until we get solid reason to add block level caching.

joao-salomao-ug:
> Closing this until we get a solid reason to add block-level caching.

@pushpak1300, any company running a serious workload with Anthropic would be interested in the block-level caching... I'm surprised the SDK dropped support for it since Prism already supported it when it was used as a gateway. Here's a solid reason: cost reduction.

pushpak1300:
Hey @joao-salomao-ug I am not against adding support for caching but anthropic does provide automatic caching and its better and currently [suppoted](https://github.com/laravel/ai/issues/119#issuecomment-4387404568).

can you please tell me real workflows you're building where you need to use block level caching and automatic caching is not working for your use case ?

joao-salomao-ug:
Absolutely @pushpak1300! First off, sorry if my earlier comment came across a bit harsh - wasn't my intention. Just wanted to share some practical context on this.

A nuance about automatic caching: it's the best choice for multi-turn conversations, where the growing message history should be cached automatically. But this changes when you have 1-turn request workloads and you know exactly where the cache breakpoints should be; it becomes even more relevant when you're working with big prompts where you'd pay $6.25/MTok | $10/MTok depending on the TTL.

TL;DR: just because a prompt block is cacheable, it doesn't mean it should be cached.

Real example: a support copilot that reads incoming tickets, classifies them, and drafts a reply for human agents. The system prompt has multiple blocks with **different change frequencies**:

```json
{
  "model": "claude-opus-4-8",
  "max_tokens": 4096,
  "system": [
    {
      "type": "text",
      "text": "<base_instructions>\n... ~3,000 tokens: agent role, classification taxonomy (billing,\n    bug, how-to, feature-request, account-access, outage),\n    secondary tags (churn-risk, vip, escalation-needed),\n    tone rules, safety boundaries, output format spec ...\n</base_instructions>",
      "cache_control": { "type": "ephemeral", "ttl": "1h" }
    },
    {
      "type": "text",
      "text": "<output_schema>\n... ~1,500 tokens: structured output format the model must follow\n    (CLASSIFICATION, DRAFT REPLY, AGENT NOTES sections) ...\n</output_schema>",
      "cache_control": { "type": "ephemeral", "ttl": "1h" }
    },
    {
      "type": "text",
      "text": "<company_config>\n... ~2,000 tokens: product catalog (plans, tiers, pricing),\n    billing rules (refund thresholds, proration policy),\n    escalation matrix, brand voice guidelines,\n    knowledge base article index ...\n</company_config>",
      "cache_control": { "type": "ephemeral", "ttl": "5m" }
    },
    {
      "type": "text",
      "text": "<ticket>\n  From: j.reyes@example.com\n  Account: Growth tier, monthly\n  Subject: Charged twice this month\n  Body: Hey, I have been charged twice this month and honestly\n  I am about ready to switch to a competitor. Can you sort it out today?\n</ticket>"
    },
    {
      "type": "text",
      "text": "<account_context>\n  Customer since: 2022-03-14\n  MRR: $480\n  Open tickets: 0\n  Last interaction: billing inquiry, resolved, 2025-11-02\n  Health score: at-risk\n</account_context>"
    }
  ],
  "messages": [
    {
      "role": "user",
      "content": "Classify this ticket and draft a response."
    }
  ]
}
```

Three explicit breakpoints, two different TTLs, nothing implicit:

**Breakpoint 1 (1h TTL)** on the base instructions (~3,000 tokens). Universal across all companies using the copilot. This is separated from the output schema because these blocks are owned by different teams and ship independently - if the output schema gets a new field, this cache survives.

**Breakpoint 2 (1h TTL)** on the output schema (~4,500 tokens total). Also universal, same TTL. When this block changes, only the schema needs a cache write; breakpoint 1 still hits.

**Breakpoint 3 (5m TTL)** on the company config (~6,500 tokens total). Contains the product catalog, billing rules, and escalation matrix for a specific company. Only needs to live for the duration of that company's ticket batch, so 5m is enough and we avoid paying the 2x write premium.

Everything after breakpoint 3 (ticket body, account context, user message) is dynamic per request and **never cached**, even if by coincidence two consecutive requests carry the same ticket (e.g. a retry). We know the next request will be a different ticket, so caching those blocks is guaranteed waste. Automatic caching has no way to know that; it sees repeated content and caches it, paying the write premium on tokens that will never hit.

### What this looks like with explicit breakpoints

**Processing 200 tickets for Company A:**
- Request 1: cache miss on all three breakpoints → write costs
- Requests 2-200: full cache hit on all 6,500 static tokens → $0.50/MTok. The ticket and account context blocks are sent as regular input at $5/MTok, no write premium.

**Switching to Company B:**
- Company config changes (different products, billing rules) → breakpoint 3 misses
- Breakpoints 1 and 2 **still hit** from the 1h TTL → only the company config pays a cache write
- Requests 2-200 for Company B: full cache hit again

**Updating the output schema (new classification tag added):**
- Breakpoint 2 misses → schema gets rewritten
- Breakpoint 1 **still hits** → base instructions untouched
- Breakpoint 3 also needs a rewrite (prefix changed above it)

### Now compare: what happens with automatic caching on the same workload

With automatic caching you'd send something like:

```json
{
  "model": "claude-opus-4-8",
  "max_tokens": 4096,
  "cache_control": { "type": "ephemeral" },
  "system": [
    { "type": "text", "text": "<base_instructions>... ~3,000 tokens ...</base_instructions>" },
    { "type": "text", "text": "<output_schema>... ~1,500 tokens ...</output_schema>" },
    { "type": "text", "text": "<company_config>... ~2,000 tokens ...</company_config>" },
    { "type": "text", "text": "<ticket>... different every request ...</ticket>" },
    { "type": "text", "text": "<account_context>... different every request ...</account_context>" }
  ],
  "messages": [
    { "role": "user", "content": "Classify this ticket and draft a response." }
  ]
}
```

[Per the docs](https://platform.claude.com/docs/en/build-with-claude/prompt-caching#automatic-caching), the system places the breakpoint on the last cacheable block automatically. Here's what happens:

**Request 1:** The API caches **everything** up to the last block, including the ticket and account context. You pay the write premium (1.25x at 5m, or 2x at 1h) on ~8,500 tokens instead of just 6,500.

**Request 2 (different ticket):** The prefix diverges at the `<ticket>` block. The API reads from cache up to the company config (great), but the ticket + account context from request 1 was written to cache for nothing. It'll never be read again because the next ticket is always different. And now the API writes a new cache entry that includes request 2's ticket and account context, wasting the write premium again.

**This repeats on every single request.** Every ticket pays the write premium on ~2,000 tokens of dynamic content that will never hit. Across 200 tickets that's 400k tokens of wasted cache writes.

On top of that, you're locked to **a single TTL for everything**. If you pick 5m, the base instructions might expire between batches when switching companies. If you pick 1h, you're paying 2x on the company config block that only needs to live for the current batch, and 2x on the dynamic ticket blocks that will never be read from cache at all.

### Math with Opus pricing (per 6,500 static tokens)

- **No caching**: $5/MTok = $0.0325/req
- **Cache hit**: $0.50/MTok = $0.00325/req
- **1h cache write** (base + schema, 4,500 tok): $10/MTok = $0.045 (once)
- **5m cache write** (company config, 2,000 tok): $6.25/MTok = $0.0125 (once per company)

200 tickets across 2 companies: **~$1.36 with explicit caching vs ~$13.00 without**. ~89% savings.

With automatic caching the savings are lower because you're also paying write premiums on the dynamic blocks every request, and you can't mix TTLs to keep the base instructions warm across company switches.

Both have their place. Automatic is great for conversational agents where the prefix grows naturally. Explicit breakpoints are for when you control the prompt structure and know the change frequency of each block.

Issue #388, opened 2026-04-12
Gemini HTTP 400 errors in v0.4.5


seankndy:
After updating to 0.4.5 and thus using the new Gemini gateway (#312), I'm getting HTTP 400 errors when there are tool calls.  Went back to 0.4.3, works fine again.  Using model gemini-3.1-flash-lite-preview.  I am not using any provider options for Gemini.

Stacktrace:

```log
[2026-04-12 00:17:18] production.ERROR: HTTP request returned status code 400 {"userId":2,"exception":"[object] (Illuminate\\Http\\Client\\RequestException(code: 400): HTTP request returned status code 400 at /home/xxxx/app/vendor/laravel/framework/src/Illuminate/Http/Client/Response.php:332)
[stacktrace]
#0 /home/xxxx/app/vendor/laravel/framework/src/Illuminate/Http/Client/Response.php(347): Illuminate\\Http\\Client\\Response->toException()
#1 /home/xxxx/app/vendor/laravel/framework/src/Illuminate/Http/Client/PendingRequest.php(1064): Illuminate\\Http\\Client\\Response->throw()
#2 /home/xxxx/app/vendor/laravel/framework/src/Illuminate/Support/helpers.php(393): Illuminate\\Http\\Client\\PendingRequest->Illuminate\\Http\\Client\\{closure}()
#3 /home/xxxx/app/vendor/laravel/framework/src/Illuminate/Http/Client/PendingRequest.php(1043): tap()
#4 /home/xxxx/app/vendor/laravel/framework/src/Illuminate/Support/helpers.php(328): Illuminate\\Http\\Client\\PendingRequest->Illuminate\\Http\\Client\\{closure}()
#5 /home/xxxx/app/vendor/laravel/framework/src/Illuminate/Http/Client/PendingRequest.php(1041): retry()
#6 /home/xxxx/app/vendor/laravel/framework/src/Illuminate/Http/Client/PendingRequest.php(885): Illuminate\\Http\\Client\\PendingRequest->send()
#7 /home/xxxx/app/vendor/laravel/ai/src/Gateway/Gemini/Concerns/HandlesTextStreaming.php(267): Illuminate\\Http\\Client\\PendingRequest->post()
[...]

PR #419, opened 2026-04-17
fix(openai, anthropic): cast empty tool args to object

CodeWrap:
This fixes replaying assistant tool calls for Anthropic and OpenAI when a stored `ToolCall` has no arguments. `mapAssistantMessage()` rebuilds the tool blocks from the stored calls, and the empty case was being serialized as `[]` instead of `{}`. Both APIs reject that shape on replay, so conversations that included argument-less tool calls would 400.

The change is just to cast the empty case to `(object) []` via `$toolCall->arguments ?: (object) []`, which matches what Groq, Mistral, and Xai already do. Non-empty arguments are unchanged.

Gemini already got its version of this in #409 by skipping empty args entirely, but Anthropic and OpenAI were missed there. This came up off the back of #388 for Gemini, and we linked that thread to #389.

I validated this against both live APIs. Anthropic was returning a 400 on the tool input shape, and OpenAI was returning a 400 on the arguments shape. This adds four tests covering empty and non-empty arguments for both gateways. Existing Groq/Mistral/Xai coverage stays green.

### DIFF
--- a/src/Gateway/Anthropic/Concerns/MapsMessages.php
+++ b/src/Gateway/Anthropic/Concerns/MapsMessages.php
@@ -86,7 +86,7 @@ trait MapsMessages
                     'type' => 'tool_use',
                     'id' => $toolCall->id,
                     'name' => $toolCall->name,
-                    'input' => $toolCall->arguments,
+                    'input' => $toolCall->arguments ?: (object) [],
                 ];
             }
         }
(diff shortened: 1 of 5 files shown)


pushpak1300:
Can you fix conflicts ?

CodeWrap:
done

stuartbrameld:
we're hitting this too

CleanShot 2026-04-29 at 12 06 48@2x

PR #421, opened 2026-04-17
Add missing timeout to VoyageAiGateway HTTP client

sumaiazaman:
## What

`VoyageAiGateway::client()` was missing a `->timeout()` call, meaning HTTP requests could hang indefinitely. `CohereGateway` and `JinaGateway` both apply a 30-second timeout — this brings `VoyageAiGateway` in line with that.

## Why

Consistent timeout behaviour across all gateway HTTP clients prevents requests from hanging under slow or unresponsive upstream APIs.


sumaiazaman:
Closing as the upstream refactoring in the VoyageAi gateway restructure already includes timeout support via the `CreatesVoyageAiClient` concern.

PR #422, opened 2026-04-17
Remove redundant match expressions in ElevenLabsGateway

sumaiazaman:
## What

`generateTranscription()` used two `match (true)` expressions that each had a single arm — `$audio instanceof TranscribableAudio`. Since `$audio` is already typed as `TranscribableAudio` in the method signature, the condition is always true and the match adds no value.

## Why

Simplifies the code and removes the risk of an `UnhandledMatchError` if PHP ever evaluates the match with no matching arm (e.g. in a mocked test context). Direct method calls on the typed parameter are clearer.

### DIFF
--- a/src/Gateway/ElevenLabsGateway.php
+++ b/src/Gateway/ElevenLabsGateway.php
@@ -62,13 +62,10 @@ class ElevenLabsGateway implements AudioGateway, TranscriptionGateway
         bool $diarize = false,
         int $timeout = 30
     ): TranscriptionResponse {
-        $audioContent = $audio->content();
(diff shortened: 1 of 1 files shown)


balu-lt:
Great, so the variables are unnecessary, just use `$audio->` in the `$reponse =`.

sumaiazaman:
> Great, so the variables are unnecessary, just use '$audio->' in the '$reponse ='.

Good catch! Updated the PR to inline $audio->content() and $audio->mimeType() directly into the attach() call and removed the intermediate variables.

pushpak1300 (approved):
LGTM

PR #424, opened 2026-04-17
Add HandlesFailoverErrors to VoyageAiGateway

sumaiazaman:
## What

`VoyageAiGateway` was making raw HTTP calls without error handling. All other gateways use the `HandlesFailoverErrors` concern to convert 429 and 503 responses into `RateLimitedException` and `ProviderOverloadedException`, enabling the SDK's failover mechanism.

This adds the trait and wraps both `generateEmbeddings` and `rerank` HTTP calls with `withErrorHandling()`.

## Why

Without this, rate limit or overload responses from Voyage AI bubble up as raw `RequestException` instead of failoverable exceptions, so the SDK's failover system never triggers for this provider.


pushpak1300:
Thanks.

PR #427, opened 2026-04-17
Memoize ElevenLabs gateway instances

maherelgamil:
`ElevenLabsProvider::audioGateway()` and `transcriptionGateway()` use the `??` operator instead of `??=`, so the memoized property is never assigned — every call allocates a fresh `ElevenLabsGateway` instead of reusing the cached one.

This diverges from the pattern every other provider follows (`AnthropicProvider`, `OpenAiProvider`, `CohereProvider`, `JinaProvider`, `VoyageAiProvider`, etc., all use `??=`).

## Fix

```diff
- return $this->audioGateway ?? new ElevenLabsGateway;
+ return $this->audioGateway ??= new ElevenLabsGateway;
```

(and the same for `transcriptionGateway()`)

## Test plan

Includes a regression test that verifies both accessors return the same instance on repeated calls. Verified the test fails on `main` (without the fix) and passes with it.

### DIFF
--- a/src/Providers/ElevenLabsProvider.php
+++ b/src/Providers/ElevenLabsProvider.php
@@ -25,7 +25,7 @@ class ElevenLabsProvider extends Provider implements AudioProvider, Transcriptio
      */
     public function audioGateway(): AudioGateway
     {
-        return $this->audioGateway ?? new ElevenLabsGateway;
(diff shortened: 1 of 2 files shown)


pushpak1300:
Thanks.

PR #430, opened 2026-04-20
fix: detect LocalImage MIME type across native gateways

maherelgamil:
Three native gateways (OpenAI, xAI, Gemini) read `$attachment->mime` directly when mapping a `LocalImage`. That property is nullable —                 
  `LocalImage::__construct` accepts `?string $mimeType = null` — so if you don't pass a MIME explicitly, the outgoing request is malformed:              
                  
  - OpenAI / xAI send `data:;base64,<payload>` (rejected as invalid `input_image`)
  - Gemini sends `inlineData.mimeType: null`

  The other gateways (OpenRouter, Groq, Mistral, DeepSeek, Anthropic) already go through `$attachment->mimeType()`, which falls back to
  `Filesystem::mimeType($path)` for auto-detection. This PR switches the three outliers to the same call, with an `?? 'image/png'` guard for the edge
  case where detection also returns null.

  While here, I pulled that `mimeType() ?? 'image/png'` fallback up to `Image::dataUriMimeType()` since it was duplicated across every image arm in all
  seven gateways.

  Added one regression test per affected gateway — each constructs a `LocalImage` without a MIME and asserts the outgoing request carries `image/png`.
  All three fail on `main` and pass with the fix.


tests/Feature/Providers/Jina/EmbeddingTest.php
pushpak1300, line 2: hey this test file are not related to this MR. can you remove and make diff clean ?

pushpak1300 (changes requested):

maherelgamil:
@pushpak1300 while I was in here I noticed `'data:'.$mime.';base64,'.$base64` is repeated across most of the gateway attachment traits. If you're open 
  to it, I'd be happy to do a small dedup pass (probably an `asDataUri()` on the file classes) in a follow-up PR after this lands. Let me know.          
                                                                                                                                                         
  Or the slightly longer version if you want to mention the MIME fallback too:                                                                           
                                                                                                                                                         
  While reviewing this I noticed the `data:<mime>;base64,<payload>` construction is duplicated across most gateway traits, and the `mimeType() ?? 'image/png'` fallback is repeated too (this PR collapses the image one). Happy to open a follow-up that pushes both down to the file classes if that   
  direction makes sense to you.

tests/Feature/Providers/Jina/EmbeddingTest.php
maherelgamil: @pushpak1300 done , thanks

PR #436, opened 2026-04-22
use getter functions in toArray method of file classes

GigaGiorgadze:
Hello,
i noticed that data that returned in `toArray` was not matching actual data used in `StoredDocument` class, class would use basename of the file in case name was not set which was not reflected in `toArray` function, same goes for `mimeTypes`. I had issue only with `StoredDocument` class but i thought each class's `toArray` using same style would be better


pushpak1300:
Thanks

PR #440, opened 2026-04-22
Add tool mapping test suite for DeepSeek provider

sumaiazaman:
## What

Adds `ToolMappingTest.php` for the DeepSeek provider, covering the standard tool schema mapping scenarios present for all other text providers (Groq, Mistral, Ollama, OpenRouter, Xai).

## Tests added

- Tool with parameters includes the correct JSON schema
- Tool with empty schema includes required parameters structure
- Tool parameters are not wrapped in a schema definition


pushpak1300:
Thanks

PR #447, opened 2026-04-24
Add image generation feature tests for xAI provider

sumaiazaman:
## Summary

- Adds `tests/Feature/Providers/Xai/ImageGenerationTest.php` covering the `XaiImageGateway`
- Tests verify the request body includes `model`, `prompt`, and `response_format: b64_json`
- Tests confirm `size` and `quality` are not included (xAI image gateway does not send them)
- Tests verify the response is parsed correctly and bearer token auth is sent
- Follows the same pattern as `tests/Feature/Providers/OpenAi/ImageGenerationTest.php`

### DIFF
--- a/tests/Feature/Providers/Xai/ImageGenerationTest.php
+++ b/tests/Feature/Providers/Xai/ImageGenerationTest.php
@@ -21,7 +21,7 @@ function fakeXaiImageResponse(): PromiseInterface
     ]);
 }
 
-test('image request includes model and prompt', function () {
+test('image request includes model, prompt, and b64_json response format', function () {
     Http::fake([
         '*' => fakeXaiImageResponse(),
     ]);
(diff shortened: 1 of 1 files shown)


tests/Feature/Providers/Xai/ImageGenerationTest.php
```diff
@@ -0,0 +1,103 @@
...
+    ]);
+
+    Image::of('A red apple')->generate(provider: 'xai', model: 'grok-imagine-image');
+
+    Http::assertSent(function (Request $request) {
+        $body = json_decode($request->body(), true);
+
+        return ! array_key_exists('quality', $body);
+    });
+});
```
pushpak1300: These four tests fake the same response, issue the same call, and each assert one key in the body. Collapse into a single test with one `assertSent` closure that checks model, prompt, response_format, and absence of size/quality together. ~50 lines of boilerplate for 4 one-liner assertions.

tests/Feature/Providers/Xai/ImageGenerationTest.php
```diff
@@ -0,0 +1,103 @@
...
+    ]);
+
+    Image::of('A red apple')->generate(provider: 'xai', model: 'grok-imagine-image');
+
+    Http::assertSent(function (Request $request) {
+        $body = json_decode($request->body(), true);
+
+        return ! array_key_exists('size', $body);
+    });
+});
```
pushpak1300: `XaiImageGateway::generateImage()` accepts `$size` and `$quality` parameters but silently drops them. To actually test the drop (not just the default), pass size: '1:1' / quality: 'high' into generate() and assert the keys are still absent from the body. Also worth renaming to omits size (unsupported by xAI) so the why lives in the test name

tests/Feature/Providers/Xai/ImageGenerationTest.php
```diff
@@ -0,0 +1,103 @@
...
+test('image response is correctly parsed', function () {
+    Http::fake([
+        '*' => fakeXaiImageResponse(),
+    ]);
+
+    $response = Image::of('A red apple')->generate(provider: 'xai', model: 'grok-imagine-image');
+
+    expect($response->images)->toHaveCount(1)
+        ->and($response->images->first()->image)->toBe(base64_encode('fake-image'))
+        ->and($response->meta->provider)->toBe('xai');
```
pushpak1300, line 80: Assert `$response->images->first()->mimeType` is `'image/jpeg'` too — the gateway hardcodes that at `XaiImageGateway.php:50` and it's currently untested.

pushpak1300 (changes requested):
Few changes.

sumaiazaman:
All three points addressed:
- Collapsed the four request-body tests into a single `image request includes model, prompt, and b64_json response format` that checks model, prompt, response_format, and absence of size/quality together
- Added `->square()` and `->quality('high')` calls to the size/quality tests and renamed them to `omits size when provided because xAI does not support it` / `omits quality when provided because xAI does not support it`
- Added `->and($response->images->first()->mime)->toBe('image/jpeg')` to the response test

PR #451, opened 2026-04-24
Add base URL feature tests for VoyageAi provider

sumaiazaman:
## Summary

- Adds `tests/Feature/Providers/VoyageAi/BaseUrlTest.php`
- VoyageAi supports a configurable `url` key in `config/ai.php` but had no test verifying it works
- Tests cover: custom URL for embeddings, custom URL for reranking, and fallback to default `https://api.voyageai.com/v1`
- Follows the same pattern as `Groq`, `Mistral`, `Ollama`, and other provider `BaseUrlTest` files

### DIFF
--- a/tests/Feature/Providers/VoyageAi/BaseUrlTest.php
+++ b/tests/Feature/Providers/VoyageAi/BaseUrlTest.php
@@ -54,10 +54,10 @@ test('voyageai reranking requests use the configured base url', function () {
 });
 
 test('voyageai requests fall back to the default base url', function () {
-    config(['ai.providers.voyageai' => [
-        ...config('ai.providers.voyageai'),
-        'key' => 'test-key',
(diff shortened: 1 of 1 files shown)


pushpak1300 (changes requested):
Looks good overall and the added coverage makes sense.

nit: the fallback test doesn’t seem to truly exercise the gateway fallback path, since `config/ai.php` already sets `voyageai.url` to the default value. It would be stronger to set url to null or remove it explicitly, so the test verifies the actual fallback logic rather than the config default.

sumaiazaman:
Good catch, thanks! Updated the fallback test to use `array_diff_key` to strip the `url` key entirely from the config, so the gateway's hardcoded default is what's actually exercised rather than the value from `config/ai.php`.

PR #456, opened 2026-04-25
Centralize attachment serialization logic across file types and gateways

maherelgamil:
Attachment serialization is duplicated across provider mappings, and remote/provider files should not gain inline helpers they cannot safely support. This centralizes inline serialization for content-backed files while keeping remote URL mappings from probing external files.

### Approach

- Add an `InlineFile` contract for Base64, local, and stored image/document/audio files only.
- Share data URI serialization through `SerializesInlineContent`.
- Use `HasBase64Content` for Base64-backed files and `EncodesContentToBase64` for local/stored files.
- Keep inline MIME fallbacks on the abstract file types and use explicit remote MIME values for Gemini file URLs.
- Move repeated uploaded-file data URI conversion into `MapsUploadedFiles`.


pushpak1300 (approved):

src/Gateway/Gemini/Concerns/MapsAttachments.php
```diff
@@ -40,8 +39,8 @@ protected function mapAttachments(Collection $attachments): array
                 ],
                 $attachment instanceof Base64Image => [
                     'inlineData' => [
-                        'mimeType' => $attachment->mime,
-                        'data' => $attachment->base64,
+                        'mimeType' => $attachment->resolvedMimeType(),
```
crynobone: We have `content()` but here it's `resolvedMimeType()`. Can't it just be `mimeType()`?

src/Gateway/Gemini/Concerns/MapsAttachments.php
```diff
@@ -52,16 +51,14 @@ protected function mapAttachments(Collection $attachments): array
...
                     ],
                 ],
                 $attachment instanceof StoredImage => [
                     'inlineData' => [
-                        'mimeType' => $attachment->mimeType() ?? 'image/png',
-                        'data' => base64_encode(
-                            Storage::disk($attachment->disk)->get($attachment->path)
-                        ),
+                        'mimeType' => $attachment->resolvedMimeType(),
+                        'data' => $attachment->base64(),
```
crynobone: We have `asDataUri()`, why not `asEncoded()` instead of `base64()` so it's future proof. Also may consider  `dataUri()` and `encoded()`

crynobone (changes requested):

taylorotwell:
No info on what this improves or fixes for end users.

pushpak1300:
> No info on what this improves or fixes for end users.

@taylorotwell it's mostly doesn't change the api surface just address the tech debt by centralising the logic.

PR #459, opened 2026-04-27
Update remembered conversation timestamp after assistant replies

### LINKED ISSUE #381: latestConversationId() can return the wrong conversation because parent updated_at is never refreshed

dhrupo:
This updates the parent conversation timestamp when an assistant message is stored.

Without that, \ can return the most recently created conversation instead of the most recently active one when remembered conversations are resumed.

Fixes #381.


pushpak1300 (approved):

pushpak1300 (approved):

taylorotwell:
The messages table already lets you pull most recently updated conversations.

PR #462, opened 2026-04-28
fix: preserve requested model alias in tool-call follow-up requests

j-bill:
When OpenAI resolves a model alias (e.g. gpt-4.1-mini) it returns the snapshot it actually ran (e.g. gpt-4.1-mini-2025-04-14) in the responde body's model field. The laravel/ai package reads that snapshot and uses it directly as the model in any followup tool result requests.

On project scoped API keys where only the alias is allowed this causes a 403 on the follow up request: 

> HTTP request returned status code 403: "Project `proj_XXX` does not have access to model `gpt-4.1-mini-2025- ..."

**Steps to reproduce:**
1. Create a project in your organizations settings.
2. Create a API key for that specific project.
3. Under Limits -> Model Usage -> Allowed models -> allow gpt-4.1.-mini.
4. Add project ID and API key in your application.
5. Prompt an agent to use a tool with the model gpt-4.1-mini.
6. First call succeeds, follow up gets a 403.

**Fix:**
The original model alias needs to be passed into parseTextResponse, processResponse, and continueWithToolResults as $requestedModel. This allows follow-up HTTP requests to use the alias, while Meta keeps reporting the resolved snapshot for observability and traceability. 

The test is updated to run two consecutive tool call rounds (three HTTP requests total). This is required to trigger the second continuation where the bug actually shows up. To support this, ToolUsingAgent now has an optional maxStepsOverride to let the loop run long enough to hit that specific code path.


pushpak1300 (approved):

pushpak1300:
I'll defer this until #349 is merged.

PR #469, opened 2026-04-29
Allow configuring the database connection for DatabaseConversationStore

### LINKED ISSUE #393: Unable to change connection for DatabaseConversationStore
As it uses `DB::table` it's not possible to change the connection without having to override the entire file. Changing the connection is useful for applications using multi-tenancy, as currently it's trying to connect to the landlord table, rather than the tenant.

sumaiazaman:
## Summary

Fixes #393.

`DatabaseConversationStore` previously used `DB::table()` for all queries, making it impossible to target a non-default database connection — a common requirement in multi-tenant applications.

This change introduces:
- A `connection` constructor parameter (defaults to `null`, which keeps the existing behaviour of using the default connection)
- A protected `table()` helper that routes all queries through `DB::connection($this->connection)`
- A `conversations.connection` key in `config/ai.php` (driven by `AI_CONVERSATION_CONNECTION`) that is passed to the store at boot time

No behaviour change when `AI_CONVERSATION_CONNECTION` is not set.

### DIFF
--- a/config/ai.php
+++ b/config/ai.php
@@ -22,6 +22,21 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Conversation Store
+    |--------------------------------------------------------------------------
+    |
+    | Here you may configure the database connection used by the conversation
+    | store to persist agent conversations and messages. Set "connection" to
+    | null to use your application's default database connection.
+    |
+    */
+
+    'conversations' => [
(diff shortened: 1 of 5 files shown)


pushpak1300 (approved):

WendellAdriel:
LGTM

PR #470, opened 2026-04-29
Support configurable base URL for Jina provider

sumaiazaman:
## Summary

`JinaGateway` had a hardcoded `https://api.jina.ai/v1` base URL, unlike every other provider in the SDK (VoyageAi, Cohere, Mistral, OpenAI, etc.) which read the URL from `additionalConfiguration()`.

This makes Jina impossible to point at a proxy or self-hosted endpoint.

Changes:
- `JinaGateway::client()` now calls `baseUrl()` instead of using a string literal
- `JinaGateway::baseUrl()` reads `url` from `additionalConfiguration()`, falling back to `https://api.jina.ai/v1`
- `config/ai.php` adds `'url' => env('JINA_URL', 'https://api.jina.ai/v1')` to the jina provider block
- Adds `BaseUrlTest` covering custom URL for embeddings, custom URL for reranking, and fallback to default

No behaviour change when `JINA_URL` is not set.

### DIFF
--- a/config/ai.php
+++ b/config/ai.php
@@ -106,6 +106,7 @@ return [
         'jina' => [
             'driver' => 'jina',
             'key' => env('JINA_API_KEY'),
+            'url' => env('JINA_URL', 'https://api.jina.ai/v1'),
         ],
 
         'mistral' => [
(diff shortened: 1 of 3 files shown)


pushpak1300:
thanks.

PR #471, opened 2026-04-29
Add audio feature tests for OpenAI provider

sumaiazaman:
## Summary

Adds missing audio feature tests for the OpenAI provider. `OpenAiGateway` implements `generateAudio()` but had no corresponding feature test.

Tests cover:
- Request includes model, input, response format, and posts to the correct endpoint
- Default-female voice alias resolves to `alloy`
- Default-male voice alias resolves to `ash`
- Custom voice ID is passed through unchanged
- Instructions are included in the request when provided
- Response is base64-encoded with `audio/mpeg` mime type and correct meta
- Request sends bearer token authorization


pushpak1300:
Thanks.

PR #472, opened 2026-04-29
Support configurable base URL for ElevenLabs provider

sumaiazaman:
## Summary

`ElevenLabsGateway` had hardcoded `https://api.elevenlabs.io/v1` URLs for both audio and transcription endpoints, unlike all other providers in the SDK which read the URL from `additionalConfiguration()`.

This makes ElevenLabs impossible to point at a proxy or enterprise endpoint.

Changes:
- Extracts a `client()` helper (mirroring the pattern in `JinaGateway`, `VoyageAiGateway`, etc.) that uses `baseUrl()` instead of hardcoded strings
- `baseUrl()` reads `url` from `additionalConfiguration()`, falling back to `https://api.elevenlabs.io/v1`
- `config/ai.php` adds `'url' => env('ELEVENLABS_URL', 'https://api.elevenlabs.io/v1')` to the eleven provider block
- Adds `BaseUrlTest` covering custom URL for audio, custom URL for transcription, and fallback to default

No behaviour change when `ELEVENLABS_URL` is not set.

### DIFF
--- a/config/ai.php
+++ b/config/ai.php
@@ -90,6 +90,7 @@ return [
         'eleven' => [
             'driver' => 'eleven',
             'key' => env('ELEVENLABS_API_KEY'),
+            'url' => env('ELEVENLABS_URL', 'https://api.elevenlabs.io/v1'),
         ],
 
         'gemini' => [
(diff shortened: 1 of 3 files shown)


pushpak1300:
Thanks.

PR #474, opened 2026-04-29
Add config option to disable LLM-generated conversation titles

### LINKED ISSUE #369: Add a config option to disable generating titles or prevent them from using an LLM
It would be great if we could disable the LLM title-generation feature via a config var. Was wondering why I had random requests to haiku.

https://github.com/laravel/ai/blob/e6a470f0b6c66743f03cb35f3869d02ad67da148/src/Middleware/RememberConversation.php#L69-L84

sumaiazaman:
## Summary

Fixes #369.

Every new conversation currently fires a hidden LLM request to generate a 3-5 word title, which surprises users when they see unexpected API calls (e.g. to Claude Haiku).

This adds an `ai.conversations.title` config key (driven by `AI_CONVERSATION_TITLE`, defaulting to `true`) that, when set to `false`, skips the LLM call entirely and uses a truncated version of the first prompt as the title instead — matching the existing fallback behaviour.

No behaviour change when `AI_CONVERSATION_TITLE` is not set.

### DIFF
--- a/config/ai.php
+++ b/config/ai.php
@@ -22,6 +22,21 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Conversation Store
+    |--------------------------------------------------------------------------
+    |
+    | Here you may configure the behaviour of the conversation store used to
+    | persist agent conversations. Set "title" to false to skip LLM-generated
+    | titles and use a truncated version of the first prompt instead.
+    |
+    */
+
+    'conversations' => [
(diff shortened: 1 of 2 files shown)


pushpak1300 (approved):

pushpak1300:
Blocked by #469. Drafting till that gets merged.

pushpak1300 (approved):

PR #477, opened 2026-04-30
fix(gemini): ensure tool response parts are serialized as JSON arrays

HinchK:
The Gemini API strictly requires the 'parts' array in content blocks to be indexed JSON arrays. If tool results originate from collections with non-sequential keys, PHP serializes them as JSON objects, causing 400 Bad Request errors. Wrapping the parts in array_values() ensures correct serialization.


pushpak1300:
Duplicate of #473

PR #481, opened 2026-04-30
Add euria AI of Infomaniak

martin-lechene:


pushpak1300:
Thanks for your pull request!

PR #485, opened 2026-04-30
Add Model Context Protocol (MCP) support

josephkerkhof:
Adds MCP client support so agents can call MCP servers.

Example:

```php
public function mcpServers(): array
{
    return [
        'filesystem',
        Mcp::server('github')->only(['search_repositories']),
    ];
}
```

## Why?

MCP servers are mature. The Laravel ecosystem already has a great MCP server package (`laravel/mcp`). I think it makes sense to allow Laravel agents the ability to call these external tools. Other AI libraries like CrewAI and LangGraph have these MCP client APIs packaged in already, and I think `laravel/ai` should have it too.

## Methodology

Treat this as an MVP to flesh out the API and see if there's appetite for this by the project maintainers.

## What's included

- The API shape
- stdio transport

## Deferred for the future

- More transports: HTTP or SSE can be added in follow up PRs.
- More protocol versions:  Right now only `2025-11-25` is supported and an exception is thrown if the server doesn't implement that.


pushpak1300:
Thanks for sending Pull request but we will be tackling this in little different way. 

https://github.com/laravel/mcp/pull/149 adding client support make sense for the laravel/mcp and form there you can use mcp with ai sdk.

PR #500, opened 2026-05-05
Bedrock : support Base64 image for BedrockTextGateway

DenneulinThomas:
## Description

Adds image attachment support to `BedrockTextGateway`.

## Approach

Introduced a dedicated concern for handling attachments, keeping the gateway lean and the logic reusable. The concern covers:

- **Document mapping** – existing document handling moved into the concern
- **Base64 image handling** – images are encoded and mapped to the appropriate Bedrock content block format
- **MIME type validation** – only MIME types supported by the Bedrock API are accepted

The concern structure follows the same pattern used in **AnthropicGateway**.

## Tests

Added a test asserting that the image content block is correctly appended to the message payload.

Doc: https://docs.aws.amazon.com/fr_fr/bedrock/latest/userguide/conversation-inference-call.html


pushpak1300:
thanks

DenneulinThomas:
Thanks for merging!

PR #505, opened 2026-05-05
[0.x] Add image generation support to AzureOpenAiProvider

### LINKED ISSUE #238: AzureOpenAiProvider does not implement image generation
I found that Azure provider cannot be used for image generation in the Laravel AI SDK, even though Azure OpenAI supports image models/deployments.

Azure provider should support image generation when Azure image deployment is configured (for example dall-e-3 deployment), similar to text and embeddings support.

Is this something that’s planned to be fixed?
Thanks for your work! incredible tool!

maherelgamil:
AzureOpenAiProvider doesn't implement ImageProvider even though Azure OpenAI supports DALL-E models via deployments. Users currently can't generate images through the Azure provider. Closes #238.

Azure OpenAI DALL-E documentation: https://learn.microsoft.com/en-us/azure/ai-services/openai/dall-e-quickstart


pushpak1300:
Thanks for this. I removed edit option for now to think more about it on how to do it correctly.

PR #509, opened 2026-05-06
Fix OpenAI citation deduplication by URL

Anoop-Kadachi:
Avoid dropping distinct sources with the same title by deduplicating citations by URL and add regression coverage.


Anoop-Kadachi:
Ready for review. This fixes OpenAI citation deduplication by URL and adds regression test coverage.

pushpak1300:
Thanks.

PR #511, opened 2026-05-06
Add error handling tests for OpenAI image generation

sumaiazaman:
Adds error path coverage to `tests/Feature/Providers/OpenAi/ImageGenerationTest.php`.

The image generation gateway uses `withErrorHandling()` but the test file had no tests verifying that 429, 503, and other error responses are correctly converted to the appropriate exceptions.


pushpak1300:
Thanks for your pull request!

PR #512, opened 2026-05-06
Add error handling tests for Gemini image generation

sumaiazaman:
Adds error path coverage to `tests/Feature/Providers/Gemini/ImageGenerationTest.php`.

The image generation gateway uses `withErrorHandling()` but the test file had no tests verifying that 429, 503, and other error responses are correctly converted to the appropriate exceptions.

### DIFF
--- a/config/ai.php
+++ b/config/ai.php
@@ -22,6 +22,21 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Conversation Store
+    |--------------------------------------------------------------------------
+    |
+    | Here you may configure the database connection used by the conversation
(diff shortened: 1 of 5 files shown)


config/ai.php
```diff
@@ -20,6 +20,21 @@
...
+    |
+    | Here you may configure the database connection used by the conversation
+    | store to persist agent conversations and messages. Set "connection" to
+    | null to use your application's default database connection.
+    |
+    */
+
+    'conversations' => [
+        'connection' => env('AI_CONVERSATION_CONNECTION'),
+    ],
```
pushpak1300, line 36: I'm not sure if this changes does make sense here.

PR #513, opened 2026-05-06
Align xAI URL citations with OpenAI response parsing

Anoop-Kadachi:
## Summary

xAI responses were deduplicating `url_citation` annotations with `Collection::unique('title')`, which could drop distinct URLs that share the same title and was inconsistent with how OpenAI responses are parsed.

## Changes

- Stop deduplicating citations by title; return every `url_citation` annotation in order.
- Map optional `start_index` / `end_index` from the API into `UrlCitation`, matching the OpenAI gateway.
- Add feature tests for span indices and for citations without span fields.

## Testing

```bash
vendor/bin/pest tests/Feature/Providers/Xai/RequestMappingTest.php


pushpak1300:
Thanks.

PR #515, opened 2026-05-06
Test Anthropic streaming tool_use finish reason without tool blocks

Anoop-Kadachi:
## Summary

The `tool_use` finish-reason dataset case was skipped under the assumption that tool use never yields `StreamEnd`. That is only true when the stream contains `tool_use` blocks and the runtime executes the streaming tool-call path.

## Change

- Remove the obsolete skip for the text-only fixture with `stop_reason: tool_use`.
- Clarify the dataset label so the scenario is obvious to future readers.

## Testing

`vendor/bin/pest tests/Feature/Providers/Anthropic/StreamingTest.php`


pushpak1300:
Thanks

PR #517, opened 2026-05-07
Add missing parameter name in EmbeddingsResponse PHPDoc

mahfuz-rahman007:
The `@param` annotation for the `$embeddings` constructor parameter in `EmbeddingsResponse` is missing the variable name.

### Before
```php
@param  array<int, array<float>>
```

### After
```php
@param  array<int, array<float>>  $embeddings
```


pushpak1300:
Thanks.

PR #518, opened 2026-05-07
Add missing type declaration to ToolResult $result property

mahfuz-rahman007:
The `$result` property in `ToolResult` is the only constructor-promoted property without a type declaration. Since tool results can be of any type, `mixed` is the appropriate type hint.

### Before
```php
public $result,
```

### After
```php
public mixed $result,
```

All other properties in the class (`$id`, `$name`, `$arguments`, `$resultId`) are already typed.


pushpak1300:
Thanks

PR #524, opened 2026-05-07
Handle broadcast failures without aborting queued streams

### LINKED ISSUE #174: BroadcastAgent broadcasts ToolResult/ToolCall events that exceed WebSocket message limits
`BroadcastAgent` broadcasts every `StreamEvent` via the configured broadcast driver (Reverb/Pusher), including `ToolResult` and `ToolCall` events. Tool results from agents with data-returning tools (listing records, searching, etc.) regularly exceed the **10 KB WebSocket message limit**, causing a `BroadcastException` that fails the entire queued job — killing the stream mid-conversation.

## Steps to reproduce

```php
// 1. Create an agent with a tool that returns a large result
#[MaxSteps(5)]
class MyAgent implements Agent, HasTools
{
    use Promptable;

    {
        return 'You help search for records.';
    }

    {
        return [new SearchTool];
    }
}

// 2. The tool returns >10KB of data
class SearchTool implements Tool
{
    public function description(): string
    {
        return 'Search records';
    }

    public function schema(): array
    {
        return ['type' => 'object', 'properties' => ['query' => ['type' => 'string']]];
    }

    public function handle(Request $request): string
    {
        // Returns ~15KB of JSON — normal for listing 50+ records
        return json_encode(Record::query()->limit(50)->get()->toArray());
    }
}

// 3. Broadcast the agent on a queue
$agent = new MyAgent;
$agent->broadcastOnQueue('Search for all active records', $channel);
```

**Expected:** The stream completes successfully and the frontend receives text deltas.

**Actual:** The job fails with:
```
Illuminate\Broadcasting\BroadcastException: Pusher error: Payload too large.
```

The exception originates in `PusherBroadcaster::broadcast()` when `Pusher::trigger()` rejects the oversized payload. The entire queued job fails, so the user sees no response at all — not even the text deltas that were broadcast before the tool result.

## Root cause

In `BroadcastAgent::handle()`, every `StreamEvent` is broadcast without any size check or filtering:

```php
public function handle(): void
{
    $this->withCallbacks(fn () => $this->agent->stream(...)
        ->each(function (StreamEvent $event) {
            $event->broadcastNow($this->channels);  // <-- broadcasts everything
        })
    );
}
```

The `ToolResult` event includes the full tool result string in its `toArray()` payload:

```php
// ToolResult::toArray()
return [
    'result' => $this->toolResult->result,  // <-- can be 10KB+
    // ...
];
```

Reverb and Pusher both enforce a **10 KB per-message limit** on WebSocket frames. Tool results from real-world agents (listing records, search results, email bodies) easily exceed this.

## Impact

- **Job failure kills the entire stream.** The `BroadcastException` is unhandled, so the queued job fails. All previously broadcast text deltas are lost because the frontend relies on `stream_completed`/`stream_end` to finalize the conversation — which never arrives.
- **No partial recovery.** Even though the AI response and tool results are persisted in the `agent_conversation_messages` table, the user sees a broken/hanging chat because `is_processing` is never reset (the `.then()` callback never fires).

### Option A: Allow opting out of broadcasting specific event types

Allow agents to opt out of broadcasting specific event types. Many frontends display tool calls in real-time (e.g. showing a spinner with the tool name, or streaming partial results), so tool events shouldn't be dropped silently by default. But the current behavior — broadcasting the full, untruncated tool result — makes it impossible to use tools that return more than 10 KB.

```php
// Promptable::broadcastOnQueue()
public function broadcastOnQueue(
    string $prompt,
    Channel|array $channels,
    array $attachments = [],
    array $except = [],  // e.g. [ToolResult::class] to skip large payloads
): QueuedAgentResponse
```

### Option B: Truncate large payloads before broadcasting

Add a size check in `StreamEvent::broadcast()` and truncate the payload if it exceeds a threshold:

```php
public function broadcast(Channel|array $channels, bool $now = false): void
{
    $payload = $this->toArray();

    if (strlen(json_encode($payload)) > 9_000) {
        $payload['result'] = mb_substr($payload['result'] ?? '', 0, 5_000) . '... (truncated)';
    }

    // ... broadcast $payload
}
```

This would allow frontends to display a summary (e.g. tool name + "loading...") while the full result is available from the conversation history after the stream completes.

**Downside:** Lossy — consumers that display the real-time tool result would see truncated data. Needs a clear indicator (e.g. a `truncated: true` flag) so the frontend knows to fetch the full result from the database.

### Option C: Catch BroadcastException and continue streaming

Wrap the broadcast call in a try/catch so a single oversized event doesn't kill the entire stream:

```php
->each(function (StreamEvent $event) {
    try {
        $event->broadcastNow($this->channels);
    } catch (BroadcastException $e) {
        report($e);  // Log but don't fail the job
    }
})
```

**Downside:** Masks the real problem. The oversized events are still being serialized and sent to the broadcast driver, wasting resources. The client silently misses the tool result with no indication that it was dropped.

### Option D: Allow agents to configure which events to broadcast

Add a `broadcastableEvents()` method to the `Agent` contract:

```php
interface Agent
{
    // ...
    public function broadcastableEvents(): array;  // default: all StreamEvent subclasses
}
```

```php
// In BroadcastAgent::handle()
$allowed = $this->agent->broadcastableEvents();

->each(function (StreamEvent $event) use ($allowed) {
    if (! in_array(get_class($event), $allowed)) {
        return;
    }
    $event->broadcastNow($this->channels);
})
```

**Upside:** Maximum flexibility. Agents that want real-time tool results can opt in.
**Downside:** More API surface. Most agents would use the same default.

## Recommendation

A combination of **Option B + Option D** would cover the most use cases:

- **Truncate by default** (Option B) so tool events are still broadcast for frontends that show real-time tool activity (tool name, partial result), but without exceeding the WebSocket limit. Include a `truncated: true` flag so the frontend knows to fetch the full result from conversation history.
- **Allow opting out** (Option D) for agents that don't need tool events broadcast at all, avoiding unnecessary serialization overhead.

Happy to provide PR.

## Environment

- laravel/ai: v0.1.5
- Broadcasting driver: Reverb (also affects Pusher)
- PHP: 8.4
- Laravel: 12.x

Anoop-Kadachi:
## Summary

`BroadcastAgent` currently broadcasts each stream event without guarding broadcast failures. If a broadcaster rejects a payload (for example, oversized tool result payloads), the queued job fails mid-stream.

This PR makes broadcasting resilient by catching per-event broadcast exceptions so streaming can continue and response callbacks still run.

Fixes #174

## Changes

- Wrap `broadcastNow()` in `BroadcastAgent::handle()` with exception handling.
- Report broadcast exceptions and continue processing subsequent events.
- Add regression test in `tests/Feature/BroadcastAgentTest.php` to verify:
  - broadcast exception does not fail the job,
  - streamed response is still resolved and delivered to `then` callbacks.

## Testing

```bash
vendor/bin/pest tests/Feature/BroadcastAgentTest.php

### DIFF
--- a/src/Jobs/BroadcastAgent.php
+++ b/src/Jobs/BroadcastAgent.php
@@ -33,7 +33,11 @@ class BroadcastAgent implements ShouldQueue
 
         $this->agent->stream($this->prompt, $this->attachments, $this->provider, $this->model)
             ->each(function (StreamEvent $event) {
-                $event->broadcastNow($this->channels);
+                try {
+                    $event->broadcastNow($this->channels);
+                } catch (\Throwable $e) {
+                    report($e);
+                }
             })
             ->then(function ($response) use (&$streamedResponse) {
                 $streamedResponse = $response;
--- a/tests/Feature/BroadcastAgentTest.php
+++ b/tests/Feature/BroadcastAgentTest.php
@@ -1,6 +1,8 @@
 <?php
 
 use Illuminate\Broadcasting\Channel;
+use Illuminate\Broadcasting\BroadcastException;
+use Illuminate\Support\Facades\Broadcast;
 use Illuminate\Support\Facades\Event;
 use Laravel\Ai\Jobs\BroadcastAgent;
 use Laravel\Ai\Responses\StreamedAgentResponse;
@@ -77,5 +79,36 @@ test('streamed response passed to then is fully resolved', function () {
     expect($received)->not->toBeNull('then() callback was never invoked')
         ->toBeInstanceOf(StreamedAgentResponse::class)
         ->and($received->events)->not->toBeEmpty()
+        ->and($received->text)->toBe('Hello world');
+});
+
+test('broadcast failures do not fail the queued stream job', function () {
+    Event::fake();
+    AssistantAgent::fake(['Hello world']);
+
+    $pendingBroadcast = Mockery::mock();
+    $pendingBroadcast->shouldReceive('as')->andReturnSelf();
+    $pendingBroadcast->shouldReceive('with')->andReturnSelf();
+    $pendingBroadcast->shouldReceive('sendNow')
+        ->andThrow(new BroadcastException('Payload too large'));
+
+    Broadcast::shouldReceive('on')
+        ->andReturn($pendingBroadcast);
+
+    $received = null;
+
+    $job = new BroadcastAgent(
+        agent: new AssistantAgent,
+        prompt: 'Say hello',
+        channels: new Channel('test-channel'),
+    );
+
+    $job->then(function ($response) use (&$received) {
+        $received = $response;
+    });
+
+    expect(fn () => $job->handle())->not->toThrow(BroadcastException::class);
+
+    expect($received)->toBeInstanceOf(StreamedAgentResponse::class)
         ->and($received->text)->toBe('Hello world');
 });


Anoop-Kadachi:
This PR is intentionally scoped to a safe reliability fix: if broadcasting a stream event fails (for example due to oversized payload), the queued stream job should continue instead of failing mid-stream.

After this is merged, I plan to open a second PR focused on Issue #174 improvements:
1. add configurable filtering of broadcasted stream event types (so apps can skip large `ToolResult`/`ToolCall` events if desired),
2. add payload-size handling for large broadcast events (truncate with an explicit `truncated` indicator, or equivalent safe guard),
3. add tests covering large tool-result broadcast behavior and non-failing stream completion.

I kept this PR minimal so the immediate failure mode can be fixed quickly.

pushpak1300:
Thanks for tackling this. After review I'd rather solve #174 at the source by truncating oversized event payloads in `toArray()` with a `truncated: true` flag (Option B from the issue), rather than catching the exception after the fact. Closing this in favor of that approach — happy to take a PR for it, or I'll pick it up.

Anoop-Kadachi:
Thanks for the review, agreed on fixing this at the source. I’ve opened a new PR that truncates oversized payloads in `toArray()` and sets `truncated: true` when trimming (Option B ), configurable via `AI_BROADCAST_MAX_PAYLOAD_BYTES`, with tests. Happy to tweak the limit or which fields are considered “heavy” if you’d like.

PR #532, opened 2026-05-08
Add NVIDIA NIM text provider

A909M:
Adds NVIDIA NIM as a Lab provider for text generation against https://integrate.api.nvidia.com/v1.            

Follows the DeepSeek provider pattern (NIM is OpenAI-compatible for chat completions):                        
              
- `Lab::Nvidia` enum + `AiManager::createNvidiaDriver()  `                                                        
- `NvidiaProvider implements TextProvider`
- `NvidiaGateway implements TextGateway` with streaming, tools, structured output, and image attachments        
-` config/ai.php` block with NVIDIA_API_KEY and overridable NVIDIA_NIM_URL
                                                                                                              
## Scope         
                                                                                                              
Text only. Embeddings, reranking and image generation live on a different host (ai.api.nvidia.com) and use    
vendor-specific request shapes, so they're cleaner as separate follow-up PRs.
                                                                                                              
## Defaults (free-tier build.nvidia.com)

- defaultTextModel: meta/llama-3.3-70b-instruct                                                               
- cheapestTextModel: meta/llama-3.1-8b-instruct
- smartestTextModel: same as default - heavier models on the catalog are credit-gated.                        
                                                                                                              
All overridable via `config('ai.providers.nvidia.models.text.*')`.
                                                                                                              
**Tested locally against a free API key - prompts, streaming, and tool calls all round-trip.**

### DIFF
--- a/config/ai.php
+++ b/config/ai.php
@@ -112,6 +112,12 @@ return [
             'key' => env('MISTRAL_API_KEY'),
         ],
 
+        'nvidia' => [
+            'driver' => 'nvidia',
+            'key' => env('NVIDIA_API_KEY'),
+            'url' => env('NVIDIA_NIM_URL', 'https://integrate.api.nvidia.com/v1'),
+        ],
+
         'ollama' => [
             'driver' => 'ollama',
             'key' => env('OLLAMA_API_KEY', ''),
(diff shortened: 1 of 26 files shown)


pushpak1300:
Deferring the new provider addition until we merged #349

PR #534, opened 2026-05-08
Fix: DeepSeek 400 Bad Request on multi-turn tool-calls by preserving `reasoning_content`

### LINKED ISSUE #533: DeepSeek V4 thinking mode: multi-turn reasoning_content handling and unsupported parameters
**Description:**

DeepSeek V4’s thinking mode introduces a `reasoning_content` field and silently ignores `temperature`, `top_p`, `presence_penalty`, and `frequency_penalty`. The current SDK doesn’t handle these correctly.

**Key issues:**

1. **Unsupported parameters** – When thinking mode is on, the API ignores `temperature`, `top_p`, `presence_penalty`, `frequency_penalty`. No error is thrown, but users may be unaware they have no effect.

2. **`reasoning_content` stitching** – In multi-turn conversations:
   - **Without tool calls** between user messages: `reasoning_content` from previous assistant messages must **not** be passed. The API will ignore it.
   - **With tool calls** between user messages: `reasoning_content` from the intermediate assistant **must** be passed back in all subsequent turns.

The SDK currently doesn’t distinguish these cases, which can break multi-turn conversations, especially when function calling is involved.

**References:**  
- https://api-docs.deepseek.com/zh-cn/guides/thinking_mode  
- https://api-docs.deepseek.com/zh-cn/guides/thinking_mode#多轮对话拼接  
- https://api-docs.deepseek.com/zh-cn/guides/thinking_mode#工具调用

k1rana:
## The Problem

Resolves #533.

The DeepSeek API requires that any `AssistantMessage` containing `tool_calls`
must also include the exact `reasoning_content` from that turn in all subsequent
requests. Without it, the API returns a `400 Bad Request`.

The previous implementation had two issues:

1. During recursive internal tool-call loops, `reasoning_content` was not being
   captured from the response and replayed back to the API.
2. When reconstructing historical messages (e.g. from a database) where
   `reasoning_content` was not persisted, the API would reject the request
   because `tool_calls` were present without their accompanying reasoning context.

## The Solution

1. **`ParsesTextResponses` and `HandlesTextStreaming`** — Updated to capture
   `reasoning_content` into `providerContentBlocks` and replay it during
   recursive tool-call loops.
2. **`MapsMessages@mapMessagesToChat`** — Updated to handle historical assistant
   messages that have `tool_calls` but no `reasoning_content`. When this occurs,
   `tool_calls` are stripped from the payload and the paired `ToolResultMessage`
   is skipped, preventing orphaned tool outputs while preserving conversational
   context via the assistant's text response.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ CLAUDE.md
 .env
 .claude
 AGENTS.md
+.zed
--- a/src/Contracts/Gateway/ImageGateway.php
+++ b/src/Contracts/Gateway/ImageGateway.php
@@ -12,7 +12,7 @@ interface ImageGateway
      * Generate an image.
      *
      * @param  array<Image>  $attachments
-     * @param  'low'|'medium'|'high'  $quality
+     * @param  'low'|'medium'|'high'|null  $quality
      */
     public function generateImage(
         ImageProvider $provider,
--- a/src/Contracts/Gateway/ImageGateway.php
+++ b/src/Contracts/Gateway/ImageGateway.php
@@ -3,7 +3,7 @@
 namespace Laravel\Ai\Contracts\Gateway;
 
 use Laravel\Ai\Contracts\Providers\ImageProvider;
-use Laravel\Ai\Files\Image as ImageFile;
+use Laravel\Ai\Files\Image;
 use Laravel\Ai\Responses\ImageResponse;
 
 interface ImageGateway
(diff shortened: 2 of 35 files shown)


k1rana:
Hi! Just wondering if there’s still anything that should be adjusted or reviewed for this PR before it can be merged?

I mainly opened this because DeepSeek stopped working properly for me after updating to the latest version, so I figured this change might help others running into the same thing too.

PR #535, opened 2026-05-08
Add missing type declaration to StructuredAnonymousAgent $schema property

sumaiazaman:
`StructuredAnonymousAgent::$schema` is always assigned a `SerializableClosure` instance in the constructor but the property lacks an explicit type declaration.


pushpak1300:
Thanks for your pull request!

PR #536, opened 2026-05-08
Broadcast stream_failed event when BroadcastAgent job fails

### LINKED ISSUE #341: Feature: Broadcast failure in the BroadcastAgent job
I think it would be nice to send a stream_failed event similar to the stream_started and stream_end events on job failure.

sumaiazaman:
Fixes #341.

When `BroadcastAgent` throws an unhandled exception the frontend has no way to know the stream failed — leaving the user with a permanently loading UI. Laravel's `failed()` hook is now used to broadcast a `stream_failed` `Error` event to the configured channels so clients can handle the failure gracefully.

### DIFF
--- a/src/Jobs/BroadcastAgent.php
+++ b/src/Jobs/BroadcastAgent.php
@@ -5,6 +5,7 @@ namespace Laravel\Ai\Jobs;
 use Illuminate\Broadcasting\Channel;
 use Illuminate\Contracts\Queue\ShouldQueue;
 use Illuminate\Foundation\Queue\Queueable;
+use Illuminate\Support\Str;
 use Laravel\Ai\Contracts\Agent;
 use Laravel\Ai\Enums\Lab;
 use Laravel\Ai\Streaming\Events\Error;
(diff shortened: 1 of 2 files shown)


pushpak1300:
The approach is mostly right, but I’d like to see two changes before merge:

1. Add a regression test for `BroadcastAgent::failed()` that asserts a `stream_failed` event is broadcast with `recoverable: false`.
2. Avoid broadcasting the raw exception message directly to the frontend, since it may leak internal/provider details. Prefer a generic public message like `The agent stream failed.` and only include safe metadata if needed.

sumaiazaman:
> The approach is mostly right, but I’d like to see two changes before merge:
> 
> 1. Add a regression test for `BroadcastAgent::failed()` that asserts a `stream_failed` event is broadcast with `recoverable: false`.
> 2. Avoid broadcasting the raw exception message directly to the frontend, since it may leak internal/provider details. Prefer a generic public message like `The agent stream failed.` and only include safe metadata if needed.

@pushpak1300, Thanks for the feedback.

Addressed the requested changes for the `BroadcastAgent::failed()` handling:

* Added a regression test asserting the `stream_failed` event broadcasts with `recoverable: false`
* Updated the frontend-facing failure message to use a generic public message instead of exposing raw exception/provider details

PR #537, opened 2026-05-08
Add error handling tests for AzureOpenAI image generation

mosabbirrakib:
## Summary

Adds error-handling test coverage for the AzureOpenAI image generation endpoint. Mirrors the pattern of the recently merged #528 (OpenAI), #502 (xAI), and #498 (OpenRouter embeddings).

The existing `ImageGenerationTest.php` covered request shape (deployment URL, quality, size, moderation, response_format) and usage parsing, but had no assertions for failure modes — even though `AzureOpenAiGateway::generateImage()` is wrapped in `withErrorHandling()`.

## Coverage added

- 429 rate-limit response → `RateLimitedException`
- 503 overloaded response → `ProviderOverloadedException`
- Generic 4xx response → `RequestException`

## Testing

```
vendor/bin/pest tests/Feature/Providers/AzureOpenAi/ImageGenerationTest.php
# 15 passed (20 assertions)
```

### DIFF
--- a/tests/Feature/Providers/AzureOpenAi/ImageGenerationTest.php
+++ b/tests/Feature/Providers/AzureOpenAi/ImageGenerationTest.php
@@ -213,8 +213,8 @@ test('image rate limit response throws rate limited exception', function () {
     Http::fake([
         'test-resource.openai.azure.com/*' => Http::response([
             'error' => [
-                'type' => 'rate_limit_error',
-                'message' => 'Rate limit exceeded',
+                'code' => '429',
+                'message' => 'Requests to the Generations_Create Operation under Azure OpenAI API have exceeded call rate limit of your current OpenAI S0 pricing tier.',
             ],
         ], 429),
     ]);
@@ -226,7 +226,7 @@ test('image overloaded response throws provider overloaded exception', function 
     Http::fake([
         'test-resource.openai.azure.com/*' => Http::response([
             'error' => [
-                'type' => 'server_error',
+                'code' => 'ServiceUnavailable',
                 'message' => 'The server is currently overloaded. Please try again later.',
             ],
         ], 503),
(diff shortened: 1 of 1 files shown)


tests/Feature/Providers/AzureOpenAi/ImageGenerationTest.php
```diff
@@ -205,3 +208,42 @@ function fakeAzureImageResponse(): PromiseInterface
...
+        'test-resource.openai.azure.com/*' => Http::response([
+            'error' => [
+                'type' => 'invalid_request_error',
+                'message' => 'Invalid model',
+            ],
+        ], 400),
+    ]);
+
+    Image::of('A red apple')->generate(provider: 'azure', model: 'gpt-image-1');
+})->throws(RequestException::class);
```
pushpak1300, line 256: AI ASSISTED REVIEW:

nit: the 400 payload reads as OpenAI-shaped rather than Azure-shaped. Per the [Azure OpenAI Images REST API reference](https://learn.microsoft.com/en-us/rest/api/azureopenai/images/generate?view=rest-azureopenai-2023-10-01-preview), errors use the `Error` / `ErrorResponse` / `InnerError` envelope (typically `{error: {code, message, innererror?}}`). For an unknown deployment, Azure returns **404 `DeploymentNotFound`** at the gateway ([Microsoft Q&A](https://learn.microsoft.com/en-us/answers/questions/1189176/how-do-i-resolve-error-deploymentnotfound-for-azur)). A more authentic inference-layer 400 example would be `contentFilter` / content-policy ([Microsoft Q&A](https://learn.microsoft.com/en-au/answers/questions/2100832/dall-e-3-error-content-policy-violation-on-image-g)) or an unknown parameter ([Azure SDK issue](https://github.com/Azure/azure-sdk-for-net/issues/49965)).

This doesn't affect the assertion (`withErrorHandling` only branches on status code), but it would make the fixture feel more Azure-realistic.

mosabbirrakib: Good catch — updated the fixtures to use the Azure-shaped envelope (`error.code` instead of `error.type`, no `error.type` field) and swapped the 400 example to a `contentFilter` / `ResponsibleAIPolicyViolation` payload with the `innererror` block, which is more representative of what the inference layer actually returns.

429 now uses `code: '429'` with the real rate-limit message format, and 503 uses `code: 'ServiceUnavailable'`. Assertions are unchanged since `withErrorHandling` only branches on status code.

PR #548, opened 2026-05-08
Add tests for OpenRouter image generation response edge cases

mosabbirrakib:
## Summary

Extends OpenRouter image generation test coverage with three response-shape scenarios that #545 didn't cover. These exercise behavior already implemented in `OpenRouterGateway::generateImage()` but not previously pinned by tests.

## Coverage added

- **Empty `choices` array** — when the upstream model returns an empty `choices: []`, the parser tolerates it via `?? []` and produces an empty images collection without errors.
- **Non-base64 image URL** — the parser only matches `data:image/...;base64,...` URLs (line 177 regex). When OpenRouter routes to a provider that returns a direct https URL, that image is filtered out via `->filter()`. This pins that the regex-mismatch path works.
- **Missing `model` in response** — the response model field falls back to the requested model via `$data['model'] ?? $model` when omitted by the upstream provider.

These complement #545 (429/503/4xx HTTP errors) by covering in-band response edge cases specific to OpenRouter's chat-completions-style image API.

## Testing

```
vendor/bin/pest tests/Feature/Providers/OpenRouter/ImageGenerationTest.php
# 19 passed (30 assertions)
```

### DIFF
--- a/tests/Feature/Providers/OpenRouter/ImageGenerationTest.php
+++ b/tests/Feature/Providers/OpenRouter/ImageGenerationTest.php
@@ -236,3 +236,55 @@ test('image http error response throws request exception', function () {
 
     Image::of('A blue circle')->generate(provider: 'openrouter', model: 'google/gemini-2.5-flash-image');
 })->throws(RequestException::class);
+
+test('image response is empty when choices array is empty', function () {
+    Http::fake([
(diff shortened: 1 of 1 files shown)


pushpak1300:
To be honest, adding test cases for every single scenario doesn’t add much value. Closing this.

PR #552, opened 2026-05-08
Mark nullable size and quality docblocks across image gateways

mosabbirrakib:
The `\$size` and `\$quality` parameters on `generateImage()` are declared as `?string \$size = null` and `?string \$quality = null` in the actual signatures, but their docblocks omitted `|null`:

```php
* @param  '3:2'|'2:3'|'1:1'  \$size           // missing |null
* @param  'low'|'medium'|'high'  \$quality    // missing |null
```

The `defaultImageOptions(?string \$size = null, ?string \$quality = null)` method on `ImageProvider` already documents these correctly with `|null`, so this aligns the gateway methods with the existing convention.

Static analyzers (PHPStan / Psalm) report a type mismatch at every call site that passes `null` to these methods because the docblock claims the parameter only accepts the three string literals.

### Files

- `src/Contracts/Gateway/ImageGateway.php` (gateway contract — only had `\$quality` documented)
- `src/Gateway/Gemini/GeminiGateway.php`
- `src/Gateway/AzureOpenAi/AzureOpenAiGateway.php`
- `src/Gateway/Xai/XaiImageGateway.php` (only documents `\$quality`)
- `src/Gateway/Bedrock/BedrockImageGateway.php`
- `src/Gateway/OpenAi/OpenAiGateway.php`

The provider-side trait `Providers/Concerns/GeneratesImages.php` and contract `Contracts/Providers/ImageProvider.php` have the same docblock issue but are intentionally left out of this PR — they're being touched by #550 (ImageAttachment fix) and would conflict if updated here.

No runtime changes; PHPDoc only.

### DIFF
--- a/src/Contracts/Gateway/ImageGateway.php
+++ b/src/Contracts/Gateway/ImageGateway.php
@@ -12,7 +12,7 @@ interface ImageGateway
      * Generate an image.
      *
      * @param  array<ImageFile>  $attachments
-     * @param  'low'|'medium'|'high'  $quality
+     * @param  'low'|'medium'|'high'|null  $quality
      */
     public function generateImage(
         ImageProvider $provider,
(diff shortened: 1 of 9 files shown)


pushpak1300:
Has Conflicts.

PR #553, opened 2026-05-09
Skip temperature and top_p for OpenAI reasoning models

### LINKED ISSUE #317: #[Temperature] is incompatible with #[UseCheapestModel] when failing over across providers
## Description

When an agent uses both `#[UseCheapestModel]` and `#[Temperature(0.8)]` with failover across providers (e.g., `provider: [Lab::OpenAI, Lab::Anthropic]`), OpenAI resolves the cheapest model to `o3-mini`, which doesn't support the `temperature` parameter. This results in a 400 error:

```
OpenAI Error [400]: invalid_request_error - Unsupported parameter: 'temperature' is not supported with this model.
```

## Root Cause

The `#[Temperature]` attribute is applied unconditionally in `CreatesPrismTextRequests` (line 37-38) before the request hits the provider — there's no way to conditionally omit it per provider or per model.

`HasProviderOptions` (added in v0.3) can't solve this because `temperature` is set via `usingTemperature()` on the Prism request, not through `withProviderOptions()`. The two are separate fields in the Prism payload, so provider-specific options have no way to override or nullify the temperature value.

## Reproduction

```php
use Laravel\Ai\Agents\Agent;
use Laravel\Ai\Attributes\Temperature;
use Laravel\Ai\Attributes\UseCheapestModel;
use Laravel\Ai\Enums\Lab;

#[UseCheapestModel]
#[Temperature(0.8)]
class MyAgent extends Agent
{
    protected $provider = [Lab::OpenAI, Lab::Anthropic];

    {
    }
}
```

When OpenAI is the first provider in the failover chain:

1. `#[UseCheapestModel]` resolves OpenAI to `o3-mini`
2. `#[Temperature(0.8)]` sets `usingTemperature(0.8)` on the Prism request
3. OpenAI rejects the request with a 400 error because `o3-mini` doesn't support `temperature`
4. The agent fails over to Anthropic (which works fine), but the OpenAI attempt is wasted

## Expected Behaviour

The SDK should handle model/temperature incompatibility gracefully. Either:

1. **Skip temperature for models that don't support it** — the SDK could maintain a list of models (or check provider metadata) and omit `temperature` when unsupported.
2. **Allow `HasProviderOptions` to override/nullify temperature per provider** — so agents can set `temperature: 0.8` for Anthropic and `null` for OpenAI.
3. **Catch the incompatibility before sending** — and either silently omit the parameter or throw a more descriptive exception.

## Workaround

Remove `#[Temperature]` from agents that use `#[UseCheapestModel]` and accept the provider's default temperature for all providers.

## Environment

- `laravel/ai`: v0.4.0
- Providers: OpenAI (`o3-mini` via `#[UseCheapestModel]`) + Anthropic (failover)

sumaiazaman:
## Summary

- OpenAI reasoning models (`o1`, `o1-mini`, `o1-preview`, `o3`, `o3-mini`, `o3-pro`) reject `temperature` and `top_p` with a 400 error
- Adds a `modelSupportsTemperature()` helper to `BuildsTextRequests` that returns `false` for models matching the `o\d` prefix pattern
- Uses this guard in all three request-building paths: initial requests, non-streaming tool loop continuations, and streaming tool loop continuations
- Also fixes the same issue for `AzureOpenAiGateway`, which reuses the same concerns
- Adds a dataset test asserting that `temperature` and `top_p` are excluded from the request for all known reasoning model variants

Fixes #317.

## Test plan

- [x] New dataset test covers `o1`, `o1-mini`, `o1-preview`, `o3`, `o3-mini`, `o3-pro`
- [x] Existing test confirms temperature is still sent for standard models (`gpt-*`)
- [x] Full OpenAI provider test suite passes (105 tests)

### DIFF
--- a/src/Gateway/OpenAi/Concerns/BuildsTextRequests.php
+++ b/src/Gateway/OpenAi/Concerns/BuildsTextRequests.php
@@ -39,10 +39,12 @@ trait BuildsTextRequests
             $body['max_output_tokens'] = $options->maxTokens;
         }
 
-        $body = array_merge($body, Arr::whereNotNull([
-            'temperature' => $options?->temperature,
-            'top_p' => $options?->topP,
-        ]));
+        if ($this->modelSupportsTemperature($model)) {
+            $body = array_merge($body, Arr::whereNotNull([
+                'temperature' => $options?->temperature,
+                'top_p' => $options?->topP,
+            ]));
+        }
 
         $providerOptions = $options?->providerOptions(
             Lab::tryFrom($provider->driver()) ?? $provider->driver()
@@ -53,6 +55,15 @@ trait BuildsTextRequests
         }
 
         return $body;
+    }
+
+    /**
+     * Determine whether the given model supports temperature / top_p parameters.
+     * OpenAI reasoning models (o1, o3 series) reject these parameters.
+     */
+    protected function modelSupportsTemperature(string $model): bool
+    {
+        return ! preg_match('/^o\d/', $model);
     }
 
     /**
--- a/src/Gateway/OpenAi/Concerns/HandlesTextStreaming.php
+++ b/src/Gateway/OpenAi/Concerns/HandlesTextStreaming.php
@@ -377,10 +377,15 @@ trait HandlesTextStreaming
             }
 
             $body = array_merge($body, Arr::whereNotNull([
-                'temperature' => $options?->temperature,
-                'top_p' => $options?->topP,
                 'max_output_tokens' => $options?->maxTokens,
             ]));
+
+            if ($this->modelSupportsTemperature($model)) {
+                $body = array_merge($body, Arr::whereNotNull([
+                    'temperature' => $options?->temperature,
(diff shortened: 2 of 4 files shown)


src/Gateway/OpenAi/Concerns/BuildsTextRequests.php
```diff
@@ -55,6 +57,15 @@ protected function buildTextRequestBody(
...
     }
 
+    /**
+     * Determine whether the given model supports temperature / top_p parameters.
+     * OpenAI reasoning models (o1, o3 series) reject these parameters.
+     */
+    protected function modelSupportsTemperature(string $model): bool
+    {
+        return ! preg_match('/^o\d/', $model);
+    }
```
pushpak1300, line 67: This is not a correct detection. gpt-5.4-nano also doesn't support temperature.

pushpak1300:
I think without maintaining the list of models internally we can't solve this reliably but at the same time we don't want to maintain this list.

PR #554, opened 2026-05-09
Add PHPStan

pushpak1300:
The package has no static analyzer wired up, so type-level bugs (undefined methods, unsafe `new static()`, protected property access, mismatched docblocks) only surface at runtime. This PR introduces PHPStan at level 0 and fixes the issues it caught.


dumbbellcode:
about time!

PR #555, opened 2026-05-09
Support provider options in embeddings

### LINKED ISSUE #508: Voyage AI - Add Request Parameter Support
In my application, I'm generating vector embeddings using Laravel's Stringable class.
```
Str::of($input)->toEmbeddings(
    provider: Lab::VoyageAI,
    dimensions: 1024,
    model: 'voyage-4-large'
);
```

According to the Voyage AI API reference documentation [Text embedding models](https://docs.voyageai.com/reference/embeddings-api), it supports a few additional request body parameters. But I couldn't find a way to pass these parameters to the `toEmbeddings` method.

As per my use case, I like to use the `input_type` parameter to specify whether the embedding is for a document or a query.
I also like to use the `output_dtype` parameter to change the data type for the embeddings to be returned.

Could we please have the additional request parameter support?

**Tech Stack**
- PHP Version: 8.4.20
- Laravel Version: 12.24.0
- Laavel AI Version: 0.6.6

dumbbellcode:
Adds support for provider-specific options when generating embeddings. Every provider has knobs the SDK was hardcoding or omitting — Cohere and Voyage hardcode `input_type: 'search_document'`, so today there is no way to send `search_query` for the query side of a RAG retrieval. This patches that escape hatch and adds per-provider scoping for failover.

Supports: #508
Is useful for: #358

### Usage

Same options for whichever provider runs:

```php
// Indexing pass — documents
Embeddings::for($chunks)
    ->providerOptions(['input_type' => 'search_document'])
    ->generate(provider: 'cohere', model: 'embed-v4.0');

// Query pass — MUST be search_query for retrieval to work correctly
Embeddings::for([$userQuery])
    ->providerOptions(['input_type' => 'search_query'])
    ->generate(provider: 'cohere', model: 'embed-v4.0');
```

Per-provider options for failover (each provider names the same knob differently):

```php
Embeddings::for([$userQuery])
    ->providerOptions(fn ($provider) => match ($provider->driver()) {
        'cohere'   => ['input_type' => 'search_query'],
        'voyageai' => ['input_type' => 'query'],
        'jina'     => ['task' => 'retrieval.query'],
        default    => [],
    })
    ->generate(provider: ['cohere', 'voyageai', 'jina']);
```

Via the Stringable macro:

```php
Str::of($userQuery)->toEmbeddings(
    provider: Lab::VoyageAI,
    dimensions: 1024,
    model: 'voyage-4-large',
    providerOptions: ['input_type' => 'query'],
);
```

### Approach

- Threaded a `providerOptions` parameter through the embeddings pipeline (contracts, providers, gateways, pending responses, prompts, fake) so it reaches the HTTP body of every provider.
- Provisioned every embeddings gateway: OpenAI, Azure OpenAI, Cohere, Voyage, Jina, Mistral, Gemini, Ollama, OpenRouter, and Bedrock (both Titan and Cohere branches).
- Framework-controlled keys (`model`, `input`, `dimensions`) always win over user-supplied options; provider defaults that callers might legitimately want to override (Cohere `input_type`, Jina `task`) remain overridable.
- Gemini merges options into each per-input request, matching its `batchEmbedContents` contract where `taskType`, `title`, and `outputDimensionality` live per request, not top-level.
- Cache keys mix in a deterministic, key-order-insensitive fingerprint of the resolved options so `input_type=query` and `input_type=document` no longer collide.
- `providerOptions()` accepts an array or a closure resolver that receives the active provider, so failover targets get options scoped to their own API.
- Options are carried on `EmbeddingsPrompt`, `QueuedEmbeddingsPrompt`, and the fake gateway so they're assertable in tests and visible to event listeners.

### DIFF
--- a/src/Gateway/VoyageAi/VoyageAiGateway.php
+++ b/src/Gateway/VoyageAi/VoyageAiGateway.php
@@ -22,6 +22,7 @@ class VoyageAiGateway implements EmbeddingGateway, RerankingGateway
      * Generate embedding vectors representing the given inputs.
      *
      * @param  string[]  $inputs
+     * @param  array<string, mixed>  $providerOptions
      */
     public function generateEmbeddings(
         EmbeddingProvider $provider,
@@ -33,11 +34,11 @@ class VoyageAiGateway implements EmbeddingGateway, RerankingGateway
     ): EmbeddingsResponse {
         $data = $this->withErrorHandling(
             $provider->name(),
-            fn () => $this->client($provider, $timeout)->post('/embeddings', array_merge([
+            fn () => $this->client($provider, $timeout)->post('/embeddings', array_merge($providerOptions, [
                 'model' => $model,
                 'input' => $inputs,
                 'output_dimension' => $dimensions,
-            ], $providerOptions)),
+            ])),
         )->json();
 
         return new EmbeddingsResponse(
--- a/src/Gateway/VoyageAi/VoyageAiGateway.php
+++ b/src/Gateway/VoyageAi/VoyageAiGateway.php
@@ -29,14 +29,15 @@ class VoyageAiGateway implements EmbeddingGateway, RerankingGateway
         array $inputs,
         int $dimensions,
         int $timeout = 30,
+        array $providerOptions = [],
     ): EmbeddingsResponse {
         $data = $this->withErrorHandling(
             $provider->name(),
-            fn () => $this->client($provider, $timeout)->post('/embeddings', [
+            fn () => $this->client($provider, $timeout)->post('/embeddings', array_merge([
                 'model' => $model,
                 'input' => $inputs,
                 'output_dimension' => $dimensions,
-            ]),
+            ], $providerOptions)),
         )->json();
 
         return new EmbeddingsResponse(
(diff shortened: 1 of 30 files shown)


src/Gateway/VoyageAi/VoyageAiGateway.php
```diff
@@ -29,14 +29,15 @@ public function generateEmbeddings(
...
     ): EmbeddingsResponse {
         $data = $this->withErrorHandling(
             $provider->name(),
-            fn () => $this->client($provider, $timeout)->post('/embeddings', [
+            fn () => $this->client($provider, $timeout)->post('/embeddings', array_merge([
                 'model' => $model,
                 'input' => $inputs,
                 'output_dimension' => $dimensions,
-            ]),
+            ], $providerOptions)),
```
dumbbellcode: only merging the passed options for voyage as of now, as only that was requested, but we can do it for other providers when needed, let me know

pushpak1300: Yes i think merging this for all other provider should be done as that will be expected behaviour.

dumbbellcode: alright, have provisioned it for other providers now 👍

PR #556, opened 2026-05-09
Gemini context cache support via cachedContent param in agent provider options

### LINKED ISSUE #372: Feature Request: Cache Support for Gemini Gateway
I intend to migrate from prismphp to laravel ai sdk but the only thing missing between the two is crucially explicit cache support for gemini.

[Prism Implementation ](https://github.com/prism-php/prism/blob/main/src/Providers/Gemini/Handlers/Cache.php)
[Gemini Docs](https://ai.google.dev/gemini-api/docs/caching#generate-content)

dumbbellcode:
This change allows users to pass name of previously created cache in agent prompts via the cachedContent param, [see this](https://ai.google.dev/gemini-api/docs/caching#rest). 

Resolves https://github.com/laravel/ai/issues/372

### Usage

```php

// assuming user has already cached content
// pass the name of cache in agent provider options

{
    use Promptable;

    {
    }

    {
        return match ($provider) {
            Lab::Gemini => ['cachedContent' => '<name of cached item>'],
            default => [],
        };
    }
};

```

### Approach

1. The approach is slightly imperfect because all of the provider options are already being passed in Gemini API body param`generationConfig`, where as `cachedContent` is a different top level param.
2. To avoid breaking change i extracted this property from provider options and pass it in the request body.

### DIFF
--- a/src/Gateway/Gemini/Concerns/BuildsTextRequests.php
+++ b/src/Gateway/Gemini/Concerns/BuildsTextRequests.php
@@ -83,7 +83,16 @@ trait BuildsTextRequests
             'topP' => $options?->topP,
         ]));
 
-        $providerOptions = $options?->providerOptions(Lab::tryFrom($provider->driver()) ?? $provider->driver());
+        $providerOptions = $options?->providerOptions(Lab::tryFrom($provider->driver()) ?? $provider->driver()) ?? [];
+
+        // Hoist keys that need to be passed at top level, as everything else is passed in generationConfig
+        $topLevelKeys = ['cachedContent'];
+        foreach ($topLevelKeys as $key) {
+            if (array_key_exists($key, $providerOptions)) {
+                $body[$key] = $providerOptions[$key];
+                unset($providerOptions[$key]);
+            }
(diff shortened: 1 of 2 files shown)


src/Gateway/Gemini/Concerns/BuildsTextRequests.php
```diff
@@ -83,7 +83,16 @@ private function assembleRequestBody(
             'topP' => $options?->topP,
         ]));
 
-        $providerOptions = $options?->providerOptions(Lab::tryFrom($provider->driver()) ?? $provider->driver());
+        $providerOptions = $options?->providerOptions(Lab::tryFrom($provider->driver()) ?? $provider->driver()) ?? [];
+
+        // Hoist keys that need to be passed at top level, as everything else is passed in generationConfig
+        $topLevelKeys = ['cachedContent'];
+        foreach ($topLevelKeys as $key) {
```
dumbbellcode, line 90: seems like the best way to avoid braking changes

pushpak1300:
Thanks

PR #558, opened 2026-05-10
Fix buffering for server_tool_use input in streamed responses

CodeWrap:
`HandlesTextStreaming` was only buffering `input_json_delta` for client `tool_use` blocks.

That meant `server_tool_use` blocks like `web_search` and `advisor` could stream input deltas, but the SDK would drop them. By the time the completed `ProviderToolEvent` fired, and by the time the block was written into `$responseContent`, the tool input was still empty.

This change adds a parallel `$serverToolInputBuffers` accumulator for `server_tool_use` blocks and assembles the final input before emitting the completed `ProviderToolEvent`, following the same pattern we already use for client tools via `$pendingToolCalls`.

I also used `$currentBlockIndex` as the fallback index instead of `count($responseContent) - 1`, which is safer while blocks are still being streamed.

Includes a streaming test that feeds a `server_tool_use` block with multiple `input_json_delta` chunks and asserts that the completed `ProviderToolEvent` contains the fully assembled input.

I ran into this while trying to capture streamed `web_search` queries: the `ProviderToolEvent` was always arriving with empty input.

### DIFF
--- a/src/Gateway/Anthropic/Concerns/HandlesTextStreaming.php
+++ b/src/Gateway/Anthropic/Concerns/HandlesTextStreaming.php
@@ -60,6 +60,7 @@ trait HandlesTextStreaming
         $currentToolIndex = -1;
         $currentServerToolInput = '';
         $pendingToolCalls = [];
+        $serverToolInputBuffers = [];
         $responseContent = [];
 
         $inputTokens = 0;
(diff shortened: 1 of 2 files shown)


pushpak1300:
Thanks for digging into this! I tried to reproduce the underlying issue and I don't think this is quite the right fix — wanted to share what I found before we go further.

The premise that `server_tool_use` deltas are being dropped doesn't match what's in the code. We already buffer them today:

```php
} elseif ($currentBlockType === 'server_tool_use') {
    $currentServerToolInput .= $partial;
}
```

and assemble on `content_block_stop`:

```php
if ($currentServerToolInput !== '' && isset($responseContent[$index])) {
    $responseContent[$index]['input'] = json_decode($currentServerToolInput, true) ?? [];
}
```

I checked out `0.x` and ran the new test from this PR against it — it passes without your changes. So as written it's confirming existing behavior rather than guarding a regression.


Could you share the raw SSE stream (or a fixture) from the case where you saw empty input? If we can turn that into a test that **fails on `0.x`**, I'm happy to get the real fix in. The `$currentBlockIndex` fallback change is reasonable on its own, so we can keep that piece regardless.

PR #564, opened 2026-05-11
Validate empty inputs across generation entry points

Anoop-Kadachi:
`Audio::of()`, `Image::of()`, `Embeddings::for()`, `Reranking::of()`, and `Transcription::of()` / `fromBase64()` now reject empty and whitespace-only input with a clear `InvalidArgumentException` before hitting providers or the queue. The same guard now lives in `Base64Audio`, `LocalAudio`, `LocalImage`, `StoredAudio`, and `StoredImage` constructors, so empty paths / content fail fast instead of blowing up downstream.

Tests cover empty and whitespace-only strings for each API.

### ⚠️ Possible breaking change

Constructing any of the following with an empty value previously succeeded and failed lazily; it now throws `InvalidArgumentException` at construction time:

- `Audio::of('')`, `Image::of('')`, `Embeddings::for([])`, `Reranking::of([])`, `Transcription::of('')` / `Transcription::fromBase64('')`
- `new Base64Audio('')`, `new LocalAudio('')`, `new LocalImage('')`, `new StoredAudio('')`, `new StoredImage('')`

### DIFF
--- a/src/Audio.php
+++ b/src/Audio.php
@@ -11,13 +11,11 @@ class Audio
 {
     /**
      * Generate audio from the given text.
+     *
+     * @throws InvalidArgumentException if the given text is empty or whitespace-only.
      */
     public static function of(string $text): PendingAudioGeneration
     {
-        if (trim($text) === '') {
-            throw new InvalidArgumentException('Text content is required to generate audio.');
-        }
-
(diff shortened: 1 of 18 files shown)


pushpak1300:
Thanks.

PR #568, opened 2026-05-12
Introduce `Alias` Attribute for `Agent`

Junveloper:
This PR adds an `Alias` attribute that lets agent classes declare a stable identifier to be stored in the agent column on `agent_conversation_messages`, in place of the fully-qualified class name.

Currently, `DatabaseConversationStore` writes `$prompt->agent::class` to the agent column. This couples persisted conversation history to PHP class names, so renaming or relocating an agent class leaves historical rows referencing an outdated FQCN. By annotating an agent with `#[Alias('linear')]`, consumers can choose a stable identifier independent of class structure. Agents without the attribute continue to store their FQCN, so the change is fully backward compatible.

### DIFF
--- a/src/Attributes/Alias.php
+++ b/src/Attributes/Alias.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace Laravel\Ai\Attributes;
+
+use Attribute;
+
+#[Attribute(Attribute::TARGET_CLASS)]
+class Alias
+{
+    public function __construct(public string $value)
+    {
+        //
+    }
+}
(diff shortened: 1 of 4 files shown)


pushpak1300:
Thanks for your pull request!

pushpak1300:
```
class LinearAgent implements Agent {
    use Promptable;
    public function __toString(): string { return 'linear'; }
}
```

Junveloper:
@pushpak1300 Thank you for your suggestion. However, I am not sure how overriding __toString() will help when `::class` will resolve to FQCN. Laravel has MorphMap for Eloquent Model and I was wondering if you'd accept the PR if mapping was done in `config.php` instead?

PR #571, opened 2026-05-12
Add unit test for WrapsPcmAudio trait

mosabbirrakib:
The `WrapsPcmAudio` trait introduced in #559 wraps raw PCM audio in a WAV container so OpenRouter can accept it as `audio/wav`. The header construction is bit-precise (RIFF/WAVE/fmt/data chunks with little-endian packed integers), and a silent regression here would corrupt every PCM-source TTS or STT call routed through OpenRouter.

The trait is currently exercised only via integration through `OpenRouterGateway::transcribe()` and `GeminiGateway::generateAudio()`, which assert end-to-end behavior but not the byte layout. This PR adds a direct unit test that pins the header structure.

## Coverage added

- RIFF/WAVE container with `fmt ` and `data` chunks at the expected offsets, and PCM payload appended after the 44-byte header.
- `RIFF` chunk size and `data` chunk size correctly report `36 + dataSize` and `dataSize` respectively.
- Default PCM-format fields: mono, 16-bit, 24kHz (the format Gemini and OpenRouter currently emit), with derived `byteRate` and `blockAlign`.
- Custom sample rate / channels / bit depth (44.1kHz stereo 24-bit) correctly recompute `byteRate` and `blockAlign`.

## Testing

```
vendor/bin/pest tests/Unit/Gateway/Concerns/WrapsPcmAudioTest.php
# 4 passed (19 assertions)
```

### DIFF
--- a/tests/Unit/Gateway/Concerns/WrapsPcmAudioTest.php
+++ b/tests/Unit/Gateway/Concerns/WrapsPcmAudioTest.php
@@ -1,0 +1,62 @@
+<?php
+
+use Laravel\Ai\Gateway\Concerns\WrapsPcmAudio;
+
+beforeEach(function () {
+    $this->trait = new class
+    {
+        use WrapsPcmAudio {
+            pcmToWav as public;
+        }
+    };
+});
+
+test('wav header begins with RIFF/WAVE container and includes fmt and data chunks', function () {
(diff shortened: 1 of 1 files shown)


pushpak1300:
Thanks for your pull request!

PR #572, opened 2026-05-12
Add unit test for HandlesFailoverErrors trait

mosabbirrakib:
The `HandlesFailoverErrors` trait is the central error mapping layer used by every HTTP-based gateway in the package (15+ implementations). Its branches — 429 → `RateLimitedException`, configurable overloaded status codes → `ProviderOverloadedException`, configurable insufficient-credit patterns → `InsufficientCreditsException`, and passthrough for everything else — are all currently exercised only via provider-level integration tests.

This adds a direct unit test that pins the trait's behavior in isolation, including the precedence rules between branches and the protected extension-point defaults.

## Coverage added

- Callback result is returned when no exception is thrown.
- Non-`RequestException` errors propagate unchanged.
- 429 → `RateLimitedException` with the provider-name interpolated into the message.
- Default 503 → `ProviderOverloadedException`.
- Custom `overloadedStatusCodes()` (e.g., 529 for Anthropic) is honored, and the default 503 is *not* overloaded when overridden.
- Matching `insufficientCreditPatterns()` against `error.message` triggers `InsufficientCreditsException`.
- 429 takes precedence over insufficient-credit pattern matching when both could apply.
- Non-matching message body falls through to the original `RequestException`.
- Default `insufficientCreditPatterns()` is empty, so providers that don't override it never trigger the credit branch.

## Testing

```
vendor/bin/pest tests/Unit/Gateway/Concerns/HandlesFailoverErrorsTest.php
# 9 passed (12 assertions)
```

### DIFF
--- a/tests/Unit/Gateway/Concerns/HandlesFailoverErrorsTest.php
+++ b/tests/Unit/Gateway/Concerns/HandlesFailoverErrorsTest.php
@@ -1,0 +1,120 @@
+<?php
+
+use GuzzleHttp\Psr7\Response as Psr7Response;
+use Illuminate\Http\Client\RequestException;
+use Illuminate\Http\Client\Response;
+use Laravel\Ai\Exceptions\InsufficientCreditsException;
+use Laravel\Ai\Exceptions\ProviderOverloadedException;
+use Laravel\Ai\Exceptions\RateLimitedException;
+use Laravel\Ai\Gateway\Concerns\HandlesFailoverErrors;
+
+beforeEach(function () {
+    $this->gateway = new class
+    {
+        use HandlesFailoverErrors {
+            withErrorHandling as public;
+        }
+    };
+
+    $this->customGateway = new class
+    {
+        use HandlesFailoverErrors {
+            withErrorHandling as public;
+        }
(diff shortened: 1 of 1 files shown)


pushpak1300 (changes requested):
Most of the tests here mirror the implementation one-to-one — "if status is 429, throw `RateLimitedException`" tests a single `if` branch that the type signature already advertises. These are change-detector tests: if someone edits the trait, they'll edit these in lockstep, so they won't catch regressions, they'll just need updating.

The ones I think are actually pulling weight:

- `custom overloadedStatusCodes is honored over the default 503` — specifically that the default 503 stops being overloaded once you override. Easy to regress.
- `429 takes precedence over insufficient credit pattern matching` — precedence is the subtle thing worth pinning.
- `non-matching message is rethrown as the original RequestException` — the fallthrough behavior.

mosabbirrakib:
Trimmed to the three you called out — override-stops-default, precedence, and fallthrough. Dropped the change-detector tests that mirror each `if` branch one-to-one. Final file is 4 assertions across 3 tests.

PR #577, opened 2026-05-13
PHP 8.2 support

angelo-myhealthafrica:
# Summary

Adds support for PHP 8.2; useful for Laravel 12 users who are running on PHP 8.2.

The only feature in the  project I found that's not supported in PHP 8.2 is the use of `array_any`. This has been replaced with `foreach` loops in usage sites.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,68 +1,68 @@
 {
-    "name": "laravel/ai",
-    "description": "The official AI SDK for Laravel.",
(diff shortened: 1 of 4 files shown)


pushpak1300:
Thanks for your pull request!

pushpak1300:
This is mostly compatible with laravel v13.

PR #581, opened 2026-05-13
Pass mime type through Image::fromUrl() and Document::fromUrl()

mosabbirrakib:
`RemoteImage` and `RemoteDocument` constructors both accept an optional `\$mimeType`, but their static factories on `Image` and `Document` only forward the URL — so callers using the canonical entry point can't set the mime type without dropping down to `new RemoteImage(\$url, \$mimeType)` directly.

`Audio::fromUrl(string \$url, ?string \$mimeType = null)` already does this correctly, so this aligns the other two with the existing pattern.

### Before

```php
// src/Files/Image.php
public static function fromUrl(string \$url): RemoteImage
{
    return new RemoteImage(\$url);
}

// src/Files/Document.php
public static function fromUrl(string \$url): RemoteDocument
{
    return new RemoteDocument(\$url);
}
```

### After

```php
// src/Files/Image.php
public static function fromUrl(string \$url, ?string \$mimeType = null): RemoteImage
{
    return new RemoteImage(\$url, \$mimeType);
}

// src/Files/Document.php
public static function fromUrl(string \$url, ?string \$mimeType = null): RemoteDocument
{
    return new RemoteDocument(\$url, \$mimeType);
}
```

Backwards compatible: existing call sites that pass only a URL keep working unchanged.

### DIFF
--- a/src/Files/Document.php
+++ b/src/Files/Document.php
@@ -41,9 +41,9 @@ abstract class Document extends File
     /**
      * Create a new remote document using the document at the given URL.
      */
-    public static function fromUrl(string $url): RemoteDocument
+    public static function fromUrl(string $url, ?string $mimeType = null): RemoteDocument
     {
-        return new RemoteDocument($url);
+        return new RemoteDocument($url, $mimeType);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


pushpak1300:
I am closing this pull request because it lacks sufficient explanation, tests, or both. It is difficult for us to merge pull requests without these things because the change may introduce breaking changes to the this package.


Feel free to re-submit your change with a thorough explanation of the feature and tests. Thanks!

PR #584, opened 2026-05-13
Use consistent X generation wording

mosabbirrakib:
`AiManager` consistently throws `Provider [...] does not support X generation.` across all "X is not supported" error paths (audio generation, embedding generation, image generation, text generation, transcription generation). `AnthropicGateway` matches that wording for image and audio (`Anthropic does not support image generation.` / `Anthropic does not support audio generation.`) but breaks the pattern for transcription and embeddings:

```php
// before
throw new LogicException('Anthropic does not support transcription.');
throw new LogicException('Anthropic does not support embeddings.');

// after
throw new LogicException('Anthropic does not support transcription generation.');
throw new LogicException('Anthropic does not support embedding generation.');
```

Pure wording change to match the existing convention; no test asserts on the old strings, no behavior change.

### DIFF
--- a/src/Gateway/Anthropic/AnthropicGateway.php
+++ b/src/Gateway/Anthropic/AnthropicGateway.php
@@ -179,7 +179,7 @@ class AnthropicGateway implements Gateway
         int $timeout = 30,
         array $providerOptions = [],
     ): TranscriptionResponse {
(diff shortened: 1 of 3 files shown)


pushpak1300:
Thanks.

PR #587, opened 2026-05-13
Rehydrate attachments when loading conversation history

### LINKED ISSUE #218: Attachments are not part of the thread
You add attachments to the conversation, AI agents can read it initially, you send additional messages, the attachments are not in the thread anymore for the agent to see. This defeats the purpose of thread history. If there is a config or an option I am missing, but as of right now, it appears that the SDK has a limitation where it does not send the attachments in the message history. This appears to be a gap in the SDK. I was able to extend it so that I can include the attachments, but it just requires higher burden on the app side now to ensure it follows the practice moving forward.

sumaiazaman:
Conversations with attachments worked on the first turn but silently lost their images, documents, and audio on every follow-up. `DatabaseConversationStore` was writing the `attachments` JSON column on each user message but never reading it back, so historical user messages were returned as plain `Message` objects. Gateways gate attachment mapping on `instanceof UserMessage && attachments->isNotEmpty()`, so once the rehydration step was skipped the provider only ever saw text from turn two onwards.

Fixes #218

### Usage

```php
$agent = new VisionAgent;
$agent->forUser($user);

// Turn 1 — attachments live on the AgentPrompt, sent to the provider as expected.
$agent->prompt('What is in this photo?', [
    new RemoteImage('https://example.com/cat.jpg', 'image/jpeg'),
]);
// → "A tabby cat sitting on a windowsill, wearing a red collar."

// Turn 2 — history is loaded from the database.
$agent->continueLastConversation($user)->prompt('What colour was its collar?');

// Before this PR:
// The store returned a plain Message with no attachments, the gateway saw
// no image in the history, and the agent replied:
// → "I don't see an image in our conversation."
//
// After this PR:
// The store rehydrates the RemoteImage from the attachments column, returns a
// UserMessage, the gateway resends the image to the provider, and the agent replies:
// → "The cat's collar was red."
```

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Ai\Files;
 
+use InvalidArgumentException;
 use Laravel\Ai\Contracts\Files\HasName;
 
 abstract class File implements HasName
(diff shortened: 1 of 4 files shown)


pushpak1300:
Thanks.

PR #594, opened 2026-05-14
Reject blank transcription language

Anoop-Kadachi:
`PendingTranscriptionGeneration::language()` now rejects empty or whitespace-only values so misconfigured ISO language hints fail fast instead of being sent to providers.


pushpak1300:

PR #597, opened 2026-05-15
Reject blank audio voice and instructions

mosabbirrakib:
`PendingAudioGeneration::voice()` and `PendingAudioGeneration::instructions()` previously accepted empty or whitespace-only strings, which would silently land in the request payload and produce confusing provider errors (or, for `instructions`, attach an empty system-style hint). They now reject blank values with a clear `InvalidArgumentException` at the chain site, matching the validation pattern landed for image size/quality (#593) and transcription language (#594).

For `voice()`, the validation runs after the BackedEnum unwrap so an enum case backed by an empty string is rejected the same way as a literal `''`.

### Coverage added in `tests/Feature/AudioFakeTest.php`

- `voice('')` / `voice(" \t\n")` → `Audio voice cannot be blank.`
- `instructions('')` / `instructions(" \t\n")` → `Audio instructions cannot be blank.`

### DIFF
--- a/src/PendingResponses/PendingAudioGeneration.php
+++ b/src/PendingResponses/PendingAudioGeneration.php
@@ -39,7 +39,13 @@ class PendingAudioGeneration
      */
     public function voice(BackedEnum|string $voice): self
     {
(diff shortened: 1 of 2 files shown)


pushpak1300:

PR #598, opened 2026-05-15
Reject blank description in SimilaritySearch::withDescription()

mosabbirrakib:
`SimilaritySearch::withDescription()` previously accepted empty or whitespace-only strings, which then surface as a blank description on the rendered tool — agents would receive a tool with no description and have no idea when to call it.

Mirrors #592, which rejected blank model and column names on the same class, and matches the broader "reject blank X" pattern recently landed across `PendingImageGeneration`, `PendingEmbeddingsGeneration`, `PendingReranking`, and `PendingTranscriptionGeneration`.

### Coverage added in `tests/Feature/SimilaritySearchTest.php`

- `withDescription('')` → `A description is required for similarity search.`
- `withDescription("  \t\n")` → same message.

### DIFF
--- a/src/Tools/SimilaritySearch.php
+++ b/src/Tools/SimilaritySearch.php
@@ -100,6 +100,10 @@ class SimilaritySearch implements Tool
      */
     public function withDescription(string $description): self
     {
+        if (blank($description)) {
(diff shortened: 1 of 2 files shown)


pushpak1300:

PR #599, opened 2026-05-15
Reject blank MIME type in File::withMimeType()

mosabbirrakib:
`File::withMimeType()` previously accepted empty or whitespace-only strings. Setting a blank MIME type silently corrupts the attachment payload — providers either reject the request with a confusing error or, worse, treat the file as `application/octet-stream` and skip vision/audio processing.

Mirrors the validation pattern that recently landed for `PendingImageGeneration::size()` and `::quality()` (#593), `PendingTranscriptionGeneration::language()` (#594), and `SimilaritySearch` (#592). Applies at the abstract `File` base, so all 12 file subclasses inherit the same guard.

### Coverage added in `tests/Unit/Files/FileTest.php`

- `withMimeType('')` → `MIME type cannot be blank.`
- `withMimeType("  \t\n")` → same message.

### DIFF
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -79,6 +79,10 @@ abstract class File implements HasName
      */
     public function withMimeType(string $mimeType): static
     {
+        if (blank($mimeType)) {
+            throw new InvalidArgumentException('MIME type cannot be blank.');
+        }
+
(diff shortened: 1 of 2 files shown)


pushpak1300:

PR #602, opened 2026-05-15
Reject blank content in Base64Image and Base64Document constructors

mosabbirrakib:
`Base64Audio::__construct()` already rejects blank base64 content with a clear `InvalidArgumentException`, but the two sibling base64 file types (`Base64Image`, `Base64Document`) accept any string. An empty base64 attachment then ships to the provider as a zero-byte file, which surfaces as a confusing 4xx error from each provider.

This PR closes the gap so all three `Base64*` constructors validate consistently. Mirrors the path-validation pattern landed for `LocalAudio`/`StoredAudio`/`LocalImage`/`StoredImage` in #564 and the recent "reject blank X" wave (#590, #591, #593, #594).

### Changes

- `Base64Image::__construct()` — reject blank `\$base64` with `Base64 image content cannot be empty.`
- `Base64Document::__construct()` — reject blank `\$base64` with `Base64 document content cannot be empty.`
- New `tests/Unit/Files/Base64ContentTest.php` covers empty + whitespace-only for both classes.

### DIFF
--- a/src/Files/Base64Document.php
+++ b/src/Files/Base64Document.php
@@ -4,6 +4,7 @@ namespace Laravel\Ai\Files;
 
 use Illuminate\Contracts\Support\Arrayable;
 use Illuminate\Http\UploadedFile;
+use InvalidArgumentException;
 use JsonSerializable;
 use Laravel\Ai\Contracts\Files\StorableFile;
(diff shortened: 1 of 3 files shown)


pushpak1300:
Test are failing.

PR #604, opened 2026-05-15
Add tests for blank file path constructor validation

Anoop-Kadachi:
Adds unit tests for existing constructor validation that rejects empty or whitespace-only paths on local/stored document, image, and audio file classes (including behavior introduced in #600). No new validation logic.


pushpak1300:
thanks.

PR #605, opened 2026-05-15
Fail loudly when stored conversation attachments cannot be reconstructed

Anoop-Kadachi:
When reloading remembered conversations, unknown or invalid attachment types were silently dropped via `filter()` after `File::fromArray()` returned null. That could make user messages lose attachments without any error.

This throws an `InvalidArgumentException` instead, consistent with existing loud failures for malformed attachment JSON and missing required fields.


pushpak1300:
Thanks for your pull request!

PR #435, opened 2026-04-22
fix(gateway): fix DeepSeek usage stats

aaronlei:
- Add named parameters for readability
- Add cache write/read input token stats
- Add inference token details parsing
- Support completion_tokens_details extraction

PR #610, opened 2026-05-17
Capture missing usage details across OpenAI-shaped providers

sumaiazaman:
## Problem

Several OpenAI-compatible gateways were silently discarding nested fields from the `usage` object, causing `Usage` to underreport tokens for reasoning models and prompt-caching workloads. The same anti-pattern (streaming handlers inlining `new Usage(...)` instead of delegating to `extractUsage()`) had already caused this bug class for DeepSeek (#435).

## Fix

Extended the original Groq-only fix to every affected gateway so this bug class doesn't get re-shipped piecemeal in follow-up PRs.

**Groq** (`src/Gateway/Groq/Concerns/`)
- `ParsesTextResponses::extractUsage()` — read `completion_tokens_details.reasoning_tokens` (as `reasoningTokens`) and `prompt_tokens_details.cached_tokens` (as `cacheReadInputTokens`).
- `HandlesTextStreaming` — both inline `new Usage(...)` constructions replaced with `\$this->extractUsage(\$data)`.

**OpenRouter** (`src/Gateway/OpenRouter/Concerns/HandlesTextStreaming.php`)
- Both inline `new Usage(...)` constructions (which redundantly inlined the four-field cache/reasoning parsing) replaced with `\$this->extractUsage(\$data)`. No behaviour change today; prevents future field additions from being dropped on the streaming path.

**Mistral** (`src/Gateway/Mistral/Concerns/HandlesTextStreaming.php`)
- Both inline `new Usage(...)` constructions replaced with `\$this->extractUsage(\$data)`. Same preventative cleanup.

## Field-name verification

All field names verified against authoritative provider sources:

| Field | Source |
|---|---|
| Groq `completion_tokens_details` / `prompt_tokens_details` | Groq Chat Completions API reference — both documented as nested objects on the `usage` schema |
| Groq `prompt_tokens_details.cached_tokens` | Groq prompt-caching docs (explicit JSON example: `{"prompt_tokens": 4641, ..., "prompt_tokens_details": {"cached_tokens": 4608}}`) |
| Groq `completion_tokens_details.reasoning_tokens` | OpenAI-compatible naming Groq mirrors; same field name as DeepSeek (#435) and OpenRouter |
| OpenRouter `prompt_tokens_details.{cached_tokens,cache_write_tokens}`, `completion_tokens_details.reasoning_tokens` | OpenRouter OpenAPI spec (`ChatUsage` schema) |
| Mistral usage fields | Mistral official Python SDK `UsageInfo` model — confirms only basic fields exist; no speculative additions |

### DIFF
--- a/src/Gateway/Groq/Concerns/HandlesTextStreaming.php
+++ b/src/Gateway/Groq/Concerns/HandlesTextStreaming.php
@@ -66,10 +66,7 @@ trait HandlesTextStreaming
 
             if (! $choice) {
                 if (isset($data['usage'])) {
-                    $usage = new Usage(
-                        $data['usage']['prompt_tokens'] ?? 0,
-                        $data['usage']['completion_tokens'] ?? 0,
-                    );
+                    $usage = $this->extractUsage($data);
                 }
 
                 continue;
@@ -132,10 +129,7 @@ trait HandlesTextStreaming
             }
 
             if (isset($data['usage'])) {
-                $usage = new Usage(
-                    $data['usage']['prompt_tokens'] ?? 0,
-                    $data['usage']['completion_tokens'] ?? 0,
-                );
+                $usage = $this->extractUsage($data);
             }
         }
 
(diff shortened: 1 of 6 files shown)


pushpak1300:
Thanks.

PR #611, opened 2026-05-17
Pass `providerOptions` through to Bedrock Converse and capture reasoning content

Junveloper:
Currently, `TextGenerationOptions::providerOptions(Lab::Bedrock)` is resolved by the gateway but never merged into the Converse request. Any agent-level Converse parameter set through `providerOptions()` - including `additionalModelRequestFields` (needed to enable Claude Opus 4.7's summarized reasoning output), `guardrailConfig`, or similar - is silently dropped before the request is dispatched.

This PR merges those options into the request inside `buildConverseParameters`, so Bedrock-specific Converse parameters now flow through end-to-end.
  
Even with that fix in place, reasoning produced by the model still would not reach the application. The Bedrock streaming loop only emitted `TextDelta` and `ToolCallEvent` events - no `StreamStart`, `TextStart`/`TextEnd`, or `ReasoningStart`/`ReasoningDelta`/`ReasoningEnd`, all of which `AnthropicGateway` and `OpenAiGateway` already emit and `reasoningContent` blocks were silently discarded in both streaming and non-streaming flows. As a result, consumers had no way to observe the chain of thought the model used to arrive at a response or tool call.

This PR aligns `BedrockTextGateway` with the other reasoning-capable gateways: the missing streaming events are emitted, every response block (including `reasoningContent` and `redactedContent`) is captured into `AssistantMessage::providerContentBlocks`, and those blocks are round-tripped verbatim through `formatAssistantMessage` and `buildAssistantConversationMessage` - mirroring the fast path already used by `AnthropicGateway`.

Bedrock's Converse API normalizes reasoning into a single [`ReasoningContentBlock` (https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ReasoningContentBlock.html) shape regardless of the underlying model, so these gateway changes apply uniformly across every reasoning-capable model on Bedrock, not just Anthropic's Claude family.


james-astalty:
Without this, we'll be unable to use this package, so this would be a much welcomed feature!

pushpak1300 (approved):

PR #616, opened 2026-05-17
Adopt larastan and resolve PHPStan level 2 errors

pushpak1300:
Bumping phpstan to level 2 surfaced 180 errors, most of them Laravel-shaped: `Storage::disk()->mimeType()` on the `Filesystem` contract, higher-order collection proxies (`->map->delta->join('')`), facade returns, etc. Fixing those with vanilla phpstan means writing workarounds that obscure intent.

Switching to larastan resolves the Laravel-specific noise outright and surfaces only the real structural gaps in the codebase, which this PR addresses.

### Real fixes (not larastan stubs)

A new base contract makes the four methods every provider exposes part of the type system instead of being only on the abstract `Providers\Provider` class:

```php
// before: gateways accept FileProvider but call methods on the abstract class
$provider->name(); // PHPStan: method not found on FileProvider

// after: Contracts\Providers\Provider declares name(), driver(),
// providerCredentials(), additionalConfiguration(). Every capability
// contract extends it.
```

A new `Contracts\RemembersConversation` interface replaces `class_uses_recursive` sniffing for conversation-aware agents:

```php
// before
if (in_array(RemembersConversations::class, class_uses_recursive($agent))
    && $agent->hasConversationParticipant()) { ... }

// after
if ($agent instanceof RemembersConversation
    && $agent->hasConversationParticipant()) { ... }
```

BC note: `StoreProvider::addFileToStore` gains an optional `array $metadata = []` parameter to match the `ManagesStores` trait implementation. Third-party implementers will need to add it.


danielhe4rt:
Thought all laravel packages we're enforced on phpstan lvl.5 minimum by default 

this will break any contract?

PR #617, opened 2026-05-18
Fix transcription usage mapping for output tokens

Anoop-Kadachi:
## Summary
- Map transcription `Usage::completionTokens` from `output_tokens` (or `completion_tokens`) instead of `total_tokens`
- Fixes inflated completion token counts for OpenAI and OpenRouter when the API returns separate output/total fields
- Fall back to `total_tokens - input_tokens` when only legacy usage fields are present

## Test plan
- [x] `tests/Unit/Gateway/ExtractsTranscriptionUsageTest.php`
- [x] `tests/Feature/Providers/OpenAi/TranscriptionTest.php`
- [x] `tests/Feature/Providers/OpenRouter/TranscriptionTest.php`

### DIFF
--- a/src/Gateway/Concerns/ExtractsTranscriptionUsage.php
+++ b/src/Gateway/Concerns/ExtractsTranscriptionUsage.php
@@ -1,0 +1,23 @@
+<?php
+
+namespace Laravel\Ai\Gateway\Concerns;
+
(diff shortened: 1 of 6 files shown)


pushpak1300:
Thanks for your pull request!

PR #620, opened 2026-05-19
Support Inline Conversation on OpenAI tool follow-ups for Zero Data Retention

Junveloper:
Currently, every OpenAI tool-call follow-up - both streaming (`handleStreamingToolCalls`) and non-streaming (`continueWithToolResults`) - POSTs to `/responses` with `previous_response_id` pointing at the prior assistant turn, relying on OpenAI to load that turn's state server-side. 

Organizations with Zero Data Retention enabled reject any request containing `previous_response_id` with `Previous response cannot be used for this organization due to Zero Data Retention`, because ZDR forbids OpenAI from storing the prior response that the ID would reference. As a result, any agent that uses tools is unusable on a ZDR account: the first response succeeds, the tool runs, then the follow-up is rejected with a 400.

This PR introduces a `zero_data_retention` provider config (driven by `OPENAI_ZERO_DATA_RETENTION`, default `false`) and a new `buildToolFollowUpBody` helper in `BuildsTextRequests` that picks between the two body shapes. When the flag is on, the follow-up omits `previous_response_id` and sends the full conversation inline - system instructions, prior user messages, the assistant tool calls produced this turn (with their reasoning blocks), and the tool results - through `mapMessagesToInput`, so OpenAI has the same context server-side chaining would have given it.

Both paths now thread a `priorMessages` array through the recursive tool loop, so the inline conversation accumulates correctly across multi-step tool chains. The streaming path reuses `mapAssistantMessage` via an `AssistantMessage` constructed from the streamed `ToolCall` DTOs - rather than re-implementing the OpenAI input shape inline - keeping a single source of truth for how an assistant turn with tool calls and reasoning gets serialized.

The flag is per-provider in `config/ai.php`, so accounts with mixed ZDR settings can be modeled by registering separate provider entries.

### DIFF
--- a/config/ai.php
+++ b/config/ai.php
@@ -122,6 +122,7 @@ return [
             'driver' => 'openai',
             'key' => env('OPENAI_API_KEY'),
             'url' => env('OPENAI_URL', 'https://api.openai.com/v1'),
+            'zero_data_retention' => env('OPENAI_ZERO_DATA_RETENTION', false),
         ],
 
         'openrouter' => [
--- a/src/Gateway/OpenAi/Concerns/BuildsTextRequests.php
+++ b/src/Gateway/OpenAi/Concerns/BuildsTextRequests.php
@@ -54,6 +54,34 @@ trait BuildsTextRequests
     }
 
     /**
+     * Build the request body for a tool-result follow-up request.
+     *
+     * Zero Data Retention accounts reject `previous_response_id`, so the
+     * full conversation is sent inline instead of relying on server-side chaining.
+     */
+    protected function buildToolFollowUpBody(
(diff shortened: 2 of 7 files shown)


pushpak1300:
Can you share any documentation around this ?

Junveloper:
Hey @pushpak1300, I will close this one and open another one. I hadn't realised there is another option provided by OpenAI for this exact use case as shown in `https://developers.openai.com/api/docs/guides/migrate-to-responses`

Junveloper:
Hey @pushpak1300  https://github.com/laravel/ai/pull/622 in favour of this PR :)

PR #621, opened 2026-05-19
Fix stored tool conversation replay order

gracjankubicki:
## Summary

This PR fixes the replay order for stored assistant messages that contain tool calls, tool results, and final assistant content.

`DatabaseConversationStore` stores an assistant response as a single database row containing:

- final assistant text in `content`
- tool calls in `tool_calls`
- tool results in `tool_results`

Before this change, that stored aggregate row was rehydrated as:

1. an assistant message containing both `tool_calls` and final `content`
2. a tool result message

That produces an invalid or misleading tool-calling history:

1. the model requested a tool call
2. the model already produced the final answer
3. the application only then supplied the tool result

This PR rehydrates the stored row into the same logical order produced during live tool execution:

1. assistant message with tool calls
2. tool result message
3. assistant message with the final text

The database storage format stays unchanged. Only the in-memory replay order returned by `DatabaseConversationStore::getLatestConversationMessages()` changes.

## Why this order matters

This is not specific to one provider. The issue is in the provider-agnostic conversation store, and the rehydrated messages are later mapped into each provider's API format.

The correct flow for tool calling is consistently:

```text
assistant tool call -> tool result -> assistant final answer
```

### OpenAI

Docs:

- https://platform.openai.com/docs/guides/function-calling?api-mode=responses
- https://developers.openai.com/api/docs/guides/migrate-to-responses

Package code:

- `src/Gateway/OpenAi/Concerns/MapsMessages.php`

OpenAI Responses API represents tool calls as `function_call` items and tool outputs as `function_call_output` items linked by `call_id`. Replaying final assistant text before `function_call_output` breaks the logical Responses API sequence.

### Azure OpenAI

Docs:

- https://learn.microsoft.com/en-us/azure/ai-services/openai/how-to/function-calling
- https://learn.microsoft.com/en-us/azure/cognitive-services/openai/quickstart

Package code:

- `src/Gateway/AzureOpenAi/AzureOpenAiGateway.php`

Azure OpenAI uses the same OpenAI message mapping concerns in this package, so it has the same replay-order requirement as OpenAI Responses-compatible flows.

### xAI

Docs:

- https://docs.x.ai/developers/tools/function-calling

Package code:

- `src/Gateway/Xai/Concerns/MapsMessages.php`

xAI Responses API also uses `function_call_output` linked by `call_id`. Tool output must be replayed before the final assistant text.

### Anthropic

Docs:

- https://docs.claude.com/en/docs/tool-use

Package code:

- `src/Gateway/Anthropic/Concerns/MapsMessages.php`

Anthropic maps tool requests as `tool_use` blocks and tool outputs as `tool_result` blocks. The final assistant text should not be replayed in the same logical step as the `tool_use` when the corresponding `tool_result` has not yet appeared.

### Gemini

Docs:

- https://ai.google.dev/gemini-api/docs/function-calling

Package code:

- `src/Gateway/Gemini/Concerns/MapsMessages.php`

Gemini uses `functionCall` and function response parts. Rehydrating the final text as a separate model message after the function response better matches the function-calling turn sequence.

### Groq

Docs:

- https://console.groq.com/docs/tool-use/local-tool-calling

Package code:

- `src/Gateway/Groq/Concerns/MapsMessages.php`

Groq follows a Chat Completions-like flow where an assistant message with `tool_calls` is followed by `role: tool`, and then by the final assistant response.

### Mistral

Docs:

- https://docs.mistral.ai/capabilities/function_calling

Package code:

- `src/Gateway/Mistral/Concerns/MapsMessages.php`

Mistral documents the function-calling conversation flow as assistant function call, tool result, then assistant answer. This PR makes stored conversation replay match that order.

### DeepSeek

Docs:

- https://api-docs.deepseek.com/guides/tool_calls

Package code:

- `src/Gateway/DeepSeek/Concerns/MapsMessages.php`

DeepSeek uses a Chat Completions-like tool-calling format. Tool result messages should follow the assistant tool call and precede the final assistant answer.

### OpenRouter

Docs:

- https://openrouter.ai/docs/guides/features/tool-calling

Package code:

- `src/Gateway/OpenRouter/Concerns/MapsMessages.php`

OpenRouter standardizes tool calling across models and providers using a Chat Completions-like interface. Replaying the stored history in tool-call order avoids passing a misleading message sequence to routed models.

### Ollama

Docs:

- https://docs.ollama.com/capabilities/tool-calling

Package code:

- `src/Gateway/Ollama/Concerns/MapsMessages.php`

Ollama's tool-calling loop appends the model message with tool calls, appends the tool result, and then continues the loop until the model produces the final answer. This PR makes database replay follow that same order.

### Amazon Bedrock

Docs:

- https://docs.aws.amazon.com/bedrock/latest/userguide/tool-use.html
- https://docs.aws.amazon.com/bedrock/latest/APIReference/API_runtime_ToolResultBlock.html

Package code:

- `src/Gateway/Bedrock/BedrockTextGateway.php`

Bedrock Converse API uses `toolUse` and `toolResult`, linked by `toolUseId`. A tool result is the response to a previous model tool request, so the final assistant response should be replayed after the `toolResult`.

## Scope

This change intentionally keeps the PR small:

- no database schema changes
- no storage format changes
- no public `ConversationStore` contract changes
- no provider gateway changes
- only stored conversation rehydration changes

Stored tool conversations now replay as:

```php
AssistantMessage('', $toolCalls)
ToolResultMessage($toolResults) // when present
AssistantMessage($content) // when present
```

## Tests

Added and updated storage tests to assert that stored tool conversations replay in the logical tool-calling order.

Local verification:

- `composer test -- --filter DatabaseConversationStoreTest`
- `composer test:lint`
- `composer test`
- `git diff --check`

Full local suite result: 1392 passed, 182 skipped due to missing external provider API keys.

### DIFF
--- a/src/Storage/DatabaseConversationStore.php
+++ b/src/Storage/DatabaseConversationStore.php
@@ -156,20 +156,26 @@ class DatabaseConversationStore implements ConversationStore
                 }
 
                 if ($toolCalls->isNotEmpty()) {
+                    if ($toolResults->isEmpty()) {
+                        return [
+                            new AssistantMessage(
+                                $record->content,
+                                $toolCalls->map(ToolCall::fromArray(...)),
+                            ),
+                        ];
+                    }
+
                     $messages = [
                         new AssistantMessage(
                             '',
                             $toolCalls->map(ToolCall::fromArray(...)),
                         ),
+                        new ToolResultMessage(
+                            $toolResults->map(ToolResult::fromArray(...)),
+                        ),
                     ];
 
-                    if ($toolResults->isNotEmpty()) {
-                        $messages[] = new ToolResultMessage(
-                            $toolResults->map(ToolResult::fromArray(...)),
-                        );
-                    }
-
-                    if ($record->content !== '') {
+                    if (filled($record->content)) {
                         $messages[] = new AssistantMessage($record->content);
                     }
 
--- a/src/Storage/DatabaseConversationStore.php
+++ b/src/Storage/DatabaseConversationStore.php
@@ -157,12 +157,9 @@ class DatabaseConversationStore implements ConversationStore
 
                 if ($toolCalls->isNotEmpty()) {
                     if ($toolResults->isEmpty()) {
-                        return [
-                            new AssistantMessage(
-                                $record->content,
-                                $toolCalls->map(ToolCall::fromArray(...)),
-                            ),
-                        ];
+                        return filled($record->content)
+                            ? [new AssistantMessage($record->content)]
+                            : [];
                     }
 
                     $messages = [
--- a/src/Storage/DatabaseConversationStore.php
+++ b/src/Storage/DatabaseConversationStore.php
@@ -159,7 +159,7 @@ class DatabaseConversationStore implements ConversationStore
                     $messages = [];
 
                     $messages[] = new AssistantMessage(
-                        $record->content ?: '',
+                        '',
                         $toolCalls->map(fn ($toolCall) => new ToolCall(
                             id: $toolCall['id'],
                             name: $toolCall['name'],
@@ -180,6 +180,10 @@ class DatabaseConversationStore implements ConversationStore
                                 resultId: $toolResult['result_id'] ?? null,
                             ))
                         );
+                    }
+
+                    if ($record->content !== '') {
+                        $messages[] = new AssistantMessage($record->content);
                     }
 
                     return $messages;
--- a/tests/Feature/Storage/DatabaseConversationStoreTest.php
+++ b/tests/Feature/Storage/DatabaseConversationStoreTest.php
@@ -217,6 +217,36 @@ test('it replays stored tool conversations before the final assistant response',
         ->and($messages[2]->toolCalls)->toBeEmpty();
 });
 
+test('it keeps the final text on the tool call message when no tool results are stored', function () {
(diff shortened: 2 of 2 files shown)


pushpak1300:
**Follow-up: hardened the resultless-tool-call edge**

While reviewing, I found an edge the original split didn't cover. When a stored assistant row has tool calls but **no** tool results (reachable when the final step is capped by `maxSteps`, so `continuationToolResults()` returns `[]`), the unconditional split produced two consecutive assistant messages:

```
AssistantMessage('', toolCalls)
AssistantMessage(content)
```

with no `ToolResultMessage` between them. Strict role-alternation providers (e.g. Gemini, which emits one `model` entry per assistant message) reject that with `INVALID_ARGUMENT`.

Fixed in two commits:
- `19797e7` — guarded the split so the trailing text message is only added when a `ToolResultMessage` is actually present.
- `038bf7f` — aligned with Vercel AI SDK's `convertToModelMessages`, which **skips resultless (`input-available`) tool parts entirely**. We now drop the dangling tool call and replay only the final text (or nothing, if there's no text), which also removes the dangling-`tool_use` risk on Anthropic.

Also switched `$record->content !== ''` to `filled()` to match the codebase convention. Added tests for both the with-text and no-text resultless cases.

PR #622, opened 2026-05-19
Support OpenAI Zero Data Retention via encrypted reasoning

Junveloper:
Currently, every OpenAI tool-call follow-up — both streaming (`handleStreamingToolCalls`) and non-streaming (`continueWithToolResults`) — POSTs to `/responses` with `previous_response_id` pointing at the prior assistant turn, relying on OpenAI to load that turn's state server-side.

Organizations with Zero Data Retention enabled reject any request containing `previous_response_id` with:

> Previous response cannot be used for this organization due to Zero Data Retention

Because ZDR forbids OpenAI from storing the prior response that the ID would reference. As a result, any agent that uses tools is unusable on a ZDR account: the first response succeeds, the tool runs, then the follow-up is rejected with a 400.

This PR introduces an `encrypted_reasoning` provider config, driven by `OPENAI_ENCRYPTED_REASONING` and defaulting to `false`, that wires up [OpenAI's recommended stateless-with-reasoning path](https://developers.openai.com/api/docs/guides/migrate-to-responses#4-decide-when-to-use-statefulness):

- Every request gets `store: false`
- Every request gets `include: ['reasoning.encrypted_content']`
- Tool follow-ups omit `previous_response_id` entirely

Instead, the follow-up sends the full conversation inline through `mapMessagesToInput`, including:

- System instructions
- Prior user messages
- The assistant tool calls produced this turn
- Reasoning blocks carrying `encrypted_content`
- Tool results

This gives OpenAI the same context that server-side chaining would have provided. OpenAI decrypts the blob in memory, uses it to reconstitute reasoning, and discards it, so reasoning context is preserved across tool steps without anything being stored.

The `encrypted_content` blob is parsed off reasoning items in both `mapToolCallsWithReasoning` for non-streaming and `processTextStream` for streaming. It is carried on `ToolCall` via a new `reasoningEncryptedContent` field and re-emitted by `mapAssistantMessage` on replay.

As a result, the existing reasoning-block round-trip plumbing handles multi-step tool chains without any new threading helpers. Both paths now thread `$instructions` and the prior `$messages` collection through the recursive tool loop, so the inline conversation rebuilds the same system message and history that the initial request had.

The flag is per-provider in `config/ai.php`, so accounts with mixed ZDR settings can be modeled by registering separate provider entries.

### DIFF
--- a/config/ai.php
+++ b/config/ai.php
@@ -64,6 +64,7 @@ return [
             'deployment' => env('AZURE_OPENAI_DEPLOYMENT', 'gpt-4o'),
             'embedding_deployment' => env('AZURE_OPENAI_EMBEDDING_DEPLOYMENT', 'text-embedding-3-small'),
             'image_deployment' => env('AZURE_OPENAI_IMAGE_DEPLOYMENT', 'gpt-image-1'),
+            'store' => env('AZURE_OPENAI_STORE', true),
         ],
 
         'bedrock' => [
@@ -122,7 +123,7 @@ return [
             'driver' => 'openai',
             'key' => env('OPENAI_API_KEY'),
             'url' => env('OPENAI_URL', 'https://api.openai.com/v1'),
-            'encrypted_reasoning' => env('OPENAI_ENCRYPTED_REASONING', false),
+            'store' => env('OPENAI_STORE', true),
         ],
 
         'openrouter' => [
--- a/config/ai.php
+++ b/config/ai.php
@@ -122,6 +122,7 @@ return [
             'driver' => 'openai',
             'key' => env('OPENAI_API_KEY'),
             'url' => env('OPENAI_URL', 'https://api.openai.com/v1'),
+            'encrypted_reasoning' => env('OPENAI_ENCRYPTED_REASONING', false),
         ],
 
         'openrouter' => [
--- a/src/Gateway/AzureOpenAi/AzureOpenAiGateway.php
+++ b/src/Gateway/AzureOpenAi/AzureOpenAiGateway.php
@@ -84,7 +84,7 @@ class AzureOpenAiGateway implements EmbeddingGateway, ImageGateway, TextGateway
 
         $this->validateTextResponse($data);
 
-        return $this->parseTextResponse($data, $provider, filled($schema), $tools, $schema, $options, $timeout);
+        return $this->parseTextResponse($data, $provider, filled($schema), $tools, $schema, $options, $body, $timeout);
     }
 
     /**
@@ -128,6 +128,7 @@ class AzureOpenAiGateway implements EmbeddingGateway, ImageGateway, TextGateway
             $schema,
             $options,
             $response->getBody(),
+            $body,
             0,
             null,
             $timeout,
--- a/src/Gateway/OpenAi/Concerns/BuildsTextRequests.php
+++ b/src/Gateway/OpenAi/Concerns/BuildsTextRequests.php
@@ -50,15 +50,35 @@ trait BuildsTextRequests
             $body = array_merge($body, $providerOptions);
         }
 
-        if ($provider->additionalConfiguration()['encrypted_reasoning'] ?? false) {
+        if ($this->isStateless($provider)) {
             $body['store'] = false;
-            $body['include'] = array_values(array_unique([
-                ...($body['include'] ?? []),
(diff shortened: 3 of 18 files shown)


james-astalty:
Keen to see this merged as we rely on ZDR!

Junveloper:
Hey @pushpak1300  would you like me to also open an issue with a detailed response from openAI? It's reproducible when using API key with ZDR policy enabled at an organisation level and invoking a tool call from an agent. Really keen to get this moving forward! :)

pushpak1300:
Hey @Junveloper no need of separate issue but can you post some more info regarding this issue here ?

james-astalty:
@pushpak1300 we are unable to use the SDK with OpenAI currently with Zero Data Retention enabled. Is there anything we can do to get this one merged? What would you need in order to help get this one across the line?

Junveloper:
@pushpak1300  The organisation's API key settings have options for data retention policy as shown below

image

As seen in this document (https://developers.openai.com/api/docs/guides/migrate-to-responses#4-decide-when-to-use-statefulness), when this setting is turned on Response API cannot be used in a stateful way due to compliance or data retention policies. If you try to use Response API by sending `previous_response_id` which the package defaults to, you will get 400 response back with the error message `Previous response cannot be used for this organization due to Zero Data Retention`.

Essentially, the PR follow the official documentation on how this should be handled. Another documentation relating to ZDR is here (https://developers.openai.com/api/docs/guides/your-data#zero_data_retention).

As far as I am aware, the setting is only available for Organiszation API key and you'd have to ask the Support to enable the option to see the options. I am happy to provide a reproducible repository but you'd need to get a ZDR policy enabled API key from OpenAI to see it anyway.

Steps to reproduce is:

1. Enter OpenAI key with ZDR policy enabled.
2. Trigger an agent that has tools and ensure a tool is invoked.
3. When the tool call tries to return `ToolResult`  with `previous_response_id` to continue the original conversation, 400 error will be thrown.

pushpak1300:
Hey @Junveloper, thanks for the thorough investigation and the encrypted-reasoning round-trip plumbing.

I'll push an alternative approach on this branch: instead of an `OPENAI_ENCRYPTED_REASONING` flag that toggles between two code paths, every OpenAI request becomes stateless by default.  I tested this on the repo where it worked without any issues but still needed more tests as this kind of more bug breaking change.

```php
// before — every follow-up POST
['model' => $m, 'previous_response_id' => $id, 'input' => [/* just tool result */]]

// after — every follow-up POST
['model' => $m, 'input' => [/* full conversation, encrypted reasoning included */]]

Junveloper:
Hey @pushpak1300 , not a problem. Wouldn't your suggested path increase input token? Not sure by how much but I am assuming it wouldn't be negligible.

pushpak1300:
https://github.com/laravel/ai/pull/660

@Junveloper Yes it will but we need by default that version off and handle previous response id correctly with some hook. let me think about it again to see what else can we do.

Junveloper:
@pushpak1300  My justification for introducing a config flag was that it's organisation's API setting specific and having default value being `false` would retain the existing behaviour of using `previous_response_id`. Passing the entire message trail just to make ZDR work felt excessive as some people may be sensitive about optimising tokens. 

Really interested in seeing what you come up with! Thank you for having a thorough look :)

pushpak1300:
I agree @Junveloper after thinking multiple option i keep coming back to this solution so probably worts adding this under flag. i am still debating my choice if we need to document this config variable in `ai.php`. let me think about it more and we can get it merged soon. Let me know what do you think of this approach. @Junveloper ?

Junveloper:
Hi @pushpak1300 . Are you asking me if config flag should live elsewhere or whether the changes in your latest commit look good? 

For the first part, I can't think of a better place than where it is currently and flag name "store" suits! And changes/refactors seem much cleaner. Thank you!

pushpak1300:
Is it working with your organisation on this branch ?

Junveloper:
Just manually tested and it is working as intended! Would love to see it merged and released! :) Thank you so much @pushpak1300

Junveloper:
Hi @taylorotwell - I was wondering if you could please have a look at this PR. Many organisations will need to turn on Zero Data Retention policy (including ours) and not having this option available in the library is a major blocker for us. Thank you and your team as always :)

PR #623, opened 2026-05-19
Use long text columns for conversation payloads

gracjankubicki:
## Summary

This updates the default database conversation storage schema to use `longText` columns for conversation message payloads.

The following `agent_conversation_messages` columns now use `longText`:

- `content`
- `attachments`
- `tool_calls`
- `tool_results`
- `usage`
- `meta`

## Motivation

Agent conversation messages may contain large payloads, including long assistant responses, attachment metadata, tool calls, tool results, usage information, and provider metadata.

On MySQL-compatible databases, regular `text` columns can be too small for these payloads. Using `longText` gives the default conversation storage schema more room for larger AI conversation data without changing the public API.

## Upgrade note

New installations will receive `longText` payload columns from the default conversation migration.

Applications that have already published and run the previous conversation migration should update their published migration or add an application-level migration to alter the existing `agent_conversation_messages` payload columns from `text` to `longText`.

## Tests

- `composer test -- tests/Feature/Storage/DatabaseConversationStoreTest.php`
- `composer test:lint`
- `vendor/bin/phpstan --memory-limit=-1`
- `composer test`

### DIFF
--- a/database/migrations/2026_01_11_000001_create_agent_conversations_table.php
+++ b/database/migrations/2026_01_11_000001_create_agent_conversations_table.php
@@ -29,12 +29,12 @@ return new class extends AiMigration
             $table->foreignId('user_id')->nullable();
             $table->string('agent');
             $table->string('role', 25);
-            $table->text('content');
-            $table->text('attachments');
-            $table->text('tool_calls');
-            $table->text('tool_results');
-            $table->text('usage');
-            $table->text('meta');
+            $table->longText('content');
+            $table->longText('attachments');
+            $table->longText('tool_calls');
+            $table->longText('tool_results');
+            $table->longText('usage');
+            $table->longText('meta');
             $table->timestamps();
 
             $table->index(['conversation_id', 'user_id', 'updated_at'], 'conversation_index');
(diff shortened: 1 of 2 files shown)


database/migrations/2026_05_19_000002_update_agent_conversation_message_payload_columns_to_long_text.php
pushpak1300, line 1: hey migration provided by the sdk should be forward looking. we probably don't need this migration. we can add it in release notes though.

gracjankubicki: Sounds good, removed the follow-up migration and left the upgrade path as a release note in the PR description.

pushpak1300 (approved):

pushpak1300:
Thanks for the PR and the detailed write-up.

I’ve decided against merging this.  After reviewing it, the 64 KB text limit only becomes an issue in a very specific scenario mainly when base64-encoded attachments are stored inline and we haven’t actually encountered anyone hitting it.  Defaulting every payload column to longText to address this edge case feels like a poor trade-off for the default schema. It removes a useful safeguard and isn’t something the wider user base needs.

The conversation migration is now published. If your application genuinely requires more space, you can adjust the column types directly in your own published migration (or add an ->change() migration) and you’re all set – no library changes are necessary.

PR #624, opened 2026-05-19
Fail loudly on invalid stored conversation JSON

gracjankubicki:
## Summary

This updates database conversation storage to fail loudly when stored conversation JSON is invalid instead of silently treating it as empty data.

The database conversation store now uses `JSON_THROW_ON_ERROR` when encoding conversation payloads and when decoding stored attachment, tool call, and tool result JSON.

## Motivation

Stored conversation payloads can contain tool calls, tool results, attachment metadata, usage, and provider metadata. If one of these stored JSON payloads is truncated or otherwise invalid, `json_decode` previously returned `null`, which was then collected as an empty collection. In practice, this could silently drop stored tool calls or tool results during conversation replay.

Failing loudly makes corrupted or truncated stored JSON visible immediately instead of replaying incomplete conversation state.

## Notes

Sparse numeric-keyed legacy tool call and tool result payloads are still accepted and normalized with `values()`, preserving the existing compatibility behavior.

## Tests

- `composer test -- tests/Feature/Storage/DatabaseConversationStoreTest.php`
- `composer test:lint`
- `vendor/bin/phpstan --memory-limit=-1`
- `composer test`

### DIFF
--- a/src/Storage/DatabaseConversationStore.php
+++ b/src/Storage/DatabaseConversationStore.php
@@ -7,6 +7,7 @@ use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\DB;
 use Illuminate\Support\Str;
 use InvalidArgumentException;
+use JsonException;
 use Laravel\Ai\Contracts\ConversationStore;
 use Laravel\Ai\Files\File;
 use Laravel\Ai\Messages\AssistantMessage;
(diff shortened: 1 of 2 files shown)


gracjankubicki:
Closing this for now to rework the approach before submitting.

PR #625, opened 2026-05-19
Expose provider name as property on failover exceptions

sumaiazaman:
## Summary

`RateLimitedException`, `ProviderOverloadedException`, and `InsufficientCreditsException` all embed the provider name inside the exception message, but provide no way to access it programmatically. Catching code that needs to know which provider failed must parse the message string — which is fragile and couples application logic to a human-readable sentence.

This PR adds a `public readonly string $provider` property to all three failover exception classes. The `forProvider()` factory method is updated to populate it. The existing message format is unchanged.

Before:
```php
} catch (RateLimitedException $e) {
    // must parse: "Application rate limited by AI provider [anthropic]."
    preg_match('/\[(.+)\]/', $e->getMessage(), $m);
    $provider = $m[1];
}
```

After:
```php
} catch (RateLimitedException $e) {
    $provider = $e->provider; // 'anthropic'
}
```

## Changes

- `src/Exceptions/RateLimitedException.php` — add constructor + `provider` property
- `src/Exceptions/ProviderOverloadedException.php` — add constructor + `provider` property
- `src/Exceptions/InsufficientCreditsException.php` — add constructor + `provider` property
- `tests/Unit/Exceptions/FailoverExceptionProviderTest.php` — 4 assertions covering all three classes

## Test plan

- [x] All 4 new unit tests pass
- [x] All existing `HandlesFailoverErrorsTest` tests pass
- [x] All existing provider `ErrorHandlingTest` tests pass

### DIFF
--- a/src/Exceptions/InsufficientCreditsException.php
+++ b/src/Exceptions/InsufficientCreditsException.php
@@ -6,12 +6,22 @@ use Throwable;
 
 class InsufficientCreditsException extends AiException implements FailoverableException
 {
+    public function __construct(
+        public readonly string $provider,
+        string $message = '',
+        int $code = 0,
+        ?Throwable $previous = null
+    ) {
+        parent::__construct($message, $code, $previous);
+    }
+
     public static function forProvider(string $provider, int $code = 0, ?Throwable $previous = null): self
     {
         return new self(
+            $provider,
             'AI provider ['.$provider.'] has insufficient credits or quota.',
             $code,
-            $previous
+            $previous,
         );
     }
 }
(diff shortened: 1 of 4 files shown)


pushpak1300:
I want to make sure this solves a real problem. Looking at `Promptable::withModelFailover()`, `FailoverableException` is caught inside the SDK and the `AgentFailedOver` event fires with the full `Provider` object. The exception only escapes user code after the entire failover chain is exhausted — and in the non-failover case, the caller already knows which provider they invoked.

Can you share a concrete use case where `$e->provider` gives you something you can't already get from the variable you used to invoke the provider?

PR #628, opened 2026-05-19
Document @throws InvalidArgumentException on SimilaritySearch::usingModel

mosabbirrakib:
## Summary

- `SimilaritySearch::usingModel()` validates that both the model class name and vector column name are non-blank — throwing `InvalidArgumentException` when either is missing — but the docblock did not advertise that contract.
- Adds the missing `@throws InvalidArgumentException` annotation. Pairs with #592 which introduced the validation, and mirrors #585 / #576.

No behavior change — docblock only.

## Test plan

- [x] \`vendor/bin/pint src/Tools/SimilaritySearch.php\`
- [x] \`vendor/bin/pest tests/Feature/SimilaritySearchTest.php\` (all 6 tests pass)

### DIFF
--- a/src/Tools/SimilaritySearch.php
+++ b/src/Tools/SimilaritySearch.php
@@ -25,6 +25,8 @@ class SimilaritySearch implements Tool
 
     /**
      * Create a new similarity search tool instance.
+     *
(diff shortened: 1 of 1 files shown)


pushpak1300:
Closing because merged in #626

PR #629, opened 2026-05-19
Document @throws InvalidArgumentException on PendingReranking::__construct

mosabbirrakib:
## Summary

- `PendingReranking::__construct()` validates that the documents argument is a list and that each entry is a non-blank string (introduced in #591), throwing `InvalidArgumentException` otherwise, but the docblock did not advertise that contract.
- Adds the missing `@throws InvalidArgumentException` annotation. Mirrors the existing docblock on `PendingEmbeddingsGeneration::__construct`, which documents the same validation pattern.

No behavior change — docblock only.

## Test plan

- [x] \`vendor/bin/pint src/PendingResponses/PendingReranking.php\`
- [x] \`vendor/bin/pest --filter=Reranking\` (55 passed)

### DIFF
--- a/src/PendingResponses/PendingReranking.php
+++ b/src/PendingResponses/PendingReranking.php
@@ -21,6 +21,8 @@ class PendingReranking
      * Create a new pending reranking instance.
      *
      * @param  array<int, string>  $documents
+     *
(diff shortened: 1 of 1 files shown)


pushpak1300:
Closing because merged in #626

PR #632, opened 2026-05-20
fix: preserve all OpenAI text parts in tool loops

### LINKED ISSUE #503: Multiple text parts by message
In long tool loop workflows, it would be useful if the agent could return partial text outputs at each step instead of only a final message. This would improve visibility and UX during long-running processes.

mosabbirrakib:
## Summary
- preserve all `output_text` parts in OpenAI tool-loop assistant responses instead of only keeping one segment
- add regression coverage for multiple text parts within a single assistant message

## Test plan
- Not run in this environment (`php` and `composer` are not installed on the agent host)

Closes #503


pushpak1300:
@mosabbirrakib what exactly the problem being solved here ?

Issue #182, opened 2026-02-18
Add Per-provider model configuration for failover


lowbits:
### Problem

When using multiple providers as failover, there's no way to specify a different model per provider. Each provider typically requires its own model identifier (e.g. `gpt-5-mini` for OpenAI, `mistral-large-latest` for Mistral), but the current API only accepts a single model value.

### Current behavior

**Via attributes:**
```php
#[Provider([Lab::OpenAI, Lab::Mistral])]
#[Model('gpt-5-mini')]
class WorkoutProgrammerAgent implements Agent, Conversational, HasTools
```

`#[Model]` only accepts one value — there's no way to map a model to each provider.

**Via prompt method:**
```php
(new WorkoutProgrammerAgent($workoutPlan))
    ->prompt((string) $prompt, provider: [Lab::OpenAI, Lab::Mistral], model: 'gpt-5-mini');
```

Same limitation — `model` is a single string with no per-provider mapping.

### Expected behavior

Something like:

```php
#[Provider([Lab::OpenAI, Lab::Mistral])]
#[Model(['gpt-5-mini', 'mistral-large-latest'])]
```

or:

```php
->prompt($prompt, provider: [
    Lab::OpenAI => 'gpt-5-mini',
    Lab::Mistral => 'mistral-large-latest',
]);
```

Minoangelo:
Given that the failover functionality is actually really useful, I can only strongly support this recommendation. It would be great if this were to be implemented in the short-term :)

trantor-terminus:
Along the same lines if the providers are $provider => $model associative array in prompt() the code treats them as keyed per provider.

```
        foreach ($providers as $provider => $model) {
            $provider = Ai::fakeableImageProvider($provider);
```

So you can not really have a single provider with multiple models.  If I want to create a hierarchy of provider and models based on cost, speed and other attributes something like this ... [[{"gemini":"gemini-2.5-flash-image"},{"openai":"dall-e-2"},{"gemini":"gemini-3-pro-image-preview"},{"openai":"dall-e-3"}]] right now it is not possible. I would like to see something along the lines of 

```
        foreach ($providers as $pm) {
            $provider=key($pm); $model = $pm[$provider];
            $provider = Ai::fakeableImageProvider($provider);
```

This allows to build more than one failover hierarchy for a given provider.

PR #633, opened 2026-05-20
Apply model attribute and explicit model to all providers in a failover list

sumaiazaman:
## Summary

Fixes a bug where a shared model configured for an agent failover provider list was silently ignored.

This applies to the existing failover API where providers are tried in order:

```php
$agent->prompt(
    "Hello",
    provider: ["primary", "backup"],
    model: "gpt-4.1-mini",
);
```

or via attributes:

```php
#[Provider(["primary", "backup"])]
#[Model("gpt-4.1-mini")]
{
    use Promptable;
}
```

Before this change, array providers skipped agent-level model resolution, and numeric provider-list entries were normalized with `null` as the model. The SDK then fell back to each provider default model instead of using the explicitly configured shared model.

This is related to #182, but it does not add a new provider/model failover API. Per-provider models for unique providers continue to use the existing associative form:

```php
$agent->prompt("Hello", provider: [
    "openai" => "gpt-4.1-mini",
    "mistral" => "mistral-large-latest",
]);
```

## Current Failover Behavior

- Providers are attempted in the order provided.
- A mapped model is used when present.
- If no model is mapped, the provider default text model is used.
- Failover only happens for `FailoverableException`.
- If every provider fails, the last failover exception is thrown.
- Streaming failover only happens before any stream events have been emitted.

### DIFF
--- a/src/Promptable.php
+++ b/src/Promptable.php
@@ -254,7 +254,7 @@ trait Promptable
             }
         }
 
-        if (! is_array($provider) && is_null($model)) {
+        if (is_null($model)) {
             if (method_exists($this, 'model')) {
                 $model = $this->model();
             } else {
--- a/src/Providers/Provider.php
+++ b/src/Providers/Provider.php
@@ -61,9 +61,9 @@ abstract class Provider
             return [$providers => $model];
         }
 
-        return (new Collection($providers))->mapWithKeys(function ($value, $key) {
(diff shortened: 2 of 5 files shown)


pushpak1300:
Thanks

taylorotwell:
Reverting this as the associative approach was always the intended use.

PR #634, opened 2026-05-20
Pass cache_control through to Anthropic and Bedrock system prompts

timko212:
Lets agents mark the system prompt for caching via providerOptions:

    return ['cache' => [['target' => 'system']]];

Anthropic gets a cache_control: ephemeral block on the system string; Bedrock gets a cachePoint marker appended to the system list. The key is stripped from providerOptions before the rest is merged so it never ends up in the API payload.

### DIFF
--- a/src/Gateway/Anthropic/Concerns/BuildsTextRequests.php
+++ b/src/Gateway/Anthropic/Concerns/BuildsTextRequests.php
@@ -63,7 +63,34 @@ trait BuildsTextRequests
             'top_p' => $options?->topP,
         ]));
 
+        if (! empty($providerOptions['cache'])) {
+            $body = $this->withCacheControl($body, $providerOptions['cache']);
+            unset($providerOptions['cache']);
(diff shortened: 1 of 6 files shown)


pushpak1300:
Hey, thanks a lot for putting this together! 🙏

The Anthropic side here is already covered though — you can turn on auto prompt caching today without any changes, just by returning this from your agent:

```php
{
    return $provider === Lab::Anthropic
        ? ['cache_control' => ['type' => 'ephemeral']]
        : [];
}
```

That flows straight through to Anthropic and caches the system prompt (same ~90% savings). More background in #119.

So this part overlaps with what's already there, and I'd rather not add a second way to do the same thing.

The Bedrock piece is the one genuinely new bit, since Bedrock doesn't have that simple top-level switch. If you've got a real use case for Bedrock caching, I'd happily look at a smaller PR scoped just to that — would make it much easier to reason about and merge.

PR #637, opened 2026-05-20
Fix transcription usage mapping for output tokens

Anoop-Kadachi:
## Summary
- Map transcription completion usage from `output_tokens` / `completion_tokens` instead of `total_tokens`
- Fall back to `total_tokens - input_tokens` when only total usage is available
- Fix OpenAI and OpenRouter transcription usage parsing

## Test plan
- `vendor/bin/pint --dirty`
- `vendor/bin/pest tests/Unit/Gateway/ExtractsTranscriptionUsageTest.php tests/Feature/Providers/OpenAi/TranscriptionTest.php tests/Feature/Providers/OpenRouter/TranscriptionTest.php`


pushpak1300 (changes requested):
nice catch on the `total_tokens` vs `output_tokens` thing 👍

i think the trait can probably go away. both providers here only ever return `input_tokens`/`output_tokens`/`total_tokens` per the docs:

- openai openapi: https://github.com/openai/openai-openapi (transcription usage is either `{type: tokens, input_tokens, output_tokens, total_tokens, input_token_details}` or `{type: duration, seconds}`)
- openrouter STTUsage schema: https://openrouter.ai/docs/api/api-reference/transcriptions/create-audio-transcriptions (`cost`, `input_tokens`, `output_tokens`, `seconds`, `total_tokens`)

so the `?? completion_tokens` and `?? (total - input)` fallbacks aren't really reachable from either endpoint. could just inline:

```php
new Usage(
    $data['usage']['input_tokens'] ?? 0,
    $data['usage']['output_tokens'] ?? 0,
);
```

in both gateways and drop the trait + unit test.

PR #642, opened 2026-05-22
BedrockTextGateway: Add cache usage

liambenson:
This adds the ability for the BedrockTextGateway to track cache usage


pushpak1300:
Thanks

PR #650, opened 2026-05-22
Add tests for Promptable provider and model resolution precedence

mosabbirrakib:
## Summary

\`Promptable::getProvidersAndModels()\` decides which provider/model an agent prompt targets by walking a four-step chain — explicit argument → method on the agent → attribute on the class → \`config('ai.default')\`. None of those precedence edges had tests, and \`AgentAttributeTest\` only asserts that the \`#[Provider]\` attribute is _used_, not how it ranks against the other sources.

This PR adds tests that pin:

- explicit arguments win over both method and attribute
- attribute is used when no method exists
- method takes precedence over attribute when both are present
- **a method returning null short-circuits the attribute lookup and falls through to \`config('ai.default')\`** — this is a deliberate deviation from \`TextGenerationOptions::resolve()\`, which _does_ fall back to the attribute on a null method return. A refactor that unifies the two paths will fail this test.
- \`config('ai.default')\` is used when nothing else is set
- an array-form failover provider skips model resolution from method or attribute (the per-entry model is honored instead)

These are weight-pulling tests — they pin the precedence contract rather than mirror single-if branches.

## Test plan

- [x] \`vendor/bin/pint tests/Feature/AgentProviderModelResolutionTest.php\`
- [x] \`vendor/bin/pest tests/Feature/AgentProviderModelResolutionTest.php\` (6 passed)

### DIFF
--- a/tests/Feature/AgentProviderModelResolutionTest.php
+++ b/tests/Feature/AgentProviderModelResolutionTest.php
@@ -1,0 +1,116 @@
+<?php
+
+use Laravel\Ai\Attributes\Model;
+use Laravel\Ai\Attributes\Provider;
+use Laravel\Ai\Contracts\Agent;
+use Laravel\Ai\Promptable;
+
+function resolveProvidersAndModels(Agent $agent, mixed $provider = null, ?string $model = null): array
+{
+    $method = new ReflectionMethod($agent, 'getProvidersAndModels');
+    $method->setAccessible(true);
+
+    return $method->invoke($agent, $provider, $model);
+}
+
+#[Provider('anthropic')]
+#[Model('attribute-model')]
+class ProviderAttributesAgentFixture implements Agent
+{
+    use Promptable;
+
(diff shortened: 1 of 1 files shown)


pushpak1300:
Hey @mosabbirrakib, I’ve noticed that several of these recent PRs appear to be low-value and AI-generated, including the descriptions.

As per Laravel’s contribution policy, creating low-value or spam PRs that don’t add meaningful value may result in a block:
https://laravel.com/docs/13.x/contributions#ai-generated-contributions

I appreciate the effort you’re putting in, but opening multiple PRs in a short amount of time without clear value is not the right way to contribute. Please treat this as a first warning.

Going forward, please make sure each PR is meaningful, well thought out, and has a human-written description that clearly explains the change. I’ll be closing the PRs from this current iteration.

Feel free to send future changes in smaller batches, ideally 1 to 2 PRs at a time, once you’ve reviewed them carefully.

PR #651, opened 2026-05-22
Add filesystem tools for AI agents

pushpak1300:
Currently there's no built-in way to give an agent access to a Laravel filesystem disk. If you want an agent that can browse, read, or manage files on a `Storage` disk, you have to hand-roll a `Tool` for every operation.

This PR adds a set of `Storage`-backed tools and a `FileStorageTools` factory, so you can hand a set to an agent:

```php
use Laravel\Ai\Tools\FileStorageTools;

{
    return FileStorageTools::all('local');       // all 8 tools
    // return FileStorageTools::readOnly('local'); // just the 5 read-only tools
}
```

Both return a `Collection`, so you can narrow further:

```php
FileStorageTools::all('s3')->reject(fn ($t) => $t instanceof DeleteFile);
```

Every tool takes a disk name or `Filesystem` instance (default disk if omitted), so you can sandbox an agent to a folder with Laravel's `scoped` driver and it cannot escape it. Tools are `#[Strict]`, descriptions are written for tool-calling, and there are end to end tests (agent prompt to real disk side effect) using `Storage::fake()`.

### DIFF
--- a/src/Tools/FileStorage.php
+++ b/src/Tools/FileStorage.php
@@ -14,7 +14,7 @@ use Laravel\Ai\Tools\Filesystem\ListFiles;
 use Laravel\Ai\Tools\Filesystem\ReadFile;
 use Laravel\Ai\Tools\Filesystem\WriteFile;
 
-class FileStorageTools
+class FileStorage
 {
     /**
      * Create all of the file storage tools for the given disk.
--- a/src/Tools/FileStorageTools.php
+++ b/src/Tools/FileStorageTools.php
@@ -1,0 +1,48 @@
+<?php
+
+namespace Laravel\Ai\Tools;
+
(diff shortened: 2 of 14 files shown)


peterjthomson:
Makes a lot of sense to give agents (bounded) file system access. Also rounds out the previously confusing difference between giving agents:
Database Access (query builder, eloquent or schema)
Embedding Access (create vector embedding "file" from models, upload to vector "file" store, and search/access vector "file")
File Access (this PR)

danielhe4rt:
I kinda liked and it would replace a few agents which I build myself.

But I also would like to see some concrete usage examples in the docs. In the future, would be good to have another page in Laravel Docs with explicit tools and how to use it, like we have to Arr/Helpers etc.

taylorotwell:
Renamed `FileStorageTools` to `FileStorage`

PR #652, opened 2026-05-23
Extract recursive tool loop into TextGenerationLoop and convert OpenAI

vinitkadam03:
Currently every gateway reimplements the same recursive multi-step tool loop. Adding a provider means copying the orchestration (tool dispatch, max-step enforcement, step accumulation, final response assembly) on top of the actual request and response mapping.

This PR introduces `TextGenerationLoop` as a single orchestrator and converts OpenAI and Azure OpenAI into single-step gateways. A gateway now builds one request and parses one response (or one event stream); the loop owns everything else: tool execution, the step budget, accumulating steps, and assembling the final `TextResponse`.

Before, OpenAI's `ParsesTextResponses` had a `processResponse` that recursively called itself after executing tools and built the final response inline, and every other provider duplicated that shape.

After, a gateway implements `StepTextGateway`:

```php
public function handleStep(...): StepResponse
{
    $body = $this->buildStepBody(...);
    $response = $this->client($provider, $timeout)->post('responses', $body);

    return $this->parseTextResponse($response->json(), $provider, filled($schema));
}

public function streamStep(...): Generator
{
    // yields StreamEvents, returns the aggregate StepResponse
    return yield from $this->processTextStream(...);
}
```

`generateText`/`streamText` come from the `DelegatesToTextGenerationLoop` trait and just hand off to the loop.

A few notes:

- `StepTextGateway` is an opt-in sibling to the existing `TextGateway`, so the other nine providers keep working on their own loops while we migrate one at a time. The shared single-step plumbing lives in the `HandlesTextSteps` trait, used by both OpenAI and Azure.
- Streaming returns the per-step aggregate via the generator's return value (`$stream->getReturn()`) instead of yielding a sentinel, so the event stream stays purely `StreamEvent`s and `StreamEnd` stays generic (finish reason and usage, no provider leakage).
- Unknown tool calls now throw `NoSuchToolError` instead of being silently dropped, matching the AI SDK. BC-ish: a model calling an unregistered tool used to end the turn quietly, now it surfaces.


vinitkadam03:
Hey @pushpak1300, following up on #349. Per your suggestion about cherry-picking rather than merging everything in one go, I took a stab at a smaller foundation-only path: this PR ships StepLoop + the new contract with just OpenAI (and Azure, since they share traits) converted. Every other gateway is untouched.

Kept it in draft for now; no rush, just trying to help. [PR2](https://github.com/vinitkadam03/ai/pull/4) for the Anthropic migration is ready to stack on this once the approach is confirmed.

Honest open questions called out in the description (StreamEnd carrier, SingleTurnTextGateway as public contract, StepContext bag pattern). Happy to iterate on any of them.

pushpak1300:
- Renamed `StepLoop` to `TextGenerationLoop` so the name is clearer.
-  `generateText()` / `streamText()` still exist and now delegate through the loop.
- Fixed the streaming behavior so intermediate provider-turn `StreamEnd` events are not exposed before the tool loop finishes.
- Fixed max-step behavior so tools are not executed on the final allowed step when their results cannot be sent back to the provider.
- Added focused tests for both regressions.

gracjankubicki (commented):
Great direction — extracting the loop makes adding providers much cheaper, and the `TurnStreamEnd` sentinel keeps the public `StreamEnd` clean. I've been building on these internals recently (#700), so I read the loop closely. A few `generate()` vs `stream()` parity observations that might be worth addressing while the loop is still fresh:

1. **Continuation condition differs.** `generate()` continues only when `finishReason === FinishReason::ToolCalls && filled($toolCalls)`, while `stream()` continues whenever any `ToolCall` events were collected — `finishReason` from `TurnStreamEnd` isn't consulted. A turn that emits a tool call but finishes with a non-tool reason would stop in sync mode and continue in streaming mode.

2. **`AssistantMessage` content differs.** Sync uses the parsed `$lastResult->text`; streaming rebuilds it by concatenating `TextDelta`s. If a provider ever carries text outside `TextDelta`, the conversation history sent on the next turn differs between modes.

3. **History append asymmetry.** `generate()` appends the final `AssistantMessage` even when stopping; `stream()` appends only when continuing (the append sits below the `break`s). Harmless today since `stream()` doesn't use `$allMessages` after the loop, but a subtle trap for future persistence work.

4. **Missing `StreamEnd` edge case.** If `streamTurn()` completes without yielding a `TurnStreamEnd`, the loop returns silently and the consumer never receives a public `StreamEnd` (and no usage). Maybe worth an explicit error event or a contract guarantee.

A possible incremental fix for 1–3 without touching the contracts: normalize the turn outcome and move the continue-decision + message construction into one shared `advanceStep()` helper, leaving the two methods to own only transport (yield vs collect). Happy to send a PR against this branch if useful.

pushpak1300 (approved):

PR #655, opened 2026-05-25
Add MCP client and server tool support

### LINKED ISSUE #565: Support listing Remote MCPs as tools
This library only supports listing `function` tools. It would be greate to add support for `mcp` tools.
Here is how it is done for [OpenAi](https://developers.openai.com/api/docs/guides/tools-connectors-mcp#quickstart)
This is useful because in our case we offload vectorial operations to a python service that is available as an MCP.

pushpak1300:
Currently, if you're using the Laravel MCP package, you can't return MCP client tool primitives directly from an agent's `tools()` method. You have to manually wrap them yourself.

This PR lets you pass MCP client tools straight through:

```php
use Laravel\Mcp\Client;

public function tools(): array
{
    return [
        ...Client::web('https://mcp.example.com')
            ->withToken($token)
            ->tools(),

        new MyOwnTool,
    ];
}
```

For a named MCP client:

```php
use Laravel\Mcp\Facades\Mcp;

public function tools(): array
{
    return [
        ...Mcp::client('github')->tools(),
        new MyOwnTool,
    ];
}
```

For a local MCP server:

```php
use Laravel\Mcp\Client;

public function tools(): array
{
    return [
        ...Client::local('php', ['artisan', 'mcp:start'])->tools(),
    ];
}
```

It also works the other way: MCP **server** tool classes (the ones you'd register on a `Laravel\Mcp\Server` for an MCP host to call) can be reused as agent tools, so the same tool serves both the MCP server and any AI agent in your app:

```php
use App\Mcp\Tools\CurrentWeatherTool;

public function tools(): array
{
    return [
        new CurrentWeatherTool,
        new MyOwnTool,
    ];
}
```

The tool's `Laravel\Mcp\Request` is resolved through the container, so dependency injection, `$request->user()`, and `$request->validate()` keep working. Returned `Response`, `ResponseFactory`, arrays, and streamed generators are normalised down to the final non-notification response (notifications and intermediate updates are ignored).

The `GeneratesText` concern now checks `McpTool::supports()` and `McpServerTool::supports()` during tool resolution, so MCP primitives are wrapped automatically. `McpSchema` maps the MCP JSON schema shape to Laravel's JSON Schema types.

Tests included for both client and server tool flows, schema translation, structured/error responses, and streaming.


SergkeiM:
Tried this together with [#227](https://github.com/laravel/mcp/pull/227)  works like a charm thank you @pushpak1300

PR #663, opened 2026-05-27
Add cache directive to Bedrock system prompt via providerOptions

timko212:
Follow-up to #634, which was closed because it overlapped with how Anthropic already works.

Bedrock is the part that's actually missing. The Converse API doesn't accept `cache_control` on the system string; it wants an extra `cachePoint` block appended to the `system` list, which isn't something a generic passthrough can do. So there's no clean way to enable Bedrock prompt caching from an agent today.

This PR adds a `cache` directive that `BedrockTextGateway` recognises and turns into the `cachePoint` block. From the agent side it looks like this:

```php
public function providerOptions(Lab|string \$provider): array
{
    return \$provider === Lab::Bedrock
        ? ['cache' => [['target' => 'system']]]
        : [];
}
```

The directive is consumed by the gateway and stripped before the rest of `providerOptions` is merged into the Converse parameters, so AWS never sees the `cache` key.

The new test (\`tests/Feature/Providers/Bedrock/CacheControlTest.php\`) checks three things: the marker shows up when the directive is set, it doesn't leak through to AWS, and the system block stays untouched when no directive is set. There's also a small \`capturingBedrockClient\` helper in \`BedrockHelpers\` so the test can read the exact Converse payload.

### DIFF
--- a/src/Gateway/Bedrock/BedrockTextGateway.php
+++ b/src/Gateway/Bedrock/BedrockTextGateway.php
@@ -673,6 +673,16 @@ class BedrockTextGateway implements EmbeddingGateway, TextGateway
 
         $providerOptions = $options?->providerOptions(Lab::Bedrock);
 
+        if (! empty($providerOptions['cache'])) {
+            foreach ($providerOptions['cache'] as $entry) {
+                if (($entry['target'] ?? null) === 'system' && ! empty($parameters['system'])) {
+                    $parameters['system'][] = ['cachePoint' => ['type' => 'default']];
+                }
+            }
+
+            unset($providerOptions['cache']);
+        }
+
         if (! empty($providerOptions)) {
             $parameters = array_merge($parameters, $providerOptions);
         }
--- a/tests/Feature/Providers/Bedrock/BedrockHelpers.php
+++ b/tests/Feature/Providers/Bedrock/BedrockHelpers.php
@@ -3,8 +3,10 @@
 namespace Tests\Feature\Providers\Bedrock;
 
 use Aws\BedrockRuntime\BedrockRuntimeClient;
+use Aws\CommandInterface;
 use Aws\MockHandler;
 use Aws\Result;
+use GuzzleHttp\Promise\FulfilledPromise;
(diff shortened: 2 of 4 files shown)


pushpak1300:
AI-REVIEW: _Automated review — Claude Opus 4.8 running `/code-review` (high) plus a multi-axis pass. Bedrock suite (incl. `CacheControlTest`) run and passing on the branch._

The core change is correct: the `cachePoint` block shape `['cachePoint' => ['type' => 'default']]` matches the Bedrock Converse `SystemContentBlock` spec, the directive is stripped before merge in the normal case, and caching correctly applies to both sync and streaming paths (they share `buildConverseParameters`).

**Should-fix**

- `src/Gateway/Bedrock/BedrockTextGateway.php:676-684` — an empty/falsy `cache` value leaks to AWS. The `unset($providerOptions['cache'])` is nested *inside* the `if (! empty($providerOptions['cache']))` guard. If an agent returns a present-but-falsy value (e.g. `['cache' => []]` from a conditional like `$shouldCache ? [[...]] : []`), the guard is false, `unset` never runs, and the subsequent `array_merge` (L686-687) passes `cache => []` straight into the Converse request — breaking the PR's stated guarantee that AWS never sees the `cache` key, and risking a request-validation error. Fix: unset unconditionally when the key exists (move `unset` out of the guard, or `if (array_key_exists('cache', $providerOptions))`). Verified by simulation: falsy values leak; a valid non-empty directive does not.

**Nits**

- `BedrockTextGateway.php:677-681` — unrecognized `target` values (e.g. `'messages'`, `'tools'`) are silently dropped (no cachePoint, no signal). Matches the system-only scope and doesn't leak, but a typo'd target becomes a silent no-op. Worth documenting that only `target: 'system'` is supported.
- `BedrockTextGateway.php:677-681` — duplicate `['target' => 'system']` entries append multiple `cachePoint` markers (no dedupe in the loop). Low impact, easy to guard.
- All three tests exercise only the sync path; one streaming assertion would lock in that streaming also emits the cachePoint. Optional.

pushpak1300:
Closing becuase of #869

PR #671, opened 2026-05-29
Fix Mistral transcription requests

### LINKED ISSUE #60319: HTTP requests payload regression for multipart arrays
### Laravel Version

13.12.0

### PHP Version

8.5.0

### Database Driver & Version

_No response_

### Description

The `v13.9.0` release included a fix that changes the serialization of array fields in multipart forms (See PR #59984). As far as I understand, this was intended as a minor fix that shouldn't cause any regressions. But it did break some things, for example in the `laravel/ai` package.

I haven't looked into this too much, but it seems like some parsers interpret multipart forms differently, and the change in serialization breaks at least some Python parsers.

Here's an AI-generated summary of the issues I've ran into debugging with Gemini:

#### The Serialization Discrepancy

When passing an array value such as  ['timestamp_granularities' => ['segment']]  in a multipart request:

Before  v13.9.0: Laravel serialized the array by repeating the exact key name:

```
Content-Disposition: form-data; name="timestamp_granularities"

segment
```

After  v13.9.0: Guzzle delegates serialization to PSR-7, which appends PHP-style brackets  []  to the key name:

```
Content-Disposition: form-data; name="timestamp_granularities[]"

segment
```

#### Why this breaks Python / FastAPI / Pydantic APIs

Unlike PHP, many API backends built in Python (using FastAPI, Pydantic, or  python-multipart ) do not interpret trailing  []  as arrays automatically.

Instead, the parser:

  1. Looks strictly for the exact parameter name ( timestamp_granularities ).
  2. Treats  timestamp_granularities[]  as an unrelated, unknown key and discards it.
  3. Sees the parameter  timestamp_granularities  as missing/empty, resulting in  422 Unprocessable Entity  validation errors (e.g.  timestamp granularity must be set to ['segment'], got [] ).

This regression breaks out-of-the-box integrations with third-party APIs served by Python backends (such as Mistral AI's Voxtral STT).

### Steps To Reproduce

I reproduced this error using `laravel/ai`. You can see more details in this PR: laravel/ai/pull/671

NoelDeMartin:
I recently upgraded my laravel/framework version, and I started getting errors from Mistral's transcription API. In particular, I got the following:

```json
{
  "object": "error",
  "message": {
    "detail": [
      {
        "type": "assertion_error",
        "loc": [],
        "msg": "Assertion failed, When diarize is set to True and streaming is disabled, the timestamp granularity must be set to ['segment'], got []",
        "input": {
          "model": "voxtral-mini-latest",
          "audio": "--audio-bytes--",
          "language": null,
          "temperature": null,
          "max_tokens": null,
          "stream": false,
          "return_language": false,
          "streaming": "disabled",
          "timestamp_granularities": [],
          "diarize": true,
          "context_bias": [],
          "target_streaming_delay_ms": null,
          "metadata": null
        },
        "ctx": {
          "error": {}
        }
      }
    ]
  },
  "type": "invalid_request_error",
  "param": null,
  "code": null,
  "raw_status_code": 422
}
```

Looking at the code, I was confused because the `timestamp_granularities` seems to be set to "segments" in the code... but after some digging, I could trace back the regression to a recent PR in laravel/framework: laravel/framework#59984

This was released in `v13.9.0`, and I'm not sure how many things are broken because of that. At the very least, this PR fixes Mistral's transcriptions.

I have reported this issue to laravel/framework in case you want to take a look at a more general fix: https://github.com/laravel/framework/issues/60319

### DIFF
--- a/src/Gateway/Mistral/MistralGateway.php
+++ b/src/Gateway/Mistral/MistralGateway.php
@@ -178,10 +178,7 @@ class MistralGateway implements EmbeddingGateway, TextGateway, TranscriptionGate
 
         if ($diarize) {
             $params['diarize'] = true;
-            $params[] = [
-                'name' => 'timestamp_granularities',
-                'contents' => 'segment',
-            ];
+            $params['timestamp_granularities'] = ['segment'];
         } elseif ($language) {
             $params['language'] = $language;
         }
@@ -190,7 +187,7 @@ class MistralGateway implements EmbeddingGateway, TextGateway, TranscriptionGate
             $provider->name(),
             fn () => $this->client($provider, $timeout)
                 ->attach('file', $audio->content(), $this->audioFilename($audio), ['Content-Type' => $audio->mimeType()])
-                ->post('audio/transcriptions', array_merge($providerOptions, $params)),
+                ->post('audio/transcriptions', $this->multipartParams(array_merge($providerOptions, $params))),
         );
 
         $data = $response->json();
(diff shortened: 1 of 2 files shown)


pushpak1300:
Thanks.

PR #672, opened 2026-05-29
feat(bedrock): opt-in Converse prompt caching via cachePoints provider option

yamenarahman:
## What

Adds opt-in [Bedrock Converse prompt caching](https://docs.aws.amazon.com/bedrock/latest/userguide/prompt-caching.html) to `BedrockTextGateway` via a reserved `cachePoints` provider option.

## Why

The gateway already reads `cacheReadInputTokens` / `cacheWriteInputTokens` back from the Converse response, but it never emits the `{cachePoint: {type: 'default'}}` markers the API requires — so those counters are always `0` and a stable system prompt + tool set are re-billed at the full input rate on every call. There was no way to request caching: `providerOptions` flat-merges at the top level of the request and can't reach the nested `system` / `toolConfig` blocks where the marker has to live.

## How

Agents name where to cache via the Bedrock provider option:

```php
{
    return match ($provider) {
        Lab::Bedrock => ['cachePoints' => ['system', 'tools']],
        default => [],
    };
}
```

`buildConverseParameters` pulls the directive out **before** the existing flat-merge (so it never leaks into the request body) and appends a cachePoint block to the `system` array and/or `toolConfig.tools`. Locations with no content (no instructions / no tools) are skipped.

Scoped to the static prefix — `system` + `tools` — which is the biggest, most stable cache target. Message-prefix caching can be a follow-up.

## Tests

- `applyCachePoints` unit tests: system-only, tools-only, both, no-op when unrequested or target absent.
- `buildConverseParameters` integration test: directive drives insertion and does **not** flat-merge into the body.
- Full `BedrockTextGatewayTest` suite green (45 passed); Pint + PHPStan clean on the touched files.

### DIFF
--- a/src/Gateway/Bedrock/BedrockTextGateway.php
+++ b/src/Gateway/Bedrock/BedrockTextGateway.php
@@ -655,7 +655,43 @@ class BedrockTextGateway implements EmbeddingGateway, StepTextGateway
         $providerOptions = $options?->providerOptions(Lab::Bedrock);
 
         if (! empty($providerOptions)) {
+            $parameters = $this->applyCachePoints($parameters, Arr::pull($providerOptions, 'cachePoints'));
+
             return array_merge($parameters, $providerOptions);
+        }
+
+        return $parameters;
+    }
+
+    /**
+     * Insert Bedrock Converse prompt-caching markers at the requested locations.
+     *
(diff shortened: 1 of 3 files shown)


pushpak1300:
Can you please fix conflicts ?

yamenarahman:
@pushpak1300 Resolved.

pushpak1300:
Closing because of #869

PR #679, opened 2026-06-02
Filter orphan tool calls when rehydrating database conversations

### LINKED ISSUE #639: tool_use ids were found without tool_result blocks immediately after
We occasionally get an error `tool_use ids were found without tool_result blocks immediately after`, the [docs](https://platform.claude.com/docs/en/agents-and-tools/tool-use/troubleshooting-tool-use#errors-at-request-time) list the cause and fix here. This is an agent, which has structured output and 16 tools.

FYI, the below was found by Claude, but posting in case it might be useful.

> Root cause: unexecuted tool calls included in stored response when maxSteps is exceeded
> 
> When processResponse() determines $shouldContinue = false due to hitting the maxSteps limit mid-turn, it skips executeToolCalls() — so $toolResults stays empty for that final turn. However, the StructuredTextResponse is still built by flatmapping tool calls and results across all steps, meaning the unexecuted tool calls from the final turn are included in toolCalls with no corresponding entries in toolResults.
> 
> Since there is no output_structured_data block in the response (the model never got to return its final answer), extractStructuredOutput() returns [], so the stored content is "[]" — the user receives a silent empty response with no indication the step limit was reached.
> 
> This manifests as two sequential symptoms:
> 
> 1. First: the user receives an empty [] response with no explanation
> 2. Next message: replaying the corrupt history to the API fails with "tool_use ids were found without tool_result blocks", because the AssistantMessage references a tool use ID that has no corresponding tool_result block — leaving the conversation permanently broken
> 
> Two fixes needed:
> 1. When maxSteps is exceeded with pending $realToolCalls, those unexecuted calls should be excluded from the final response's toolCalls collection. Ideally the limit should also be surfaced to the caller rather than silently swallowed.
> 2. DatabaseConversationStore::getLatestConversationMessages() should defensively filter out tool calls with no matching result before replaying history, to prevent permanently broken conversations from this corrupt state.

Anoop-Kadachi:
Fixes #639.

`DatabaseConversationStore::getLatestConversationMessages()` currently rehydrates every stored tool call on an assistant row, even if that tool call has no matching tool result.

That can happen when a multi-step agent turn hits its `maxSteps` limit after the model has requested tools, but before those tool calls are executed. The assistant message can then be stored with pending `tool_calls` and no corresponding `tool_results`.

On the next prompt, the stored conversation history is replayed with orphan tool-use IDs. Providers such as Anthropic reject that history with:

```text
tool_use ids were found without tool_result blocks immediately after


Changes
------------------
Filters stored assistant tool_calls during database conversation rehydration so only calls with a matching tool_results.id are replayed.
Preserves assistant text content when all stored tool calls are orphaned.
Skips fully empty assistant rows that only contain orphaned tool calls.
Adds regression coverage for mixed matched/orphan calls, orphan calls with assistant content, and orphan calls without content.

Why
--------------

This prevents one interrupted or max-step-limited agent turn from permanently corrupting future conversation replays.

### DIFF
--- a/src/Storage/DatabaseConversationStore.php
+++ b/src/Storage/DatabaseConversationStore.php
@@ -156,19 +156,30 @@ class DatabaseConversationStore implements ConversationStore
                 }
 
                 if ($toolCalls->isNotEmpty()) {
+                    $resultIds = $toolResults->pluck('id')->filter()->all();
+
+                    $matchedToolCalls = $toolCalls->filter(
+                        fn ($toolCall) => isset($toolCall['id'])
+                            && in_array($toolCall['id'], $resultIds, true)
+                    )->values();
+
                     $messages = [];
 
-                    $messages[] = new AssistantMessage(
-                        $record->content ?: '',
-                        $toolCalls->map(fn ($toolCall) => new ToolCall(
-                            id: $toolCall['id'],
-                            name: $toolCall['name'],
-                            arguments: $toolCall['arguments'],
-                            resultId: $toolCall['result_id'] ?? null,
-                            reasoningId: $toolCall['reasoning_id'] ?? null,
-                            reasoningSummary: $toolCall['reasoning_summary'] ?? null,
-                        ))
-                    );
+                    if ($matchedToolCalls->isNotEmpty()) {
+                        $messages[] = new AssistantMessage(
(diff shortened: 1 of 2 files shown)


pushpak1300:
can you fix conflicts ?

PR #681, opened 2026-06-02
Fix tool-calling model override causing model_not_found errors.

### LINKED ISSUE #680: [Bug] generateText tool-calling uses response model instead of original model
### Bug Description

When using `generateText` with tools, the model returned in the API response (`$response['model']`) is used for the next tool invocation instead of the original model passed to the function. This causes `model_not_found` errors when the response model differs from the provider alias.

### Steps to Reproduce

1. Call `generateText` with a model alias and tools.
2. The response includes a different model name.
3. Tool invocation fails with `model_not_found`.

### Proposed Fix

Pass the original `$model` to `parseTextResponse` and all subsequent processing instead of using `$response['model']`.

```php
// Before
$this->parseTextResponse($data, $provider, ..., $response['model']);

// After
$this->parseTextResponse($data, $provider, ..., $model);

pedram-davoodi:
Fixes https://github.com/laravel/ai/issues/680.

###  Summary

This PR fixes an issue in `generateText` tool-calling flow where the model returned from the API response was incorrectly used for subsequent tool execution steps instead of the original model provided by the developer.

This could lead to `model_not_found` errors when the provider returns a canonical or internal model name different from the configured alias.

---

###  Problem

During tool-calling flows, the following happens:

1. Developer passes a model alias (e.g. `gapgpt-qwen-3.5`)
2. API returns a response with a different internal model name (e.g. `Qwen/Qwen3.6-35B-A3B-FP8`)
3. The SDK incorrectly propagates the response model into subsequent tool execution steps
4. The next request fails with:

---

### Root Cause

The original `$model` parameter is not consistently propagated through the response lifecycle.

Instead, downstream methods (`parseTextResponse`, `processResponse`, and tool invocation flow) may rely on:

---

This PR ensures that the original requested model is preserved and used throughout the entire request lifecycle, including:

parseTextResponse
processResponse
tool-calling continuation steps

### DIFF
--- a/src/Gateway/OpenAi/Concerns/ParsesTextResponses.php
+++ b/src/Gateway/OpenAi/Concerns/ParsesTextResponses.php
@@ -60,6 +60,7 @@ trait ParsesTextResponses
         ?array $schema = null,
         ?TextGenerationOptions $options = null,
         ?int $timeout = null,
+        string $model
     ): TextResponse {
         return $this->processResponse(
             $data,
@@ -67,6 +68,7 @@ trait ParsesTextResponses
             $structured,
             $tools,
             $schema,
+            $model,
             new Collection,
             new Collection,
             maxSteps: $options?->maxSteps,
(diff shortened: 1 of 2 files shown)


pushpak1300:
Thanks for your pull request!

PR #682, opened 2026-06-03
Recursively remove additionalProperties for Gemini tools

Edward144:
Similar to the issue highlighted in #139. Tools containing nested schema objects used by Gemini throw an error for `additionalProperties`.

```
"code": 400,
"message": "Invalid JSON payload received. Unknown name \"additionalProperties\"" 
```

For example in a tool such as:
```
class ExampleTool implements Tool
{
    public function description(): Stringable|string
    {
        return '';
    }

    {
        return '';
    } 

    public function schema(JsonSchema): array
    {
        return [
            'array' => $schema->array()
                ->items($schema->object([
                    'name' => $schema->string()->required(),
                    'description' => $schema->string()->required(),
                ]))
                ->required(),
        ];
    }
}
```

To fix this I have updated the `MapsTools` concern to recursively remove the key from the schema.

### DIFF
--- a/src/Gateway/Gemini/Concerns/MapsTools.php
+++ b/src/Gateway/Gemini/Concerns/MapsTools.php
@@ -67,14 +67,13 @@ trait MapsTools
         if (filled($schema)) {
             $schemaArray = (new ObjectSchema($schema))->toSchema();
 
-            $definition['parameters'] = Arr::except(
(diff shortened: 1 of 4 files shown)


jrking4:
this fixes an issue i am having as well with Gemini tools

Issue #636, opened 2026-05-20
Feature: Expose the underlying model API response headers


cedrictwillie:
I couldn't find where the underlying response was accessible. My use case is being able to inspect the response headers for rate limit quota information. 

- x-ratelimit-limit-requests
- x-ratelimit-limit-tokens
- x-ratelimit-remaining-requests
- x-ratelimit-remaining-tokens
- etc.

OpenAI: https://developers.openai.com/api/docs/guides/rate-limits#rate-limits-in-headers

dumbbellcode:
is it something the team is interested in? then i can work on it. 

ping: @pushpak1300

pushpak1300:
How do you think an API looks for this? I’m thinking of providing raw resposne object on the result.

cedrictwillie:
@pushpak1300 @dumbbellcode The API could be similar to how "steps" is constructed in the response. Where there is a steps array containing all steps, and the top level response structure contains the final step. This covers single/multi turn calls.

Or, there could be an option to hook into or configure the HTTP client that will be used entirely? Just throwing out initial ideas. Honestly haven't looked into the code deeply to understand the implications of these ideas.

```php
use Illuminate\Support\Facades\Http;

class FooAgent implements Agent
{
    protected function withHttpClient()
    {
        return Http::baseUrl('')
            ->withHeaders(['...'])
            ->throw();
    }
}
```

dumbbellcode:
@pushpak1300 @cedrictwillie 

**Option 1**: we can expose the last http call response headers in the `TextResponse` class for example 

```php
$response = $agent->prompt('Summarise this document.')
$headers = $response->headers // accessible from the last api call
```

**Option 2**: we can expose the whole response object from the last http call, as it's available

```php
$response = $agent->prompt('Summarise this document.')
$httpResponse = $response->httpResponse // accessible from the last api call
```

personally i'd suggest we expose the whole object (option 2), without any kind of opt-in from the users, as it's readily available after api calls

pushpak1300:
`$response->raw` make more sense to me ? Are we always gonna store the latest response ?

cedrictwillie:
I like `$response->raw`! Having "hooks" into every response is the main concern I was thinking of when suggesting the ability to configure the HTTP client that will be used. This give the ability to configure middleware.

This definitely complicates the implementation, so maybe just the latest response is suitable for now? ($response->raw)

dumbbellcode:
[...]

PR #686, opened 2026-06-03
Add search query extraction to Gemini and OpenAI response processing

cutlope:
- Implemented `extractSearchQueries` method in both `ParsesTextResponses` classes for Gemini and OpenAI to gather web search queries from response data.
- Updated the `Meta` class to include `searchQueries` property.
- Modified response processing to include extracted search queries in the `Meta` object.
- Enhanced tests to verify the correct extraction of search queries from provider tool calls and responses.

Relevant Docs:
[OpenAI](developers.openai.com/api/reference/resources/responses)
[Google](https://ai.google.dev/gemini-api/docs/google-search#understanding_the_grounding_response)

### DIFF
--- a/src/Gateway/Gemini/Concerns/ParsesTextResponses.php
+++ b/src/Gateway/Gemini/Concerns/ParsesTextResponses.php
@@ -95,11 +95,12 @@ trait ParsesTextResponses
         $text = $this->extractText($parts);
         $rawToolCalls = $this->extractRawToolCalls($parts);
         $citations = $this->extractCitations($data);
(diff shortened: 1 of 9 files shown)


pushpak1300:
Test failing.

cutlope:
> Test failing.

Hmm i had a look but these tests are not related to the changes here.

pushpak1300:
Thanks for this! I don't think we should add `searchQueries` as a top-level `Meta` property though. It means plumbing each new field across both providers + streaming, which won't really scale.

Raw response / headers is a separate thing I'm tracking in #636.

Gonna close this for now.

PR #690, opened 2026-06-05
Respect morph map aliases when storing the agent on conversation messages

jeffreyvanhees:
While integrating conversations in my app I noticed the `agent` column on `agent_conversation_messages` always ends up with the full class name (`App\Ai\Agents\ConsumerAgent` in my case), even though my app registers a morph map. The store writes `$prompt->agent::class` directly, so the morph map never applies.

That means renaming or moving an agent class leaves stale class names behind, and any query I do on that column (listing conversations per agent, some basic stats) silently misses older rows. Same reason `_type` columns go through the morph map.

This changes both write sites in `DatabaseConversationStore` to use `Relation::getMorphAlias()` instead. When no alias is registered it just returns the class name, so apps without a morph map are unaffected. The package itself never reads the column back, so existing rows with FQCNs keep working too.

Added tests for both the mapped and unmapped case.

### DIFF
--- a/src/Storage/DatabaseConversationStore.php
+++ b/src/Storage/DatabaseConversationStore.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Ai\Storage;
 
+use Illuminate\Database\Eloquent\Relations\Relation;
 use Illuminate\Database\Query\Builder;
 use Illuminate\Support\Collection;
 use Illuminate\Support\Facades\DB;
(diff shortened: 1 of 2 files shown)


pushpak1300:
Thanks for the PR and the clear write-up.

I've thought about this and I'm going to pass on it, for a few reasons:

The package never reads the `agent` column back — it's written purely as informational metadata for an application's own queries and stats. Because nothing in the library consumes it, storing the morph alias vs. the FQCN doesn't fix any correctness issue on our end.

That makes this the app's data to shape for the app's purposes. If you want morph aliases in that column for your queries, you can normalize it in your own query layer or with a one-line backfill — and you'd need a backfill regardless, since this change doesn't migrate existing rows.

That's also where the real cost is: for an app that already has a morph map registered and existing conversation data, this would silently change the stored value mid-table (old rows keep the FQCN, new rows get the alias), splitting any historical query on that column. That's a behavior change I don't want to ship.

So I'd rather leave the column as the literal class name and let applications map it however suits them. Appreciate the effort here regardless.

PR #696, opened 2026-06-09
Normalize MCP tool schemas into the deserializable JSON Schema subset

pushpak1300:
MCP tools from real-world servers regularly use JSON Schema features that `Illuminate\JsonSchema` can't deserialize — `$ref`, `allOf`, `anyOf`/`oneOf`, multi-type unions, `const`, unsupported keywords like `not` and `patternProperties`. Currently those tools either throw or are silently skipped.

This PR introduces `SchemaNormalizer`, which rewrites an arbitrary schema into the subset the deserializer accepts:

- inlines local `$ref` pointers (cycle-safe)
- flattens `allOf` branches into the parent node
- collapses `anyOf`/`oneOf` into scalar multi-type unions or nullable wrappers
- strips unsupported keywords
- rewrites `const` to a single-value `enum`
- infers a type from node shape when none is present

`McpTool` now runs every input schema through the normalizer before handing it to `JsonSchemaFactory`, and wraps the factory call in a try/catch so a still-unrepresentable schema returns an empty properties list rather than aborting the tool entirely.

Also bumps `laravel/mcp` to `^0.8` and adapts the fixtures to extend the real `Tool` / `ToolResult` classes rather than hand-rolling fakes.

### DIFF
--- a/src/Providers/Concerns/GeneratesText.php
+++ b/src/Providers/Concerns/GeneratesText.php
@@ -27,7 +27,6 @@ use Laravel\Ai\Responses\StructuredAgentResponse;
 use Laravel\Ai\Tools\AgentTool;
 use Laravel\Ai\Tools\McpServerTool;
 use Laravel\Ai\Tools\McpTool;
-use Throwable;
 
 use function Laravel\Ai\pipeline;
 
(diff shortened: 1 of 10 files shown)


src/Providers/Concerns/GeneratesText.php
```diff
@@ -120,10 +121,25 @@ protected function resolveTools(Agent $agent): array
...
+    /**
+     * Determine whether the given MCP client tool can be prepared for the provider.
+     */
+    protected function mcpToolIsUsable(McpTool $tool): bool
+    {
+        try {
+            $tool->schema(new JsonSchemaTypeFactory);
+
+            return true;
+        } catch (Throwable) {
```
WendellAdriel: Is intentional to catch every `Throwable` here?
I'm asking just because this can hide/mask other issues.
But if that's intentional, it's all good

tests/Feature/McpToolTest.php
```diff
@@ -67,3 +62,106 @@ public function tools(): iterable
         ]
     );
 });
+
+test('it skips mcp client tools whose schema cannot be represented', function () {
```
WendellAdriel: This covers the normal prompt path, but `StreamsText` also goes through `resolveTools()`, so this behavior can still regress for streamed agent responses without a test catching it. We could add a focused streaming regression test that asserts that an unrepresentable MCP client tool is skipped while the stream still proceeds.

WendellAdriel (changes requested):

WendellAdriel (approved):
LGTM

PR #697, opened 2026-06-10
Add hosted tool search via ToolSearch tool and stateless output replay (OpenAI + Anthropic)

behzadsp:
Currently every tool an agent exposes is shipped to the provider on every request. With a large catalog that bloats the prompt and gives the model a long menu to choose from, which costs tokens and hurts tool-selection accuracy.

This PR lets you defer individual tools so OpenAI and Anthropic load them on demand through their hosted tool search, instead of receiving every definition up front. Deferred tools are wrapped in a `ToolSearch` provider tool, and the gateway emits the provider's hosted search entry plus `defer_loading: true` definitions for everything inside the wrapper.

**Before**, everything ships every request:

```php
{
    return [new WeatherTool, new SearchInvoices, new RefundOrder];
}
```

**After**, wrap the tools you want discovered on demand:

```php
{
    return [
        new WeatherTool,
        new ToolSearch(tools: [new SearchInvoices, new RefundOrder]),
    ];
}
```

The tools themselves need no changes — no interface, no provider options. The wrapper maps to `{"type":"tool_search"}` on OpenAI and `tool_search_tool_{regex,bm25}_20251119` on Anthropic, mirroring how the providers model it natively (a search tool plus per-tool deferral).

Anthropic's search strategy is a first-class parameter, validated at construction (`regex`|`bm25`):

```php
new ToolSearch(tools: [new SearchInvoices], strategy: 'bm25')
```

Other search-entry API fields (e.g. `cache_control`) pass through `withProviderOptions()` on the wrapper. When a deferred tool is called, the gateway resolves it by name from inside the wrapper exactly like a top-level tool, and step budgeting expands each wrapper into its deferred tool count.

Notes:

- Hosted tool search is gated to providers that support it (OpenAI, Anthropic) via the `SupportsToolSearch` marker. Every other provider fails fast before a request goes out: Azure and the Chat Completions gateways (Groq, DeepSeek, Mistral, OpenRouter), plus Gemini, xAI, Bedrock, and Ollama all throw a clear exception instead of silently dropping the wrapper's tools. Model support is enforced by the provider API, not locally.
- The support guard runs even when the wrapper is empty, so a misconfigured provider surfaces in development rather than only once the deferred tool list is non-empty in production. An empty wrapper on a supported provider emits nothing.
- Anthropic returns 400 if every tool is deferred, so the gateway requires at least one non-deferred tool alongside a non-empty wrapper.
- OpenAI hosted search requires stored responses, so a `ToolSearch` with `store=false` throws instead of silently failing.
- A `strategy` mutated past the constructor to an unknown value falls back to `regex` at mapping time instead of producing an invalid Anthropic tool type.

This also includes the stateless (`store=false`) output replay for OpenAI: the gateway captures the full ordered response output (reasoning, hosted-tool, and function-call blocks) and replays it verbatim on continuation, preserving item order. It touches the core OpenAI message-replay path, so it is worth reviewing independently of the tool-search change.

Client-side tool search (the library ranking a catalog itself, portable across every provider) is a separate concern and ships in #805; since this PR claims the `ToolSearch` name for the hosted wrapper, the client-side implementation will need a distinct name.

Tests updated accordingly. Full suite, Pint, and PHPStan clean.


pushpak1300:
## Reworked the API around a `ToolSearch` provider tool

Instead of marking each tool with `#[Deferred]` and auto-enabling search, deferral is now expressed by wrapping the deferred tools in a `ToolSearch` provider tool, consistent with the `WebSearch` / `WebFetch` / `FileSearch` family.

Before:

```php
#[Deferred]
class ArchiveCustomer implements Tool { /* ... */ }

{
    return [new GetWeather, new ArchiveCustomer]; // search auto-enabled
}
```

After:

```php
{
    return [
        new GetWeather,                               // eager
        new ToolSearch(tools: [new ArchiveCustomer]), // deferred, discovered via search
    ];
}
```

### Why the change

- **Consistency.** Tool search is a server tool, so it should be a `ProviderTool` you configure at the call site like the others, not an attribute plus implicit activation.
- **Explicit grouping.** The deferred set lives in one place instead of being scattered as attributes across N tool classes.
- **Per-agent flexibility.** The same `Tool` can be deferred in one agent and eager in another. The attribute baked that decision into the class globally.
- **Vendor-neutral strategy.** Anthropic's `regex`/`bm25` is provider-specific, so it goes through `withProviderOptions(Lab::Anthropic, ['strategy' => 'bm25'])`, following the convention from #662. Default is `regex`; OpenAI ignores it.

Verified end-to-end against the live OpenAI API (the model searches, discovers the deferred tool, and calls it).

taylorotwell:
My agent's review.

**Blocking**

- src/Gateway/OpenAi/Concerns/ParsesTextResponses.php:114 - OpenAI stateless tool loops drop hosted tool_search_call / tool_search_output items because only function calls are stored in AssistantMessage, so store=false follow-up requests cannot replay the full response output required for deferred-tool discovery.

**Should fix**

- src/Gateway/OpenAi/Concerns/MapsTools.php:30 - AzureOpenAiGateway uses this shared trait, so ToolSearch emits OpenAI tool_search / defer_loading payloads for Azure even though the PR only documents OpenAI and Anthropic support and has no Azure support contract or tests.
- src/Gateway/OpenAi/Concerns/BuildsTextRequests.php:30 - OpenAI mapping has no model guard, so ToolSearch on models older than gpt-5.4 sends an API-invalid tool_search request instead of rejecting or degrading as described.

**Consider**

- tests/Feature/Providers/OpenAi/ToolSearchTest.php:17 - Add coverage for store=false and unsupported/Azure models so the deferred-tool loop and support boundaries are exercised.

behzadsp:
**Edge case in the `store=false` tool_search replay: reasoning gets reordered after the search**

Following up on the stateless-replay work in `005e278` — there's an ordering issue in the deferred-tool replay for reasoning models (which is the only supported config, since gating requires `gpt-5.4+` and those run with `reasoning.encrypted_content` under `store=false`).

**Repro** — a `store=false` turn whose output is `reasoning → tool_search_call → tool_search_output → function_call`:

```php
'output' => [
    ['type' => 'reasoning', 'id' => 'rs_1', 'summary' => [], 'encrypted_content' => 'enc-1'],
    ['type' => 'tool_search_call', 'id' => 'ts_1', 'status' => 'completed', 'queries' => ['secret']],
    ['type' => 'tool_search_output', 'id' => 'tso_1', 'tools' => ['DeferredTool']],
    ['type' => 'function_call', 'id' => 'fc_1', 'call_id' => 'call_1', 'name' => 'DeferredTool', 'arguments' => '{}'],
],
```

The follow-up request replays `input` as:

```
system, user, tool_search_call, tool_search_output, reasoning, function_call, function_call_output
```

i.e. the `reasoning` item that originally **preceded** the search is moved **after** it. OpenAI's stateless continuation expects the prior output replayed in order, so the reasoning↔action sequence no longer matches what the model produced.

**Root cause**

- `ParsesTextResponses::extractReplayBlocks()` excludes `reasoning` (only captures `tool_search_call`/`tool_search_output`), so reasoning isn't part of the captured replay blocks.
- `MapsMessages::mapAssistantMessage()` prepends those blocks first, then reconstructs `reasoning` + `function_call` from the tool calls afterward — pushing the search items ahead of the reasoning.
- Related latent case: a `reasoning` item that precedes a search but isn't followed by a `function_call` is captured by neither path (`extractReplayBlocks` skips reasoning; `mapToolCallsWithReasoning` only keeps reasoning attached to a function call), so it's dropped from the replay entirely.

The existing `store=false` test passes because its fixture has no `reasoning` item, so the ordering isn't exercised.

**Suggested fix** (keeps the current architecture + the reject-on-unsupported behavior): capture `reasoning` together with the `tool_search` items in their original order, and emit that ordered prefix before the reconstructed function calls (rather than prepending only the search blocks). Happy to push a commit with that + a test that locks the `reasoning → search_call → search_output → function_call` order if it's useful.

pushpak1300:
All issues resolved.

PR #700, opened 2026-06-10
Add native sub-agent activity streaming

gracjankubicki:
## Goal

This PR adds the first step toward native streaming of sub-agent activity when sub-agents are invoked as tools.

The scope is intentionally narrow: the core contract, opt-in behavior for sub-agents, provenance for nested tool calls, and mapping nested events to the Vercel data protocol.

This PR is a foundation for follow-up work around a public consumer API for `data-subagent`, documentation, and examples for rendering multiple parallel and nested sub-agents on the client side.

## What changes

- Adds the `StreamableTool` contract, allowing a tool to emit events while it runs and still return a final tool result.
- Adds the `StreamsActivity` marker interface, allowing an agent to opt in to streamed activity when it is used as a sub-agent.
- Keeps `AgentTool` synchronous by default, but for agents implementing `StreamsActivity` it uses `agent->stream(...)` and forwards streamed events.
- Adds provenance to nested sub-agent events:
  - `parent_invocation_id`
  - `parent_tool_call_id`
  - `ancestor_tool_call_ids`
  - `depth`
- Supports parallel and nested sub-agents through the full tool-call path, for example `["call_001", "call_002"]`.
- Ensures `StreamableAgentResponse` does not mix text or usage from nested events into the top-level response.
- Emits nested activity in the Vercel data protocol as `data-subagent` instead of pretending those events are top-level text or tool parts.
- Adds fake gateway support for streaming tool calls so sub-agent streaming can be tested without a real provider.
- Wires streaming tool execution into the streaming paths for:
  - OpenAI
  - Azure OpenAI
  - OpenRouter
  - Anthropic
  - Gemini
  - Groq
  - DeepSeek
  - Mistral
  - Ollama
  - xAI
  - Bedrock

## What this PR does not do

- It does not change the default behavior of regular sub-agents that do not implement `StreamsActivity`.
- It does not change the synchronous `prompt()` / non-streaming tool execution path.
- It does not define the final client-side API or UI for consuming `data-subagent`.

## Tests

Run locally:

- `composer test -- tests/Feature/Providers/OpenAi/StreamingTest.php tests/Feature/Providers/AzureOpenAi/StreamingTest.php tests/Feature/Providers/OpenRouter/StreamingTest.php tests/Feature/Providers/Groq/StreamingTest.php tests/Feature/Providers/DeepSeek/StreamingTest.php tests/Feature/Providers/Mistral/StreamingTest.php tests/Feature/Providers/Gemini/StreamingTest.php tests/Feature/Providers/Anthropic/StreamingTest.php tests/Feature/Providers/Ollama/StreamingTest.php tests/Feature/Providers/Xai/StreamingTest.php tests/Feature/Providers/Bedrock/StreamingTest.php`
- `composer lint`
- `composer test`

Full test suite result: `1499 passed`, `193 skipped`.

The skipped tests require external provider API keys.

### DIFF
--- a/src/Contracts/StreamableTool.php
+++ b/src/Contracts/StreamableTool.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace Laravel\Ai\Contracts;
+
+use Generator;
+use Laravel\Ai\Streaming\Events\StreamEvent;
+use Laravel\Ai\Tools\Request;
+
+interface StreamableTool extends Tool
+{
+    /**
+     * Stream the tool's activity and return its final tool result.
+     *
+     * @return Generator<int, StreamEvent, mixed, string>
+     */
+    public function streamHandle(Request $request): Generator;
+}
--- a/src/Contracts/StreamsActivity.php
+++ b/src/Contracts/StreamsActivity.php
@@ -1,0 +1,8 @@
+<?php
+
+namespace Laravel\Ai\Contracts;
+
+interface StreamsActivity
+{
+    //
+}
--- a/src/Gateway/Anthropic/Concerns/HandlesTextStreaming.php
+++ b/src/Gateway/Anthropic/Concerns/HandlesTextStreaming.php
@@ -408,7 +408,12 @@ trait HandlesTextStreaming
                 continue;
             }
 
-            $result = $this->executeTool($tool, $toolCall->arguments);
+            $result = yield from $this->executeToolStreamingStamped(
+                $tool,
+                $toolCall->arguments,
+                $invocationId,
+                $toolCall->id,
+            );
 
             $toolResult = new ToolResult(
                 $toolCall->id,
--- a/src/Gateway/Bedrock/BedrockTextGateway.php
+++ b/src/Gateway/Bedrock/BedrockTextGateway.php
@@ -479,7 +479,7 @@ class BedrockTextGateway implements EmbeddingGateway, TextGateway
 
             $conversationMessages[] = $this->buildAssistantConversationMessage($assistantText, $toolCalls, array_values($responseContent));
 
-            $toolResults = $this->executeToolCalls($tools, $toolCalls);
+            $toolResults = yield from $this->executeToolCallsStreaming($tools, $toolCalls, $invocationId);
 
             foreach ($toolResults as $toolResult) {
                 yield (new ToolResultEvent(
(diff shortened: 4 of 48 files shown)


pushpak1300:
Thanks for your pull request!

gracjankubicki:
Thanks for the feedback — I completely understand the maintenance concern with a PR of this size.

Before I extract the sub-agent behavior into a community package, I'd like to ask whether you'd be open to a much smaller follow-up PR that only adds the generic extension point such a package needs: **letting a tool emit stream events while it runs**.

Minimal scope:

- `StreamableTool` contract — a tool may `yield` `StreamEvent`s and `return` its final string result (~15 lines)
- `InvokesTools::executeToolStreaming()` — runs such tools through a generator; `executeTool()` and the non-streaming path stay untouched (~25 lines)
- a one-line change per provider streaming loop: `$result = yield from $this->executeToolStreaming(...)`
- two nullable provenance fields on `StreamEvent` (`parentInvocationId`, `parentToolCallId`), stamped by the gateway — a tool cannot know its own tool call id
- a correctness guard so `StreamableAgentResponse` computes `$text`/`$usage` from top-level events only (2 lines)

That's roughly ~250 lines including tests, with no default behavior changes — tools that don't implement the contract run exactly as they do today.

Everything sub-agent-specific from this PR (`StreamsActivity`, the `AgentTool` changes, the Vercel `data-subagent` mapping, nested ancestry, fixtures) would live in a community package built on top of that hook.

Alternatively, if you'd prefer a single broader seam: the per-turn tool-call loop is currently duplicated across all 11 provider gateways, in both the streaming and non-streaming paths. Extracting it into a pluggable executor (with the sequential default preserving today's behavior) would de-duplicate the gateways and give the community one extension point for both streamed tool activity and concurrent tool execution.

Would a PR with either scope be considered? Happy to open it for review.

pushpak1300:
i would love to revisit this MR when we merge and complete refactor https://github.com/laravel/ai/pull/652 to see how much changes we need to do.

gracjankubicki:
That makes a lot of sense — `TextGenerationLoop` centralizes exactly the seam this feature needs. Once #652 lands, the minimal hook becomes a small, single-place change in the loop (streamable tool execution + provenance stamping) instead of touching every provider gateway.

I'll rebase a minimal version on top of the new architecture once it's merged. In the meantime, happy to help converting the remaining providers to `TurnTextGateway` (starting with Anthropic) if that would speed things up.

gracjankubicki:
As promised — rebased on the new `TextGenerationLoop` architecture: #802. The result is ~10x smaller than this PR.

PR #701, opened 2026-06-10
Add missing return type generics to embeddings and reranking responses

dfinchenko:
The constructors of `EmbeddingsResponse` and `RerankingResponse` already declare their element types (`@param array<int, array<float>> $embeddings` and `@param array<int, RankedDocument> $results`), but the methods exposing that data were typed as bare `array` and `Collection`, causing IDEs and static analysis tools to lose the underlying element types.                                                                                                         
                                                                                                                                                                                                                                         
  - `EmbeddingsResponse::first()` now declares `@return array<float>`, matching the existing `Embeddings::fakeEmbedding()` annotation.                                                                                                   
  - `RerankingResponse::collect()` now declares `@return Collection<int, RankedDocument>`, matching the constructor annotation.                                                                                                          
  - `RerankingResponse::documents()` now declares `@return Collection<int, string>`, since it maps `RankedDocument::$document`.                                                                                                          
                                                                                                                                                                                                                                         
  This improves type inference and static analysis without changing runtime behavior.

### DIFF
--- a/src/Responses/EmbeddingsResponse.php
+++ b/src/Responses/EmbeddingsResponse.php
@@ -58,6 +58,8 @@ class EmbeddingsResponse implements Arrayable, Countable, IteratorAggregate, Jso
 
     /**
      * Get an iterator for the object.
+     *
+     * @return Traversable<int, array<float>>
      */
     public function getIterator(): Traversable
     {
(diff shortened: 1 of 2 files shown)


pushpak1300:
Thanks.

pushpak1300 (approved):

PR #702, opened 2026-06-11
Keep tool calls and results balanced when the step limit is reached

### LINKED ISSUE #699: Agentic Loop Exits Too Early Breaking Agent Response
### Summary

When a `laravel/ai` agent makes sequential tool calls, the agentic loop exits one step too early and returns an incomplete or broken response with no error surfaced to the user. The `shouldContinue` guard in `processResponse` has an off-by-one error that silently caps tool executions at `maxSteps - 1` instead of `maxSteps`.

We first noticed this when `disable_parallel_tool_use: true` was set via `providerOptions` and the model was making sequential tool calls (one per turn). The agent would return mid-analysis with no final answer, and the stored conversation showed `tool_calls` with more entries than `tool_results`. After adding debug logging to trace the agentic loop, we identified the off-by-one as the root cause.

### Environment

- `laravel/ai`: v0.8.1
- Models tested: `claude-sonnet-4-6`, `claude-opus-4-7`, `gpt-4o` (similar behavior observed across all)
- PHP: 8.5
- Laravel: v13

### Steps to Reproduce

1. Create an agent with **2 tools**.
   - `maxSteps` defaults to `round(2 * 1.5) = 3`
2. Ask a question that causes the model to make **3 sequential tool calls** (one per turn, with `disable_parallel_tool_use: true`).
3. The agent returns an incomplete response: the text preamble the model wrote before its 3rd tool call, with no final analysis.
4. Check stored conversation: `tool_calls` has 3 entries but `tool_results` has only 2.

### Root Cause

**File:** `src/Gateway/Anthropic/Concerns/ParsesTextResponses.php`

```php
$shouldContinue = $finishReason === FinishReason::ToolCalls
    && filled($realToolCalls)
    && $depth + 1 < ($maxSteps ?? round(count($tools) * 1.5)); // off-by-one
```

`depth` is 0-indexed. At depth=2 (the 3rd API response), `$depth + 1 = 3` is not `< 3`, so `$shouldContinue = false` even though the model's `stop_reason` is `tool_use`. The tool call is never executed and the agent returns whatever text appeared before the tool call in that turn.

Note: Other providers have their own `ParsesTextResponses.php` implementations with similar logic. The OpenAI gateway seems to have the same bug in a slightly different form. The step is pushed before the check, so `$steps->count() < $maxSteps` is equivalent to `$depth + 1 < $maxSteps`. The fix is `$steps->count() <= $maxSteps`. Confirmed working after patching `src/Gateway/OpenAi/Concerns/ParsesTextResponses.php`. Other providers likely have the same issue and are worth auditing.

Debug log confirming the exact failure point:

```
processResponse {"depth":0,"stop_reason":"tool_use","finish_reason":"tool_calls","maxSteps":3.0,"shouldContinue":true}
processResponse {"depth":1,"stop_reason":"tool_use","finish_reason":"tool_calls","maxSteps":3.0,"shouldContinue":true}
processResponse {"depth":2,"stop_reason":"tool_use","finish_reason":"tool_calls","maxSteps":3.0,"shouldContinue":false} ← bug
```

### Suggested Fix

Change `$depth + 1 <` to `$depth <`:

```php
$shouldContinue = $finishReason === FinishReason::ToolCalls
    && filled($realToolCalls)
    && $depth < ($maxSteps ?? round(count($tools) * 1.5));
```

With this fix, `maxSteps=3` correctly allows tool executions at depths 0, 1, and 2 (3 tool calls), with the final non-tool response at depth 3. Confirmed working after patching the Anthropic gateway.

### Impact

- Any Anthropic agent with `n` tools will silently cap tool calls at `maxSteps - 1` instead of `maxSteps`, causing incomplete responses when the model needs exactly `maxSteps` tool calls.
- The broken turn is persisted to the conversation store with missing tool results, leaving the conversation in an irrecoverable state. Subsequent turns in the same conversation will continue from the broken state with no way to recover the missing analysis.

### Related

#639 describes the same symptom (`tool_use ids were found without tool_result blocks`) and may share this root cause or be caused by the same off-by-one in a different scenario.

---

**Edit:** The original description incorrectly identified a `findTool null` silent skip as the root cause, and also suspected that `disable_parallel_tool_use: true` was not being honored. After adding verbose logging to trace the agentic loop step by step, both theories were ruled out. The actual cause is the off-by-one in the `$depth + 1 < $maxSteps` guard documented above.

AI helped write most of this issue and the proposed fix, but I am happy to help resolve as a human :)

slegarraga:
Fixes #699.

## Summary

- allow the final configured non-streaming tool-call step to execute before stopping the loop
- align Anthropic's depth-based guard with the other provider step-count guards
- tighten Anthropic, Gemini, and OpenRouter regressions around the boundary case where three tool-call turns are followed by a final text response

## Validation

- `./vendor/bin/pest tests/Feature/Providers/Anthropic/ToolCallLoopTest.php tests/Feature/Providers/OpenRouter/ToolCallLoopTest.php`
- `./vendor/bin/pest tests/Feature/Providers/Gemini/ToolCallLoopTest.php tests/Feature/Providers/OpenRouter/ToolCallLoopTest.php tests/Feature/Providers/AzureOpenAi/ToolCallLoopTest.php tests/Feature/Providers/Groq/ToolCallLoopTest.php tests/Feature/Providers/DeepSeek/ToolCallLoopTest.php tests/Feature/Providers/Xai/ToolCallLoopTest.php tests/Feature/Providers/Anthropic/ToolCallLoopTest.php tests/Feature/Providers/Ollama/ToolCallLoopTest.php tests/Feature/Providers/Mistral/ToolCallLoopTest.php tests/Feature/Providers/OpenAi/ToolCallLoopTest.php`
- `composer test:lint`
- `./vendor/bin/phpstan --memory-limit=-1`
- `composer test`
- `composer lint`

### DIFF
--- a/src/Gateway/Anthropic/Concerns/HandlesTextStreaming.php
+++ b/src/Gateway/Anthropic/Concerns/HandlesTextStreaming.php
@@ -354,7 +354,7 @@ trait HandlesTextStreaming
             return;
         }
 
-        if ($stopReason === 'pause_turn' && $depth < ($maxSteps ?? 5)) {
+        if ($stopReason === 'pause_turn' && $depth + 1 < ($maxSteps ?? 5)) {
             yield from $this->resumeFromPauseTurn(
                 $invocationId,
                 $provider,
@@ -429,7 +429,7 @@ trait HandlesTextStreaming
             ))->withInvocationId($invocationId);
         }
 
-        if ($depth >= ($maxSteps ?? round(count($tools) * 1.5))) {
+        if ($depth + 1 >= ($maxSteps ?? round(count($tools) * 1.5))) {
             yield (new StreamEnd(
                 $this->generateEventId(),
                 FinishReason::ToolCalls->value,
--- a/src/Gateway/Anthropic/Concerns/HandlesTextStreaming.php
+++ b/src/Gateway/Anthropic/Concerns/HandlesTextStreaming.php
@@ -354,7 +354,7 @@ trait HandlesTextStreaming
             return;
         }
 
-        if ($stopReason === 'pause_turn' && $depth + 1 < ($maxSteps ?? 5)) {
+        if ($stopReason === 'pause_turn' && $depth < ($maxSteps ?? 5)) {
             yield from $this->resumeFromPauseTurn(
                 $invocationId,
                 $provider,
@@ -429,7 +429,7 @@ trait HandlesTextStreaming
             ))->withInvocationId($invocationId);
         }
 
-        if ($depth + 1 >= ($maxSteps ?? round(count($tools) * 1.5))) {
+        if ($depth >= ($maxSteps ?? round(count($tools) * 1.5))) {
             yield (new StreamEnd(
                 $this->generateEventId(),
                 FinishReason::ToolCalls->value,
--- a/src/Gateway/Anthropic/Concerns/ParsesTextResponses.php
+++ b/src/Gateway/Anthropic/Concerns/ParsesTextResponses.php
@@ -98,10 +98,10 @@ trait ParsesTextResponses
 
         $shouldContinue = $finishReason === FinishReason::ToolCalls
             && filled($realToolCalls)
-            && $depth + 1 < ($maxSteps ?? round(count($tools) * 1.5));
+            && $depth < ($maxSteps ?? round(count($tools) * 1.5));
 
         $shouldResumePauseTurn = $stopReason === 'pause_turn'
-            && $depth + 1 < ($maxSteps ?? 5);
+            && $depth < ($maxSteps ?? 5);
 
         if ($shouldContinue) {
             $toolResults = $this->executeToolCalls($realToolCalls, $tools);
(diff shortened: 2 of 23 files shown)


pushpak1300 (approved):

pushpak1300 (approved):

orphanedrecord:
Hey fellas, I just noticed this fix seems to make the sync `->prompt()` play nice in `ParsesTextResponses.php`, but misses the streaming path. `HandlesTextStreaming.php` has the same off-by-one in a different form inside `handleStreamingToolCalls()` on line 432:

```
// current (bug)
if ($depth + 1 >= ($maxSteps ?? round(count($tools) * 1.5))) { 
```

```
// fix
if ($depth >= ($maxSteps ?? round(count($tools) * 1.5))) {
```

So for example, I'm testing with Anthropic, and `->stream()` vs. `-> prompt()` will still hit the bug regardless of the PR merging. Other providers with dedicated streaming handlers may also need auditing.

I need to dig into this further myself …

Hope it helps and appreciate all your work :)

pushpak1300:
I though about it in more detail and max steps and tool steps are 2 different thing. Steps include the initial ai request as well.

https://github.com/laravel/ai/issues/699#issuecomment-4710318159

PR #703, opened 2026-06-11
Merge all plain-object anyOf branches instead of keeping only the first

pushpak1300:
The normalizer currently collapses `anyOf`/`oneOf` to the first branch, so discriminated union schemas lose all variant-specific properties:

```php
// before — choices and scale are gone
{type: object, properties: {kind: {enum: ['open']}, text: ...}, required: [kind, text]}

// after
{type: object, properties: {kind: {enum: ['open', 'choice', 'rating']}, text: ..., choices: ..., scale: ...}, required: [kind]}
```

All plain-object branches are unioned into one schema. For a shared discriminator property with `enum` values in each branch, the enums are merged. The `required` intersection keeps only fields required in every opinionated branch.

Three correctness fixes are also included: `required` was incorrectly promoted when only one of multiple branches declared it; the `array_merge` fallback now uses `mergeSchema` to avoid shallow-overwriting already-merged properties when both `anyOf` and `oneOf` appear on the same node; and `is_array` guards prevent a `TypeError` from non-array `enum` values in malformed MCP tool schemas.

### DIFF
--- a/src/Schema/SchemaNormalizer.php
+++ b/src/Schema/SchemaNormalizer.php
@@ -231,6 +231,10 @@ class SchemaNormalizer
 
             [$branch, $branchSeen] = $this->inlineRefs($branch, $root, $seen);
 
+            if ($branch === []) {
+                continue;
+            }
+
             if (in_array($branch['type'] ?? null, ['null', ['null']], true)) {
                 $nullable = true;
             } else {
@@ -245,7 +249,8 @@ class SchemaNormalizer
         }
 
         if ($resolved !== []) {
-            $schema = $this->mergeObjectVariants($schema, $resolved) ?? $this->mergeSchema($schema, $resolved[0]);
+            $schema = $this->mergeObjectVariants($schema, $resolved)
+                ?? $this->mergeSchema($schema, $this->firstObjectBranch($resolved) ?? $resolved[0]);
         }
 
         if (! $nullable) {
@@ -259,7 +264,7 @@ class SchemaNormalizer
     }
 
     /**
-     * Union all plain-object variants' properties; null if any branch is not a plain object.
+     * Union every plain-object variant's properties, ignoring scalar branches; null when fewer than two objects remain.
      *
      * @param  array<string, mixed>  $schema
      * @param  array<int, array<string, mixed>>  $branches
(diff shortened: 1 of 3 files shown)


src/Schema/SchemaNormalizer.php
```diff
@@ -247,10 +249,126 @@ private function mergeUnion(array $schema, array $branches, array $root, array $
...
+            }
+
+            $closed = $closed && (($branch['additionalProperties'] ?? null) === false);
+        }
+
+        $result = $schema;
+        $result['type'] = $nullable ? ['object', 'null'] : 'object';
+
+        $outerProperties = is_array($schema['properties'] ?? null) ? $schema['properties'] : [];
+        $mergedProperties = array_merge($allProperties, $outerProperties);
```
WendellAdriel: I think this merge order can drop values from the union branches. If the `outer` schema defines a generic `kind` property, and each `anyOf` / `oneOf` branch refines `kind` with `const` or `enum` for example, `$allProperties` contains the merged enum but `$outerProperties` overwrites it because it is passed second to `array_merge()`.

We should recursively merge overlapping outer and branch property schemas instead.

src/Schema/SchemaNormalizer.php
```diff
@@ -247,10 +249,126 @@ private function mergeUnion(array $schema, array $branches, array $root, array $
...
+        $allProperties = [];
+        $requiredSets = [];
+        $nullable = false;
+        $description = null;
+        $closed = true;
+
+        foreach ($objects as $branch) {
+            $nullable = $nullable || is_array($branch['type'] ?? null);
+
+            foreach ($branch['properties'] ?? [] as $key => $prop) {
```
WendellAdriel: Duplicate properties across object variants seems to only merge when both sides have array `enum` values, while every other duplicate keeps the first branch and drops later branch schema.

This can lose variant-specific shape for nested objects, such as one branch defining `payload.a` and another defining `payload.b`.

We should do one recursive property-merge path for duplicate properties, then layer enum unioning into that path.

WendellAdriel: This loop assumes `$branch['properties']` is iterable. Since MCP schemas are external input, a malformed branch with scalar `properties` can raise an `ErrorException` before the normalization finishes.

We should guard this the same way the nearby normalizer code does, only iterating when `properties` is an array.

tests/Unit/Schema/SchemaNormalizerTest.php
```diff
@@ -383,6 +383,240 @@ function normalizesWithoutThrowing(array $raw): array
...
+    ]);
+
+    $item = $normalized['properties']['questions']['items'];
+
+    expect($item['type'])->toBe('object');
+    expect($item['properties'])->toHaveKeys(['type', 'text', 'choices', 'scale']);
+    expect($item['required'])->toBe(['type']);
+});
+
+test('it falls back to the first branch when anyOf branches are not all plain objects', function () {
```
WendellAdriel: This test name says the fallback applies when branches are not all plain objects, but the case only has one object and one scalar branch, but the implementation has a different path for `[object, object, scalar]`, where object variants are merged and scalar branches are ignored.

We should either rename this case or add explicit `[object, object, scalar]` coverage.

WendellAdriel (changes requested):

src/Schema/SchemaNormalizer.php
pushpak1300: Fixed.

WendellAdriel (approved):
LGTM :fire:

PR #710, opened 2026-06-15
Support native anyOf schemas

dbpolito:
This is a follow-up to https://github.com/laravel/framework/pull/60509

Now that Illuminate JSON Schema can represent `anyOf` natively, Laravel AI should preserve `anyOf` schemas instead of always collapsing them through the compatibility path.

Older framework versions remain supported: if native `AnyOfType` is not available, Laravel AI keeps using the existing fallback behavior.

### DIFF
--- a/functions.php
+++ b/functions.php
@@ -96,6 +96,10 @@ function generate_fake_data_for_json_schema_type(Type $type): mixed
             return $result;
         })(),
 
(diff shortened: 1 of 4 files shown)


pushpak1300:
Can you resolve conflicts ?

dbpolito:
@pushpak1300 i fixed the conflicts, but now after this recent anyOf work it ended up being a bit more complex on test side...

PR #712, opened 2026-06-16
Expose the provider's native response ID on `Meta`

### LINKED ISSUE #675: Feature: Expose the provider's native response ID on `Meta`
### Problem

Every provider returns a native response identifier (e.g. OpenAI's `resp_...` from the Responses API, xAI's response `id`). Providers reference this ID in their own dashboards, usage/billing exports, and support tickets, so it's the natural key for correlating an SDK call with provider-side records.

Today the SDK never surfaces this ID to the consumer:

- The OpenAI and xAI gateways already read it — [`ParsesTextResponses.php#L94`](https://github.com/laravel/ai/blob/v0.7.2/src/Gateway/OpenAi/Concerns/ParsesTextResponses.php#L94) (`$responseId = $data['id'] ?? ''`) — but only to feed `previous_response_id` on tool follow-ups ([`#L237`](https://github.com/laravel/ai/blob/v0.7.2/src/Gateway/OpenAi/Concerns/ParsesTextResponses.php#L237)), then discard it.
- It is not carried on [`Meta`](https://github.com/laravel/ai/blob/v0.7.2/src/Responses/Data/Meta.php#L13-L19) (which exposes only `provider` / `model` / `citations`), nor on `TextResponse`, `AgentResponse`, `Step`, or any event.
- The only identifier exposed to consumers is `invocationId`, a locally generated UUID ([`GeneratesText.php#L40`](https://github.com/laravel/ai/blob/v0.7.2/src/Providers/Concerns/GeneratesText.php#L40), `(string) Str::uuid7()`) that is meaningless to the provider.

So there is currently no way to obtain the provider's own response ID short of intercepting the raw HTTP response.

### Use cases

- **Billing / cost reconciliation** — match SDK usage records against the provider's usage/billing exports, which key on the native response ID.
- **Support** — quote the provider's response ID when opening a support ticket about a specific generation.
- **Observability / tracing** — attach the provider's ID to logs/spans so an SDK call can be traced to the exact provider-side request.
- **Debugging** — locate and inspect a specific call in the provider's dashboard.

### Proposed change

Add a nullable `responseId` to `Meta`, populated by each gateway from the provider's native ID. To keep existing positional `new Meta($provider, $model[, $citations])` call sites working, it is appended as the **last** constructor parameter:

```php
class Meta implements Arrayable, JsonSerializable
{
    public Collection $citations;

    public function __construct(
        public ?string $provider = null,
        public ?string $model = null,
        ?Collection $citations = null,
        public ?string $responseId = null, // provider's native response id, e.g. OpenAI "resp_..."
    ) {
        $this->citations = $citations ?? new Collection;
    }

    // toArray()/jsonSerialize() gain a 'response_id' key
}
```

The OpenAI and xAI gateways already capture the value, so wiring it through is essentially free there. Other gateways can populate it from their equivalent field (Anthropic message `id`, Gemini `responseId`, …) or leave it `null` when the provider returns none.

### Backwards compatibility

`responseId` is nullable, defaults to `null`, and is added as the last constructor parameter, so existing positional `Meta` construction and existing consumer code are unaffected. Providers that do not expose a response ID simply leave it `null`.

---

Related to #636 (exposing the underlying response headers) — same theme of surfacing more of the underlying provider response.

I'm happy to open a PR for this if the direction sounds good.

wadakatu:
Closes #675.

## Problem

Every provider returns a native response identifier (OpenAI Responses API `resp_...`, xAI response `id`, Anthropic message `id`, Gemini `responseId`, OpenAI-compatible providers `chatcmpl-...`). Providers reference this ID in their dashboards, usage/billing exports, and support tickets, so it's the natural key for correlating an SDK call with provider-side records.

Until now the SDK never surfaced it: the OpenAI/xAI gateways already read `$data['id']`, but only to feed `previous_response_id` on tool follow-ups, then discarded it. `Meta` exposed only `provider` / `model` / `citations`. The only identifier available to consumers was `invocationId`, a locally generated UUID that is meaningless to the provider.

## Change

Add a nullable `responseId` to `Meta`, appended as the **last** constructor parameter so existing positional `new Meta($provider, $model[, $citations])` call sites are unaffected. `toArray()` / `jsonSerialize()` gain a `response_id` key.

Each text gateway populates it from the provider's native field:

| Provider | Source |
| --- | --- |
| OpenAI / xAI | `$data['id']` (`resp_...`) |
| Anthropic | `$data['id']` (`msg_...`) |
| Gemini | `$data['responseId']` |
| Groq / DeepSeek / OpenRouter | `$data['id']` (`chatcmpl-...`) |

**Ollama and Bedrock are intentionally left `null`:** Ollama's chat response carries no id field, and Bedrock's request id is only returned in HTTP headers (not the response body) — both are out of scope for this body-level change.

Streaming responses are likewise out of scope for now (the id arrives through stream events on a different code path); streamed `Meta` keeps `null` and can be wired in a follow-up.

## Backwards compatibility

`responseId` is nullable, defaults to `null`, and is the last constructor parameter, so existing positional construction and consumer code are unaffected. The OpenAI/xAI `previous_response_id` tool-loop behavior is untouched (that variable is left exactly as-is; only the value passed to `Meta` is normalized to `null` when empty).

## Tests

- Unit (`MetaTest`): defaults to `null`, stores the value when passed, and exposes `response_id` in `toArray()` / `jsonSerialize()`.
- Feature: each of the seven wired providers asserts that `$response->meta->responseId` (and the step-level `Meta`) carries the native id.

`composer test` (1537 passed), `pint`, and `phpstan` all pass.

---

Related to #636 (exposing the underlying response headers) — same theme of surfacing more of the underlying provider response.

### DIFF
--- a/src/Gateway/Anthropic/Concerns/ParsesTextResponses.php
+++ b/src/Gateway/Anthropic/Concerns/ParsesTextResponses.php
@@ -88,7 +88,8 @@ trait ParsesTextResponses
         $citations = $this->extractCitations($content);
         $usage = $this->extractUsage($data);
         $finishReason = $this->extractFinishReason($data);
-        $meta = new Meta($provider->name(), $model, $citations);
+        $responseId = $data['id'] ?? null;
+        $meta = new Meta($provider->name(), $model, $citations, $responseId);
 
         $realToolCalls = array_filter($toolCalls, fn (ToolCall $tc) => $tc->name !== 'output_structured_data');
         $hasStructuredToolCall = count($realToolCalls) < count($toolCalls);
--- a/src/Gateway/DeepSeek/Concerns/ParsesTextResponses.php
+++ b/src/Gateway/DeepSeek/Concerns/ParsesTextResponses.php
@@ -92,6 +92,7 @@ trait ParsesTextResponses
         $choice = $data['choices'][0] ?? [];
         $message = $choice['message'] ?? [];
         $model = $data['model'] ?? '';
+        $responseId = $data['id'] ?? null;
 
         $text = $message['content'] ?? '';
         $rawToolCalls = $message['tool_calls'] ?? [];
@@ -111,7 +112,7 @@ trait ParsesTextResponses
             [],
             $finishReason,
             $usage,
-            new Meta($provider->name(), $model),
+            new Meta($provider->name(), $model, responseId: $responseId),
         );
 
         $steps->push($step);
@@ -138,7 +139,7 @@ trait ParsesTextResponses
                 $toolResults,
                 $finishReason,
                 $usage,
-                new Meta($provider->name(), $model),
+                new Meta($provider->name(), $model, responseId: $responseId),
             ));
 
             $toolResultMessage = new ToolResultMessage(collect($toolResults));
@@ -172,7 +173,7 @@ trait ParsesTextResponses
                 $structuredData,
                 $text,
                 $this->combineUsage($steps),
-                new Meta($provider->name(), $model),
+                new Meta($provider->name(), $model, responseId: $responseId),
             ))->withToolCallsAndResults(
                 toolCalls: $allToolCalls,
                 toolResults: $allToolResults,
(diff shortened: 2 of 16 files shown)


pushpak1300:
https://github.com/laravel/ai/issues/675#issuecomment-4717591010

PR #713, opened 2026-06-16
feat: allow passing an mcp server to the agent to expose its tools

calebdw:
Hello!

## Support passing local MCP servers directly to agents

When an MCP server lives inside the same application, requiring agents to manually return individual `Laravel\Mcp\Server\Tool` instances (or hard-code URLs) is awkward and fragile across environments.

This adds `McpServer` so agents can reference mcp servers in the same application the same way they already spread remote client tools:

```php
use Laravel\Ai\Tools\McpServer;
use App\Mcp\Servers\AppToolsServer;

{
    return [
        ...McpServer::make(AppToolsServer::class),
        new SomeOtherTool,
    ];
}

// or simply list the server
{
    return [AppToolsServer::class];
}
```

- `McpServer::make($server)` (or passing a server class/instance) auto-expands to wrapped tools.
- Fully in-process (no transport/URLs), consistent with existing `McpServerTool` behavior for single tools.
- Works for `prompt` + `stream`, mixes cleanly with other tools/agents/provider tools.
- No changes to individual tool usage or remote MCP clients.

Thanks!

### DIFF
--- a/src/Contracts/HasTools.php
+++ b/src/Contracts/HasTools.php
@@ -11,8 +11,6 @@ interface HasTools
      * Agent instances (treated as sub-agents), raw MCP client/server primitives,
      * MCP server classes/instances, or McpServer wrappers. The SDK resolves and
      * wraps values as needed before passing them to the model.
-     *
-     * @return iterable
      */
     public function tools(): iterable;
 }
--- a/src/Contracts/HasTools.php
+++ b/src/Contracts/HasTools.php
@@ -2,14 +2,17 @@
 
 namespace Laravel\Ai\Contracts;
 
-use Laravel\Ai\Providers\Tools\ProviderTool;
-
 interface HasTools
 {
     /**
      * Get the tools available to the agent.
      *
-     * @return array<Tool|ProviderTool>
+     * Items may include Tool instances, ProviderTool instances (e.g. WebSearch),
+     * Agent instances (treated as sub-agents), raw MCP client/server primitives,
+     * MCP server classes/instances, or McpServer wrappers. The SDK resolves and
+     * wraps values as needed before passing them to the model.
+     *
+     * @return iterable
      */
     public function tools(): iterable;
 }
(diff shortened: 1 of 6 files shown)


pushpak1300:
Thanks for your pull request!

In this case, exposing an in-process MCP server's tools is a fairly niche use case, and it's already achievable without any additional API. You can list the individual tools directly:

```php
{
    return [
        new CurrentWeatherTool,
        new SomeOtherTool,
    ];
}
```

Or spread a server's resolved tools via the public context:

```php
{
    return [
        ...(new WeatherServer($transport))->createContext()->tools(),
        new SomeOtherTool,
    ];
}
```

The SDK already detects raw `Laravel\Mcp\Server\Tool` instances and wraps them, so this works as-is.

calebdw:
Hmm, that's all well and good---but not very intuitive and it certainly isn't mentioned in the docs. Would you be open to making it easier to include all the tools from a server that's already inside the application? Ideally I just want to be able to pass the class string or something like that

pushpak1300:
i think make sense to expose tools, resource and prompts via some helper method.

calebdw:
@pushpak1300, do you have a plan to do that?

Alternatively, it would be better in my opinion to have a `servers()` method, and then the ai client can take advantage of any server's tools, resources, and prompts

calebdw:
Also, I'm getting the following phpstan error when I try using your method:

```php
    #[Override]
    {
        return ([
            ...new EmsMcpServer(new FakeTransporter())->createContext()->tools(),
            ...Mcp::client('gitlab')->tools(),
        ];
    }
```

```
  90     Method WorkItemResolutionAgent::tools() should return iterable<Laravel\Ai\Contracts\Tool|Laravel\Ai\Providers\Tools\ProviderTool> but returns list<Laravel\Mcp\Client\Primitives\Tool|Laravel\Mcp\Server\Tool>.  
         🪪  return.type                                                                                                                                                                                                                            
```

PR #718, opened 2026-06-17
Fix Anthropic structured output rejecting JSON Schema constraints

### LINKED ISSUE #715: Anthropic native structured output rejects JSON Schema validation keywords (min/max, minLength, minItems, …) with a 400
### Summary

When an agent's `HasStructuredOutput::schema()` uses any value-constraint keyword
(`min`/`max` on numbers, strings, or arrays), requests to Anthropic fail with a `400`.

Anthropic's native structured output (`output_config.format` with `type: json_schema`)
only accepts a strict JSON Schema subset. Per the official docs, *"If you use an
unsupported feature, you'll receive a 400 error with details."* — so a schema containing
`minimum`/`maximum`/`multipleOf`, `minLength`/`maxLength`, or array bounds beyond
`minItems` of 0/1 fails the whole request.

### Reproduction

```php
use Laravel\Ai\Attributes\Provider;
use Laravel\Ai\Enums\Lab;
use Laravel\Ai\Promptable;

#[Provider(Lab::Anthropic)]
class ReviewAgent implements Agent, HasStructuredOutput
{
    use Promptable;

    {
        return 'You are a helpful assistant that reviews code.';
    }

    {
        return [
            'score'   => $schema->integer()->required()->min(1)->max(10),
            'summary' => $schema->string()->required()->min(1)->max(280),
            'tags'    => $schema->array()->required()->items($schema->string())->min(1)->max(5),
        ];
    }
}
```

Running this against an Anthropic model returns a `400` from the Messages API, because
`->min()`/`->max()` compile to `minimum`/`maximum`/`minLength`/`maxLength`/`maxItems`,
which `output_config.format` does not accept.

### Expected behavior

A schema built with the documented `JsonSchema` builder should work against every
provider. And when a user writes `->min(1)->max(10)`, they expect the returned value to
actually fall within that range — the constraint is part of the contract they opted into,
not a hint. So the fix should both stop the `400` and guarantee the returned value honors
the constraints. (Same guarantee gap as #189: native structured output enforces the
*structural* contract — types/required/enums — but not these value constraints.)

This is exactly what the official Anthropic SDKs (Python, TypeScript, Ruby, PHP) do. Per
the structured outputs docs, they transform a schema with unsupported features by:

> 1. **Remove unsupported constraints** (for example, `minimum`, `maximum`, `minLength`, `maxLength`)
> 2. **Update descriptions** with constraint info (for example, "Must be at least 100"), when the constraint is not directly supported with structured outputs
> 3. **Add `additionalProperties: false`** to all objects
> 4. **Filter string formats** to supported list only
> 5. **Validate responses** against your original schema (with all constraints)

### Proposed direction

Mirror this behavior:

1. **Strip unsupported keywords before sending**, folding each into the node's
   `description` as a natural-language note (e.g. `Constraints: minimum 1, maximum 10.`) so
   the model still honors it on the first attempt.
2. **Validate the parsed response against the full, constraint-bearing schema** afterwards,
   re-prompting with the specific violations until the output conforms (or retries run out).

### References

- Structured outputs — JSON Schema limitations & SDK transformation behavior:
  https://platform.claude.com/docs/en/build-with-claude/structured-outputs
- Related issue: #189 (structured output validation guarantees)

### Environment

- `laravel/ai`: v0.8.1 (also reproducible on the current `0.x` branch)
- Provider: Anthropic (native structured output via `output_config.format`, introduced in #309)

sulimanbenhalim:
Currently a schema built with the documented `JsonSchema` builder fails outright on Anthropic. Any value constraint you add compiles to a standard JSON Schema keyword (`->min()` becomes `minimum`, `minLength` or `minItems` depending on the type), and Anthropic's native structured output (`output_config.format`) only accepts a strict subset of JSON Schema. `ObjectSchema::toSchema()` passes the whole schema through untouched, so the request comes back as a 400 before the model even runs. The same agent works fine on OpenAI.

```php
{
    return [
        'score' => $schema->integer()->required()->min(1)->max(10),
        'summary' => $schema->string()->required()->min(1)->max(280),
        'tags' => $schema->array()->required()->items($schema->string())->min(1)->max(5),
    ];
}

// Anthropic: 400 invalid_request_error
// OpenAI: fine
```

This PR adds `AnthropicSchemaSanitizer` on the native structured output path. It strips the keywords Anthropic documents as unsupported and folds each one into that node's description, so the constraint still reaches the model as an instruction:

```php
$schema->integer()->required()->min(1)->max(10);

// sent as
[
    'type' => 'integer',
    'description' => 'Must be at least 1. Must be at most 10.',
]
```

### Current behaviour

The reject list is taken from Anthropic's structured output docs and covers the standard vocabulary they do not accept:

| Keywords | Result |
| --- | --- |
| `minimum`, `exclusiveMinimum`, `maximum`, `exclusiveMaximum`, `multipleOf` | stripped, described |
| `minLength`, `maxLength`, `pattern` | stripped, described |
| `maxItems`, `uniqueItems`, `contains`, `minContains`, `maxContains`, `prefixItems`, `unevaluatedItems` | stripped, described where prose reads usefully |
| `minProperties`, `maxProperties`, `patternProperties`, `propertyNames`, `dependentRequired`, `dependentSchemas`, `unevaluatedProperties` | stripped, described where prose reads usefully |
| `not`, `if`, `then`, `else`, `examples`, `deprecated`, `readOnly`, `writeOnly`, `contentEncoding`, `contentMediaType`, `contentSchema`, `$schema`, `$id`, `$anchor`, `$comment`, `$dynamicRef`, `$dynamicAnchor`, `$vocabulary` | stripped silently, prose would just be noise |

A few shapes get rewritten instead of dropped, because Anthropic accepts an equivalent form:

- `minItems` greater than 1 is clamped to 1, since 0 and 1 are the accepted values. The real count goes into the description, so the non-empty guarantee survives.
- `oneOf` becomes `anyOf`.
- A union `type` array (what `->nullable()` serializes to) becomes `anyOf` branches.
- An `enum` holding objects or arrays folds into the description, since only scalar members are accepted.
- `additionalProperties` is forced to `false`.

Anything not on the reject list passes through untouched. That part was deliberate. I built it the other way round first, keeping only a known-good allowlist and folding everything else into prose, and the failure mode is worse: a keyword Anthropic actually supports but we forgot to list gets silently demoted to a description hint and nobody ever reports it, because nothing visibly breaks. A stale reject list instead produces a 400 that names the offending keyword, which tells you exactly what to add.

Only the native path is touched. The synthetic-tool fallback still sends the schema verbatim, since tool `input_schema` is not held to the structured output subset.

### One thing worth flagging

This stops the 400, but the constraints become guidance rather than something the API enforces. A model can still hand back `score: 42` and nothing in the library rejects it. Anthropic's own SDKs pair the same transform with client-side validation against the original schema, which we do not do for any provider today (same gap as #189). Happy to follow that up separately, though it needs a JSON Schema validator dependency and a decision that applies across providers rather than just Anthropic.

Tests updated accordingly.

Fixes #715


pushpak1300:
AI-REVIEW: _Automated review — Claude Opus 4.8 running `/code-review` (high) plus a multi-axis pass. All 35 new tests pass on the branch; CI green._

Correct, well-tested fix. Every constraint Anthropic's `output_config` json_schema rejects is stripped, and nothing it accepts is stripped — cross-checked against Anthropic's structured-outputs spec (pattern/default kept, `minItems` clamped to 0/1, `maxItems`/`uniqueItems`/numeric/string constraints stripped, `format` allowlist exact). The `match()` arms are exhaustive and the synthetic-tool fallback is correctly left unsanitized.

**Should-fix**

- `src/Gateway/Anthropic/AnthropicSchemaSanitizer.php:28-29` — the `sanitize()` docblock description spans two lines. AGENTS.md requires the docblock description on a single line. Collapse into one line (every other docblock in the PR complies).

**Nits**

- `sanitize()` recurses into `properties`, `items`, and `anyOf`/`oneOf`/`allOf`, but not `$defs`/`definitions`/`$ref` or tuple `prefixItems`. Not a live bug — the sole caller feeds `ObjectSchema::toSchema()`, and the Illuminate builder always inlines and never emits those keywords. But since Anthropic *does* accept `$ref`/`$defs`, a future caller passing constraints nested inside `$defs` would have them slip through. Optional to guard now.
- `exclusiveMinimum`/`exclusiveMaximum` (`AnthropicSchemaSanitizer.php:11`) are handled defensively but never emitted by the builder; the numeric-note wording assumes the draft-2020-12 numeric form. Unreachable via the current path; harmless.

sulimanbenhalim:
fixed the one-liner

pushpak1300:
Thanks..

PR #720, opened 2026-06-18
Add collection generics to AssistantMessage toolCalls

dfinchenko:
`AssistantMessage::$toolCalls` is an untyped `Collection`, so static analysis and IDEs see its elements as `mixed`. This types it as `Collection<int, ToolCall>` (property and constructor `@param`), matching how the collection is always                                                                                                                                                             
  populated:                                                                                                                                                                                                                                                                                                                                                                                                                                                   
  - `DatabaseConversationStore` builds it via `$toolCalls->map(ToolCall::fromArray(...))`.                                                                                                                                               
  - Every gateway passes a collection of `ToolCall` instances (`collect($mappedToolCalls)`, `new Collection($toolCalls)`, etc.).

### DIFF
--- a/src/Messages/AssistantMessage.php
+++ b/src/Messages/AssistantMessage.php
@@ -3,9 +3,11 @@
 namespace Laravel\Ai\Messages;
 
 use Illuminate\Support\Collection;
+use Laravel\Ai\Responses\Data\ToolCall;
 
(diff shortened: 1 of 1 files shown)


pushpak1300:
thanks.

PR #721, opened 2026-06-18
Add response cost reporting and per-agent budget guardrails

sulimanbenhalim:
Adds a cost layer to the SDK. Every response can now tell you what it actually cost in dollars, and agents can carry a hard spend budget that stops them before they blow past it.

Today the SDK surfaces raw token counts on `$response->usage`, but nothing turns those into money, and there's no way to cap what an agent or a runaway tool loop is allowed to spend. This adds both, computed entirely from data that already comes back on every response (the `usage` plus the `provider` and `model` on `meta`), so it works across every provider with no gateway changes.

cost-card

**The cost API**

`$response->cost()` returns a `Cost` value object built from the response usage and a per-model price table:

```php
$response = (new ResearchAgent)->prompt('Summarize Q3 earnings');

$response->cost()->total();      // 0.0032
$response->cost()->format();     // "$0.0032"
$response->cost()->breakdown();  // ['input' => 0.000174, 'output' => 0.00303, ...]
```

Pricing is config driven. The SDK ships a small default table (USD per 1M tokens) that you can override or extend in `config/ai.php`, or at runtime with `PriceList::register()`. A model with no known price returns an unknown cost (`isKnown()` is false, total `0`) instead of throwing, so cost reporting is always safe. Cost also rides every `Step`, so `$response->steps->sum(fn ($s) => $s->cost()->total())` just works.

**Budgets**

Tag an agent with `#[MaxCost]` and it stops once it has spent its budget within the process, throwing `BudgetExceededException` before the next call:

```php
#[MaxCost(0.50)]
class ResearchAgent implements Agent
{
    use Promptable;
}
```

cost-dashboard

The figures above are real. They come from running three live Claude calls (opus, sonnet, haiku) through the SDK and pricing the actual token usage that came back.

**Why this doesn't break anything**

It is purely additive. No existing signature, response shape, or behavior changes:

- new classes only: `Cost`, `Pricing\{ModelPricing, PriceList, CostCalculator, BudgetMeter}`, the `EnforceBudget` middleware, the `MaxCost` attribute, and `BudgetExceededException`
- two new methods (`cost()`) on `TextResponse` and `Step`
- three singleton bindings, plus one additive `#[MaxCost]` branch in `gatherMiddlewareFor()` that only fires for agents that opt in
- a new `pricing` config key, merged via `mergeConfigFrom` so already published configs keep working Gateways, providers, events, and the tool loop are untouched.

**Scope**

v1 covers cost reporting and per-process / per-conversation budgets, enforced pre-flight through the existing middleware pipeline. _A hard mid-run abort (stopping inside a single multi-step tool loop) would need a small hook in the gateway loop, so it is left for a follow-up._

**Tests**

Added unit coverage for the cost math, price resolution (exact match, versioned-id matching, and config/runtime overrides), and the budget meter, plus feature tests that assert `$response->cost()` on a real response and the `#[MaxCost]` tripwire firing once the budget is spent. Full suite passes, Pint and PHPStan clean.

### DIFF
--- a/config/ai.php
+++ b/config/ai.php
@@ -40,6 +40,27 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Model Pricing
+    |--------------------------------------------------------------------------
+    |
+    | Per-model token rates, in USD per 1,000,000 tokens, used to compute the
+    | cost of responses via $response->cost(). Values defined here are merged
+    | over the SDK's built-in defaults, so you only need to list models you
+    | want to add or override. Models without pricing simply report an
+    | "unknown" cost rather than throwing.
+    |
+    */
+
+    'pricing' => [
+        'models' => [
+            // 'openai' => [
+            //     'gpt-4o' => ['input' => 2.50, 'output' => 10.00],
+            // ],
+        ],
+    ],
+
+    /*
+    |--------------------------------------------------------------------------
     | AI Providers
     |--------------------------------------------------------------------------
     |
--- a/src/AiServiceProvider.php
+++ b/src/AiServiceProvider.php
@@ -12,6 +12,9 @@ use Laravel\Ai\Console\Commands\MakeAgentMiddlewareCommand;
 use Laravel\Ai\Console\Commands\MakeToolCommand;
 use Laravel\Ai\Contracts\ConversationStore;
 use Laravel\Ai\Enums\Lab;
+use Laravel\Ai\Pricing\BudgetMeter;
+use Laravel\Ai\Pricing\CostCalculator;
+use Laravel\Ai\Pricing\PriceList;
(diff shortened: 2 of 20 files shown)


pushpak1300:
Thanks for your pull request!

sulimanbenhalim:
Thanks again for the review! I took your suggestion and released this as a standalone package:

**https://github.com/sulimanbenhalim/laravel-ai-cost** — `composer require sulimanbenhalim/laravel-ai-cost`

It's fully non-invasive: no changes to the SDK, it works through the existing lifecycle events. So anyone who finds this PR and wants dollar cost reporting (`AiCost::for($response)`) and per-agent `#[MaxCost]` budgets can pull it in alongside laravel/ai. Hope it's useful to others. 🙏

PR #722, opened 2026-06-18
Route top-level Ollama provider options out of the options object

cjmcintyre:
Currently, provider options set for Ollama via `HasProviderOptions` (`providerOptions(Lab::Ollama)`) are all merged into the request body's `options` object - Ollama's *model-parameters* bag (`temperature`, `top_p`, `num_ctx`, `seed`, …). 

But a few Ollama request fields are top-level rather than model parameters, notably `format` (Ollama's JSON-mode / structured-output switch) and `keep_alive`. 

Ollama ignores them under `options`, so there's currently no way to set `format` through provider options - you're pushed into the full schema path even when all you want is `"format": "json"`:

```jsonc
// providerOptions(Lab::Ollama) => ['format' => 'json'] today:
{ "model": "…", "messages": […], "stream": false, "options": { "format": "json" } }
// Ollama ignores options.format, so JSON mode never engages.
```

`buildChatRequestBody()` now hoists the keys Ollama documents at the top level of `POST /api/chat` (`format`, `keep_alive`, `think`, `logprobs`, `top_logprobs`) to the body root and keeps everything else under `options`:

```jsonc
{ "model": "…", "messages": […], "stream": false, "format": "json" }
```

Genuine model parameters (`top_k`, `repeat_penalty`, `num_ctx`, `seed`, …) are untouched and still sent under `options`, and a schema-driven `format` (structured output) still wins over a provider-options `format`. This mirrors how the Gemini gateway already hoists top-level keys out of `generationConfig`. Tests added.

### DIFF
--- a/src/Gateway/Ollama/Concerns/BuildsTextRequests.php
+++ b/src/Gateway/Ollama/Concerns/BuildsTextRequests.php
@@ -69,6 +69,17 @@ trait BuildsTextRequests
 
         $providerOptions = $options?->providerOptions($provider->driver()) ?? [];
 
+        // Hoist keys that belong at the top level, as everything else is passed in options.
+        $topLevelKeys = ['format', 'keep_alive', 'think', 'logprobs', 'top_logprobs'];
+
+        foreach ($topLevelKeys as $key) {
+            if (array_key_exists($key, $providerOptions)) {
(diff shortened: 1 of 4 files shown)


pushpak1300:
Thanks.

PR #724, opened 2026-06-18
Add cross-provider service tier support

serpentblade:
## Summary

Most inference providers now expose a "service tier" knob to trade latency
against cost — priority/faster vs flex/cheaper than the default. There's
currently no way to set it through this package. This adds an optional,
opt-in service tier that an agent can declare once and have applied to
whichever provider it runs against.

## Usage

An agent declares a tier via a no-arg `serviceTier()` method or a
`#[ServiceTier]` attribute. Both accept a raw string or a provider-specific
enum:

```php
use Laravel\Ai\Enums\Bedrock\ServiceTier;

{
    public function serviceTier(): ServiceTier|string|null
    {
        return ServiceTier::Priority; // or 'priority'
    }
}

// or, declaratively:
#[ServiceTier('priority')]
class SupportAgent implements Agent { /* ... */ }
```

## How it works

- Resolved onto `TextGenerationOptions::$serviceTier` (a plain `string`), so
  unknown/future provider values keep working without library changes. The
  per-provider string-backed enums (`Laravel\Ai\Enums\<Provider>\ServiceTier`)
  are optional sugar for discoverability and normalize to their value.
- Each gateway that has a native tier concept writes it into that provider's
  own field. Raw `providerOptions()` still take precedence, so they remain the
  escape hatch.
- An empty string is treated as "unset" and never forwarded.

| Provider | Field | Accepted values |
|---|---|---|
| Bedrock | `serviceTier: { type }` | `priority` / `default` / `flex` / `reserved` |
| OpenAI (and Azure) | `service_tier` | `auto` / `default` / `flex` / `priority` |
| Groq | `service_tier` | `auto` / `on_demand` / `flex` / `performance` |
| OpenRouter | `service_tier` | `flex` / `priority` |
| Anthropic | `service_tier` | `auto` / `standard_only` |

Providers without a request-level tier concept (DeepSeek, Mistral, xAI,
Gemini, Ollama) ignore it. Because the option defaults to `null`, nothing is
ever sent unless an agent opts in — this is fully backward compatible.

## Tests

- Resolution unit tests: method/attribute, raw string + provider enum,
  method-over-attribute precedence, empty-string-as-unset, null default.
- Per-provider request-mapping tests for all five wired gateways, including
  that `providerOptions` overrides the resolved tier.
- Full suite green; Pint and PHPStan clean.

---

Claude (Anthropic) assisted with this implementation. I have reviewed, tested,
and understand all of the code, and take responsibility for it as my own
contribution, per Laravel's AI-generated contributions guidance.

### DIFF
--- a/src/Attributes/ServiceTier.php
+++ b/src/Attributes/ServiceTier.php
@@ -1,0 +1,21 @@
+<?php
+
+namespace Laravel\Ai\Attributes;
+
+use Attribute;
+use BackedEnum;
+
+#[Attribute(Attribute::TARGET_CLASS)]
+class ServiceTier
+{
+    public string $value;
+
+    /**
+     * Accepts a raw provider string (e.g. 'priority') or any provider service
+     * tier enum, normalizing it down to its string value.
+     */
+    public function __construct(string|BackedEnum $value)
+    {
+        $this->value = $value instanceof BackedEnum ? (string) $value->value : $value;
+    }
+}
--- a/src/Contracts/ServiceTier.php
+++ b/src/Contracts/ServiceTier.php
@@ -1,0 +1,18 @@
+<?php
+
+namespace Laravel\Ai\Contracts;
+
+use BackedEnum;
+
+/**
+ * Marker interface implemented by every provider's service tier enum.
+ *
+ * Service tiers let an agent opt into a faster (higher priority) or cheaper
+ * (lower priority) processing tier than the provider default. Each provider
+ * exposes its own native values, so the enums live under the provider's own
+ * namespace and resolve down to a plain string before reaching the gateway.
+ */
+interface ServiceTier extends BackedEnum
+{
+    //
+}
--- a/src/Enums/Anthropic/ServiceTier.php
+++ b/src/Enums/Anthropic/ServiceTier.php
@@ -1,0 +1,20 @@
+<?php
+
+namespace Laravel\Ai\Enums\Anthropic;
+
+use Laravel\Ai\Contracts\ServiceTier as ServiceTierContract;
+
+/**
+ * Service tiers supported by the Anthropic Messages API.
+ *
+ * Anthropic has no explicit "priority" request value: send "auto" to
+ * opportunistically use priority capacity when available, or "standard_only"
+ * to always use standard capacity.
+ *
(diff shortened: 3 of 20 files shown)


pushpak1300:
can you share any docs about this ?

pushpak1300:
Thanks for putting this together, and for the thorough tests — I really appreciate the effort here.

One correction on the premise, though: there is already a way to configure this through the package today. Every text gateway merges `providerOptions()` into the top level of the request payload (Anthropic, OpenAI/Azure, Groq, and OpenRouter all do `array_merge($body, $providerOptions)`, while Bedrock merges into the top-level Converse parameters where `serviceTier` lives). That means an agent can already set the appropriate tier for any supported provider without any library changes:

```php
{
    return match ($provider) {
        Lab::Anthropic  => ['service_tier' => 'auto'],
        Lab::OpenAI     => ['service_tier' => 'flex'],
        Lab::Groq       => ['service_tier' => 'flex'],
        Lab::OpenRouter => ['service_tier' => 'priority'],
        Lab::Bedrock    => ['serviceTier' => ['type' => 'priority']],
        default         => [],
    };
}
```

The more difficult problem is the "declare once and apply everywhere" aspect. There isn't a shared cross-provider vocabulary for service tiers. Anthropic uses `auto` / `standard_only`, Groq uses `auto` / `on_demand` / `flex` / `performance`, Bedrock uses `priority` / `default` / `flex` / `reserved`, and so on.

Because this PR resolves to a single string and writes that value directly into each provider's field, a declaration like `serviceTier(): 'priority'` would be valid for OpenAI, OpenRouter, and Bedrock, but invalid for Anthropic and Groq. Since `serviceTier()` also has no provider context, it can't vary its behavior across a failover chain — which is exactly the scenario that `providerOptions($provider)` already handles correctly.

Given that, I'd rather not introduce a new contract, attribute, and provider-specific gateway wiring on top of functionality that's already supported. The PR also retains `providerOptions()` as the ultimate override, which reinforces that it's the appropriate extension point.

I'm going to close this in favor of the existing `providerOptions()` pattern shown above.

serpentblade:
Was worth a shot. At least this is now clearly documented for anybody that comes looking.

PR #725, opened 2026-06-19
Parse Cohere Embed v4's type-keyed embeddings shape on Bedrock

serpentblade:
### Summary

Cohere returns the `embeddings` field of an embeddings response in one of two
shapes — a bare list of vectors, or an object keyed by embedding type:

```json
{ "embeddings": [[0.1, 0.2, ...]], "response_type": "embeddings_floats" }
{ "embeddings": { "float": [[0.1, 0.2, ...]] }, "response_type": "embeddings_by_type" }
```

Which one you get follows the requested `embedding_types`, not the model
version — both Embed v3 and Embed v4 accept the parameter, and Cohere's own v2
API always returns the keyed shape. `BedrockTextGateway::generateCohereEmbeddings()`
handled only the bare list, running `array_filter($result['embeddings'], 'is_array')`
over the envelope. Against the keyed shape that keeps the `float` *value* and
produces a list-of-list-of-vectors, so `EmbeddingsResponse::first()` returns
`[[...floats]]` where callers expect a flat vector.

### Changes

1. **Vector shape** — a `ParsesEmbeddings` concern normalizes both shapes into a
   list of vectors, so `first()`/`all()` are flat regardless of which envelope
   the model returns. When a keyed response carries no `float` key (because
   `embedding_types` was overridden through provider options), an `AiException`
   names the types that were returned rather than silently handing back
   quantized vectors through an API that promises floats.

2. **Shared with the Cohere provider** — `CohereGateway` previously hard-coded
   `$data['embeddings']['float']`, which fataled on any response without that
   key. Both gateways now read one envelope the same way.

3. **Model identifiers** — dispatch matches `cohere.embed-` anywhere in the id,
   so the geo and global inference profile ids documented for Embed v4
   (`us.cohere.embed-v4:0`, `eu.cohere.embed-v4:0`, `global.cohere.embed-v4:0`)
   take the Cohere path instead of falling through to the Titan branch and
   failing validation on a Titan-shaped body.

4. **Output dimension** — `generateEmbeddings()` dropped `$dimensions` when it
   routed to the Cohere path, so a requested width was silently ignored and the
   model's default returned instead. A requested dimension is now forwarded as
   `output_dimension`. See below for why this one grew past the Bedrock gateway.

### Why the dimension fix ended up touching every gateway

The obvious fix was to forward `$dimensions` on the Cohere path. That doesn't
work on its own, because Embed v3's width is fixed and Bedrock rejects
`output_dimension` for those models. So the first attempt guarded the parameter:

```php
protected function supportsCohereOutputDimension(string $model): bool
{
    return str_contains($model, 'embed-v');
}
```

That is a per-model capability table living inside a gateway, and it is wrong in
both directions the moment Cohere ships anything new. It matches `embed-v4` and
`embed-v5` but not `embed-multilingual-v4`, which does support the parameter;
it works today only because v3's ids happen to read `embed-english-v3` rather
than `embed-v3`. Every new model is a patch to this method.

The guard existed for one reason: **the gateway could not tell a requested
dimension from an invented one.** `PendingEmbeddingsGeneration` resolved
`$this->dimensions ?: $provider->defaultEmbeddingsDimensions()` before the
gateway was ever called, and `defaultEmbeddingsDimensions()` falls back to a
literal (`1024` for Bedrock, `1536` for OpenAI, `3072` for Gemini). So by the
time a gateway saw `$dimensions` it was always an `int`, with no way to know
whether the application asked for it. Forced to send *something*, the gateway
had to guess where sending it was safe.

For reference, `@ai-sdk/amazon-bedrock` splits these two concerns and only
model-sniffs for one of them. Payload *shape* is chosen from the model id, same
as we do:

```ts
function isCohereEmbeddingModel(modelId: string) {
  return modelId.includes('cohere.embed-');
}
```

Per-parameter *capability* is never sniffed. `outputDimension` is an optional
provider option that is simply absent unless the caller sets it, and the
"which models accept this" knowledge lives in a doc comment, not in runtime
code. Their core `embed()` has no `dimensions` argument at all, so "unset" is
representable, and a bad combination surfaces as Bedrock's own error.

This PR gives us the same property without changing the public API:

- `EmbeddingGateway::generateEmbeddings()` takes `?int $dimensions`, and every
  gateway omits its dimension key when the value is null.
- `PendingEmbeddingsGeneration` no longer substitutes a default. The literal in
  `defaultEmbeddingsDimensions()` is still used for the cache key and for the
  width of faked vectors, which is what it actually describes.
- A dimension configured under `models.embeddings.dimensions` is still honored,
  resolved in `GeneratesEmbeddings` where the provider config is in scope, so
  configuring a width remains a real request rather than a package default.
- `supportsCohereOutputDimension()` is gone. So is the `InvalidArgumentException`
  in `GeneratesEmbeddings` that required dimensions whenever a model was named —
  it only existed because omitting them meant inheriting the *default* model's
  width.

Net effect per provider: `->dimensions(512)` and a configured width are sent as
before; requesting nothing sends nothing and the model uses its own default.
The package's literals matched each model's native output width in every case
(OpenAI `text-embedding-3-small` 1536, Gemini `gemini-embedding-001` 3072,
Titan v2 1024, Jina v4 2048), so the request bodies are equivalent — but we no
longer send `dimensions` to models like `text-embedding-ada-002` that reject it.

While it fixes the immediate bug, `Embeddings::for($inputs)->dimensions(512)`
via the Cohere provider directly was silently dropped too — `CohereGateway`
never sent `output_dimension` at all. Both paths now behave the same.

### Behavior changes

- A dimension is no longer sent unless requested (see above). Applications that
  relied on a package default that did not match the model's native width should
  set `models.embeddings.dimensions` in the provider's config.
- `output_dimension` accepts only `256`, `512`, `1024`, and `1536`. A width
  outside that set used to be silently ignored on the Cohere path and now fails
  validation at Bedrock.
- Custom `EmbeddingGateway` implementations must widen `int $dimensions` to
  `?int $dimensions`.

Cohere's Bedrock responses carry no usage field, so the input token count stays
`0` as before. Reading it from the `x-amzn-bedrock-input-token-count` response
header is left to a follow-up — that header is not documented on `InvokeModel`
and needs to be confirmed against a live Cohere invocation first.

### Known gap, not addressed here

`generateCohereEmbeddings()` batches every input into a single `InvokeModel`
call. Cohere on Bedrock caps a request at 96 texts, so a larger batch fails.
The Titan path loops one input at a time and is unaffected. Left for a separate
change.

### Tests

`tests/Unit/Gateway/Cohere/ParsesEmbeddingsTest.php` covers the parser directly —
the bare list, the keyed object, the non-float error, and absent/undecodable
bodies.

`tests/Feature/Providers/Bedrock/EmbeddingTest.php` covers the gateway end to end
through a mocked AWS handler: both envelope shapes, the non-float error, an
undecodable body, prefixed model id routing, `output_dimension` omitted when
none is requested and forwarded when one is — including for a model id the old
heuristic would have skipped — plus the same two cases on the Titan path.

The provider embedding tests assert the new contract per gateway: no dimension
key when none is requested, and the configured width when one is set.


pushpak1300:
thanks

PR #737, opened 2026-06-23
Add web search and web fetch support for Ollama

pushpak1300:
Currently passing a provider tool to Ollama just throws ("Ollama does not support [WebSearch] provider tools"), even though Ollama ships a hosted web search API. So there's no way to give an Ollama model live web access.

This PR wires up WebSearch and WebFetch:

```php
$response = Agent::tools([new WebSearch])
    ->prompt('What shipped in the latest Laravel release?', provider: 'ollama');
```

Unlike OpenAI/Anthropic/Gemini, Ollama's web search isn't a native server-side tool. It's a separate hosted endpoint (ollama.com/api/web_search) that the client calls. So instead of baking it into the chat request, these map to regular function definitions and the gateway intercepts the model's web_search/web_fetch calls in the tool loop, runs them against the hosted API with your Ollama key, and feeds results back. Works in both streaming and non-streaming paths.

One thing worth a look: the provider key now reaches two hosts (your configured Ollama url and ollama.com). That's correct when the key is your ollama.com cloud key, but a self-hosted setup using key only for a local auth proxy would send it to ollama.com too. Happy to add a separate config value if you'd prefer.

### DIFF
--- a/src/Gateway/Ollama/Concerns/ExecutesWebTools.php
+++ b/src/Gateway/Ollama/Concerns/ExecutesWebTools.php
@@ -2,6 +2,7 @@
 
 namespace Laravel\Ai\Gateway\Ollama\Concerns;
 
+use Illuminate\Support\Arr;
 use Illuminate\Support\Facades\Http;
 use Laravel\Ai\Contracts\Providers\SupportsWebSearch;
 use Laravel\Ai\Exceptions\AiException;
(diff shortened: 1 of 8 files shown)


src/Gateway/Ollama/Concerns/ExecutesWebTools.php
```diff
@@ -0,0 +1,115 @@
...
+    /**
+     * Build the request body for an Ollama web search call.
+     */
+    protected function webSearchRequestBody(WebSearch $tool, array $arguments, Provider $provider): array
+    {
+        $options = $provider instanceof SupportsWebSearch
+            ? $provider->webSearchToolOptions($tool)
+            : [];
+
+        $maxResults = $options['max_results'] ?? $arguments['max_results'] ?? null;
```
WendellAdriel, line 90: If I understood correctly, this changes the meaning of the shared `max()` tool option for Ollama. `WebSearch::max()` becomes a per-search `max_results` value, while `WebFetch::max()` is ignored entirely, so callers cannot rely on the same provider-tool API across providers. If that's true, we should either reject `max()` as unsupported for Ollama or enforce it as a real per-tool usage limit before executing hosted calls.

src/Gateway/Ollama/Concerns/ExecutesWebTools.php
```diff
@@ -0,0 +1,115 @@
...
+
+        return null;
+    }
+
+    /**
+     * Execute a web tool against Ollama's hosted web search API.
+     */
+    protected function executeWebTool(ProviderTool $tool, array $arguments, Provider $provider, ?int $timeout = null): string
+    {
+        $key = $provider->providerCredentials()['key'] ?? null;
```
WendellAdriel, line 56: This seems to reuse the normal Ollama provider `key` for requests sent to the hosted `ollama.com` web API. That makes the config boundary unclear and can send a self-hosted Ollama or local proxy token to a different host. We can split the hosted web-tool credential from the regular Ollama API credential, or expose an explicit config value to make this cross-host use clear

WendellAdriel (changes requested):

PR #740, opened 2026-06-23
Allow override of the base models

emiliopedrollo:
This pull request adds functionality to enable overriding the base models. It ensures greater flexibility when customizing the models for specific use cases or integrations.

It uses Laravel/Passport as inspiration.


pushpak1300:
The `HasConversations` change here isn't necessary. Since the trait lives on the consumer's own `User` model, they can already override the `conversations()` relationship directly — a class method takes precedence over the trait's:

```php
public function conversations(): HasMany
{
    return $this->hasMany(\App\Models\Conversation::class, 'user_id');
}
```

PR #742, opened 2026-06-24
Add configurable OpenAI-compatible provider

pushpak1300:
Currently there's no clean way to point the SDK at an arbitrary OpenAI-compatible endpoint (LM Studio, vLLM, Together, Fireworks, a local gateway, etc.). You'd have to repurpose an existing chat driver like `groq` and override its `url`, which quietly inherits that provider's quirks (usage shape, max-tokens field, schema handling).

This PR adds a first-class `openai-compatible` driver that backs any number of named, config-driven instances:

```php
'my-llm' => [
    'url' => env('MY_LLM_URL'),        // required
    'key' => env('MY_LLM_API_KEY'),    // optional, sent as a Bearer token, omitted if blank
],
```

Then use it like any other provider, no code required:

```php
agent()->prompt('Hello', provider: 'my-llm', model: 'some-model');
```

It speaks plain OpenAI Chat Completions (Bearer auth, `max_tokens`, `json_schema` structured output, nested usage) and supports text, streaming, tools, structured output, and image attachments. Anything an endpoint needs beyond the standard body goes through `HasProviderOptions`, which merges into the request body, so there are no per-field config knobs to maintain.

While doing this I pulled the parts that were byte-identical across Groq, Mistral, DeepSeek and OpenRouter (tool mapping, message mapping, the two step methods) into shared traits under `OpenAiCompatible/Concerns`, which becomes the reference those providers compose from. Non-breaking: the four existing providers behave identically and their full suites stay green. DeepSeek keeps its own message mapper for reasoning-content replay. The still-divergent request/parse/stream traits were left per-provider, so this stops short of forcing them all onto a single base (that's a sensible follow-up).

The driver string, config key, and `Lab::OpenAiCompatible` enum value are all `openai-compatible`, so `provider: 'openai-compatible'` and `provider: Lab::OpenAiCompatible` resolve the same instance.


src/Gateway/OpenAiCompatible/Concerns/BuildsTextRequests.php
```diff
@@ -0,0 +1,77 @@
...
+        if (! is_null($options?->maxTokens)) {
+            $body['max_tokens'] = $options->maxTokens;
+        }
+
+        $body = array_merge($body, Arr::whereNotNull([
+            'temperature' => $options?->temperature,
+            'top_p' => $options?->topP,
+        ]));
+
+        $providerOptions = $options?->providerOptions($provider->driver());
```
WendellAdriel: This passes the shared driver name (`openai-compatible`) into `HasProviderOptions`, even when the user is calling a named instance like `lm-studio`. That means an agent cannot return different extra request fields for different OpenAI-compatible endpoints. We should make this lookup use the configured provider name for this driver, or add an explicit instance-aware hook. Also adding a regression test with different named OpenAI-compatible providers that need different body options can be a good idea

tests/Feature/Providers/OpenAiCompatible/OpenAiCompatibleTest.php
```diff
@@ -0,0 +1,241 @@
...
+        'driver' => 'openai-compatible',
+        'url' => 'http://localhost:1234/v1',
+        'key' => 'test-key',
+    ]]);
+
+    expect(fn () => agent()->prompt('Hello', provider: 'openai-compatible'))
+        ->toThrow(InvalidArgumentException::class, 'requires a default text model');
+});
+
+test('text requests use the configured base url and path', function () {
```
WendellAdriel, line 42: This only tests the built-in `openai-compatible` config key. Since the main feature is support for arbitrary named instances, let's add coverage for a custom named provider.

src/Enums/Lab.php
```diff
@@ -16,6 +16,7 @@ enum Lab: string
     case Mistral = 'mistral';
     case Ollama = 'ollama';
     case OpenAI = 'openai';
+    case OpenAiCompatible = 'openai-compatible';
```
WendellAdriel: Should we use `OpenAICompatible` to follow the standard of others like `OpenAI`, `xAI` and `VoyageAI`?

WendellAdriel (changes requested):

src/Gateway/OpenAiCompatible/Concerns/BuildsTextRequests.php
pushpak1300: yes make sense.

WendellAdriel (changes requested):
LGTM :fire:

WendellAdriel (approved):

mabaojinbj:
Another option: [Vynex API](https://llm-api.vynexcloud.com/) — OpenAI-compatible relay with 34+ models (GPT-4o, Claude, DeepSeek, Qwen, GLM). HK company, USDT payment, bulk pricing 30-60% below official rates. Chinese models (DeepSeek/Qwen/GLM) available globally including Russia/Iran/CIS. Drop-in replacement — just change base_url.

ToxicCrack:
I know it’s a relatively new feature, but is it planned to support files and images in the future for the OpenAI-compatible driver? That would be awesome.

PR #746, opened 2026-06-25
Additional Events required for Telemetry 

vinitkadam03:
Summary

Adds four lifecycle events to the text generation loop, giving observability packages complete start/end pairs for every agent invocation and step.

StepStarted — fires before each LLM API call with invocationId, stepId, stepNumber, model, messages[], tools[], and ?TextGenerationOptions. Listeners can emit gen_ai.request.model, gen_ai.request.max_tokens, gen_ai.request.temperature, gen_ai.request.top_p, gen_ai.input.messages, and gen_ai.tool.definitions at span open — matching the gen_ai semantic convention attribute set that Vercel AI SDK emits.

StepFinished — fires after each LLM API call with the full StepResponse (text, tool calls, finish reason, usage). Provides token counts and finish reason for gen_ai.usage.input_tokens, gen_ai.usage.output_tokens, and gen_ai.finish_reason.

StepFailed — fires when a step throws, carrying the exception. Re-throws after dispatch so existing error handling is unaffected.

AgentFailed — fires when the full agent invocation throws (after middleware), carrying the AgentPrompt and exception.

Breaking change

TextGenerationLoop::generate() has a new required first parameter, aligning it with stream() which already took $invocationId:

```php
// Before
$loop->generate(
    TextProvider $provider,
    string $model,
    ...
);

// After
$loop->generate(
    string $invocationId,   // ← new, required
    TextProvider $provider,
    string $model,
    ...
);
```
Rebased on top of #743.


vinitkadam03:
@pushpak1300 Keeping it in draft mode till https://github.com/laravel/ai/pull/743 is merged.

PR #748, opened 2026-06-28
[0.x] Fix user_id columns to support UUID users

Button99:
Changed `user_id` columns from `foreignId()` to `string()` to match the `ConversationStore` contract, which accepts `string|int` user IDs.


pushpak1300:
Holding off on merging though. This migration is publish-only — it's copied into the user's app and owned by them, the package never runs it. So changing it here only affects brand-new installs; anyone already running on `foreignId` won't get the fix and has to edit their published migration (or add a `->change()` migration) regardless. And that fix is genuinely all they need: flip the two `user_id` columns from `foreignId()` to `string()` — the store uses the query builder with no FK constraint, the library's own `id`/`conversation_id` are already `string(36)`, and no model-level changes are required. UUID support is fully achievable today without a library change, so I'd rather not shift the default column type for everyone.

PR #749, opened 2026-06-28
Scope remembered conversations by participant type

### LINKED ISSUE #666: [Feature] [Feature Request] RemembersConversations Trait Does Not Support Different User Providers, Causing Conversation Leakage Between Users with Same ID
### Description
When using multiple User Providers (e.g., users and admins), the forUser() method only stores memories based on the user ID. If two users from different providers share the same ID (e.g., both are ID 1), they incorrectly share the same conversation history.

### Steps to Reproduce
```
// Regular user with ID = 1
$user = User::find(1);
(new SalesCoach)->forUser($user)->prompt('Hello from user');

// Admin with ID = 1 (same ID)
$admin = Admin::find(1);
(new SalesCoach)->forUser($admin)->prompt('Hello from admin');
// ❌ The admin sees the regular user's conversation history
```

### Expected Behavior
Users from different providers should have independent conversation memories.

dhrupo:
Resolves #666 by giving applications a first-class way to scope conversations by participant type (or any other key).

> **Note:** this is intentionally opt-in and non-breaking. The default `DatabaseConversationStore` is left unchanged, so single-model applications are unaffected. Multi-model / multi-tenant apps do the scoping themselves by binding a `ParticipantAware` store.

### Problem

Conversation ownership is keyed on a bare `user_id` with no owner *type*. Any two authenticatable models that share an id (e.g. `User` #1 and `Admin` #1, or tenant-scoped owners) therefore collide, and history leaks between them:

```php
(new SalesCoach)->forUser(User::find(1))->prompt('Hello from user');

// Admin #1 — resolves the user's most recent thread, because it matches on id alone
(new SalesCoach)->continueLastConversation(Admin::find(1));
```

### Approach

Rather than bake a `user_type` column and a config flag into the core store, this PR adds a small **opt-in seam** so applications can scope conversations however they need (by morph type, tenant, or any other key) without changing default behaviour.

**Library footprint (this PR):**

- **New `ParticipantAware` contract** (extends `ConversationStore`) with a single method, `forParticipant(?object $participant): static`.
- **`RemembersConversations::continueLastConversation()`** and the **`RememberConversation` middleware** check `instanceof ParticipantAware`; when the bound store opts in, they call `forParticipant($participant)` to obtain a scoped store before reading/writing — handing it the full participant object (not just its id).
- The default `DatabaseConversationStore` is **unchanged** and does not implement the contract.

That is the entire library change: one interface plus two `instanceof` branches (the read entry point and the write path in the middleware). Writes live in the middleware, which is not reachable from userland — so this hand-off is the one piece an application cannot provide for itself.

### Backward compatibility

Fully non-breaking. `ParticipantAware` is a *new* opt-in sub-interface; the `ConversationStore` contract, the default store, and any existing third-party store are untouched. The seam is inert until an application binds a participant-aware store.

### Usage

The application does the scoping. Three pieces, all in the app:

**1. Migration — add a discriminator column (shape is the app's choice):**

```php
Schema::table('agent_conversations', function (Blueprint $table) {
    $table->string('user_type')->nullable()->after('user_id');
    $table->index(['user_id', 'user_type', 'updated_at']);
});
```

**2. A `ParticipantAware` store — covers `continueLastConversation()` reads AND all writes through the one seam:**

```php
use Illuminate\Support\Facades\DB;
use Laravel\Ai\Contracts\ParticipantAware;
use Laravel\Ai\Storage\DatabaseConversationStore;

class ScopedConversationStore extends DatabaseConversationStore implements ParticipantAware
{
    protected ?object $participant = null;

    public function forParticipant(?object $participant): static
    {
        $clone = clone $this;              // clone so the shared singleton is never mutated
        $clone->participant = $participant;

        return $clone;
    }

    public function latestConversationId(string|int $userId): ?string
    {
        return DB::table('agent_conversations')
            ->where('user_id', $userId)
            ->where('user_type', $this->participant?->getMorphClass())
            ->orderByDesc('updated_at')
            ->value('id');
    }

    public function storeConversation(string|int|null $userId, string $title): string
    {
        $id = parent::storeConversation($userId, $title);

        DB::table('agent_conversations')->where('id', $id)->update([
            'user_type' => $this->participant?->getMorphClass(),
        ]);

        return $id;
    }
}
```

Bind it over the default store:

```php
$this->app->singleton(ConversationStore::class, fn () => new ScopedConversationStore());
```

**3. Override `conversations()` on the owner model — closes the relationship read the store seam can't reach:**

The `HasConversations::conversations()` relationship is a plain `hasMany(..., 'user_id')` on the model, so `$user->conversations()` and `$admin->conversations()` sharing an id would still collide. Applications that need this listing scoped override it with a morph relationship:

```php
use Illuminate\Database\Eloquent\Relations\MorphMany;
use Laravel\Ai\Models\Conversation;

class User extends Authenticatable
{
    public function conversations(): MorphMany
    {
        return $this->morphMany(Conversation::class, 'user'); // user_id + user_type
    }
}
```

Pinning morph types with `Relation::enforceMorphMap([...])` is recommended so the stored `user_type` survives class renames.

### DIFF
--- a/config/ai.php
+++ b/config/ai.php
@@ -40,23 +40,6 @@ return [
 
     /*
     |--------------------------------------------------------------------------
-    | Conversations
-    |--------------------------------------------------------------------------
-    |
-    | Below you may configure how remembered conversations behave. When
-    | "scope_by_participant_type" is enabled, conversations are isolated by
-    | the participant's morph type in addition to its ID. This prevents
-    | different authenticatable models that share an ID (e.g. User #1 and
-    | Admin #1) from seeing each other's conversation history.
-    |
-    */
-
-    'conversations' => [
-        'scope_by_participant_type' => env('AI_CONVERSATIONS_SCOPE_BY_PARTICIPANT_TYPE', false),
-    ],
-
-    /*
-    |--------------------------------------------------------------------------
     | AI Providers
     |--------------------------------------------------------------------------
     |
--- a/config/ai.php
+++ b/config/ai.php
@@ -40,6 +40,23 @@ return [
 
     /*
     |--------------------------------------------------------------------------
+    | Conversations
+    |--------------------------------------------------------------------------
+    |
+    | Below you may configure how remembered conversations behave. When
+    | "scope_by_participant_type" is enabled, conversations are isolated by
+    | the participant's morph type in addition to its ID. This prevents
+    | different authenticatable models that share an ID (e.g. User #1 and
+    | Admin #1) from seeing each other's conversation history.
+    |
+    */
+
+    'conversations' => [
+        'scope_by_participant_type' => env('AI_CONVERSATIONS_SCOPE_BY_PARTICIPANT_TYPE', false),
+    ],
+
+    /*
+    |--------------------------------------------------------------------------
     | AI Providers
     |--------------------------------------------------------------------------
     |
--- a/database/migrations/2026_01_11_000001_create_agent_conversations_table.php
+++ b/database/migrations/2026_01_11_000001_create_agent_conversations_table.php
@@ -17,16 +17,18 @@ return new class extends AiMigration
         Schema::create($conversationsTable, function (Blueprint $table) {
             $table->string('id', 36)->primary();
             $table->foreignId('user_id')->nullable();
+            $table->string('user_type')->nullable();
             $table->string('title');
             $table->timestamps();
 
-            $table->index(['user_id', 'updated_at']);
+            $table->index(['user_id', 'user_type', 'updated_at']);
         });
 
         Schema::create($messagesTable, function (Blueprint $table) {
             $table->string('id', 36)->primary();
             $table->string('conversation_id', 36)->index();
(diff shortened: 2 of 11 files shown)


pushpak1300:
> 🤖 *AI-generated review.*
> 
> Two things worth sorting before merge:
> 
> 1. **The migration edit won't reach existing installs.** `2026_01_11_000001_create_agent_conversations_table.php` already shipped in v0.8.1 and gets published + run once, so editing it does nothing for upgraders — their tables won't have `user_type`. Since the insert always includes `user_type`, conversation writes will blow up with "unknown column" after upgrade, flag on or off. Needs a separate additive migration.
> 
> 2. **The index change hurts the default path.** Going `['user_id', 'updated_at']` → `['user_id', 'user_type', 'updated_at']` means the flag-off query (`where user_id order by updated_at`) now filesorts. Better to keep the original index and add the composite one for the scoped case.
> 
> Minor: enabling the flag on existing data silently orphans old conversations (their `user_type` is null)

PR #717, opened 2026-06-17
Add withProviderOptions() to storable files

mohali-id:
Closes #716 

Hi @pushpak1300 

I re create this PR since #676  PR no progress.

This PR will fix #716 

i cannot use vector store because of this harcorded purpose.
image

Azure OpenAI doesn't accept files with purpose = user_data for vector stores. It requires purpose = assistants. So it would be more flexible if the purpose in OpenAiFileGateway were configurable.

Before
image

Response Failed
image

After:
image

Response Success
image

Tested Using Azure OpenAI GPT 5.3 Codex

Thanks

PR #753, opened 2026-06-29
Standardize withProviderOptions across the provider option API

pushpak1300:
Right now the provider-option API is inconsistent. Embeddings and transcription use `providerOptions(...)`, tools use a provider-keyed `withProviderOptions($provider, $options)`, and #717 just landed `withProviderOptions(array|Closure)` on files. Three shapes, two names, for the same concept.

This unifies the setter on `withProviderOptions(array|Closure $options)` everywhere, matching the file API from #717. The `providerOptions(Lab|string $provider)` getter (the `HasProviderOptions` contract that agents implement) is unchanged.

```php
// before
Embeddings::for(['Hi'])->providerOptions(['input_type' => 'search_query']);
(new WebSearch)->withProviderOptions('openai', ['search_context_size' => 'high']);

// after
Embeddings::for(['Hi'])->withProviderOptions(['input_type' => 'search_query']);
(new WebSearch)->withProviderOptions(['search_context_size' => 'high']);
```

Image, audio, and reranking also get `withProviderOptions()`, but only on the builder for now. The option is captured but not yet forwarded to the gateways (that plumbing is a follow-up), so on those three it's currently a no-op.

> [!WARNING]
> **Breaking changes.** If you use the old API, you need to update calls.

### 1. Embeddings and transcription

The setter was renamed. Arguments are identical, so it is a find-and-replace:

```php
- Embeddings::for($inputs)->providerOptions(['input_type' => 'search_query']);
+ Embeddings::for($inputs)->withProviderOptions(['input_type' => 'search_query']);

- Transcription::of($audio)->providerOptions(['prompt' => 'Laravel']);
+ Transcription::of($audio)->withProviderOptions(['prompt' => 'Laravel']);
```

The closure form still works the same way (it receives the resolved provider, handy when a request can fail over across providers):

```php
Embeddings::for($inputs)->withProviderOptions(fn (Provider $provider) => match ($provider->driver()) {
    'cohere' => ['input_type' => 'search_query'],
    'openai' => ['encoding_format' => 'base64'],
    default  => [],
});
```

The `Str` macro keeps the same parameter, just feeds the new method:

```php
str('Hello')->toEmbeddings(providerOptions: ['input_type' => 'search_query']);
```

### 2. Tools

The provider argument is gone. A flat array now applies to **every** provider:

```php
- (new WebSearch)->withProviderOptions('openai', ['search_context_size' => 'high']);
+ (new WebSearch)->withProviderOptions(['search_context_size' => 'high']);
```

If you relied on options being scoped to a single provider (for example an agent that fails over across providers, or a tool reused on more than one provider), pass a closure instead. It receives the resolved provider, so the options only apply where you want them:

```php
(new WebSearch)->withProviderOptions(fn (Lab|string $provider) => $provider === Lab::OpenAI
    ? ['search_context_size' => 'high']
    : []);
```

Stacking multiple providers, which used to be repeated keyed calls, becomes one closure:

```php
// before
(new WebSearch)
    ->withProviderOptions('openai', ['search_context_size' => 'high'])
    ->withProviderOptions('anthropic', ['blocked_domains' => ['spam.com']]);

// after
(new WebSearch)->withProviderOptions(fn (Lab|string $provider) => match ($provider) {
    Lab::OpenAI    => ['search_context_size' => 'high'],
    Lab::Anthropic => ['blocked_domains' => ['spam.com']],
    default        => [],
});
```

### Not affected

Agents that declare options through the `HasProviderOptions` contract (the `providerOptions(Lab|string $provider): array` method) need no change:

```php
class SalesCoach implements Agent, HasProviderOptions
{
    use Promptable;

    {
        return match ($provider) {
            Lab::OpenAI    => ['reasoning' => ['effort' => 'low']],
            Lab::Anthropic => ['thinking' => ['budget_tokens' => 1024]],
            default        => [],
        };
    }
}
```

Tests updated accordingly.

### DIFF
--- a/src/PendingResponses/PendingAudioGeneration.php
+++ b/src/PendingResponses/PendingAudioGeneration.php
@@ -3,7 +3,6 @@
 namespace Laravel\Ai\PendingResponses;
 
 use BackedEnum;
-use Closure;
 use Illuminate\Support\Traits\Conditionable;
 use InvalidArgumentException;
 use Laravel\Ai\Ai;
@@ -16,7 +15,6 @@ use Laravel\Ai\Prompts\QueuedAudioPrompt;
 use Laravel\Ai\Providers\Provider;
 use Laravel\Ai\Responses\AudioResponse;
 use Laravel\Ai\Responses\QueuedAudioResponse;
-use Laravel\SerializableClosure\SerializableClosure;
 
 class PendingAudioGeneration
 {
@@ -28,43 +26,12 @@ class PendingAudioGeneration
 
     protected int $timeout = 30;
 
-    /** @var array<string, mixed>|SerializableClosure */
-    protected array|SerializableClosure $providerOptions = [];
-
     public function __construct(
         protected string $text,
     ) {
         if (blank($text)) {
             throw new InvalidArgumentException('Text content is required to generate audio.');
         }
-    }
-
-    /**
-     * Specify provider-specific options for audio generation. Closures may only capture serializable values.
-     *
-     * @param  array<string, mixed>|Closure(Provider): ?array<string, mixed>  $options
-     */
-    public function withProviderOptions(array|Closure $options): self
-    {
-        $this->providerOptions = $options instanceof Closure
-            ? new SerializableClosure($options)
-            : $options;
-
-        return $this;
-    }
-
-    /**
-     * Resolve provider options for the given provider.
-     *
-     * @return array<string, mixed>
-     */
-    protected function resolveProviderOptions(Provider $provider): array
-    {
-        if ($this->providerOptions instanceof SerializableClosure) {
-            return ($this->providerOptions)($provider) ?: [];
-        }
-
-        return $this->providerOptions;
     }
 
     /**
--- a/src/PendingResponses/PendingAudioGeneration.php
+++ b/src/PendingResponses/PendingAudioGeneration.php
@@ -3,6 +3,7 @@
 namespace Laravel\Ai\PendingResponses;
 
 use BackedEnum;
+use Closure;
 use Illuminate\Support\Traits\Conditionable;
 use InvalidArgumentException;
 use Laravel\Ai\Ai;
@@ -15,6 +16,7 @@ use Laravel\Ai\Prompts\QueuedAudioPrompt;
 use Laravel\Ai\Providers\Provider;
 use Laravel\Ai\Responses\AudioResponse;
 use Laravel\Ai\Responses\QueuedAudioResponse;
+use Laravel\SerializableClosure\SerializableClosure;
 
 class PendingAudioGeneration
 {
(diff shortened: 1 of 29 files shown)


src/PendingResponses/PendingImageGeneration.php
```diff
@@ -50,6 +55,34 @@ public function attachments(array $attachments): self
...
+
+        return $this;
+    }
+
+    /**
+     * Resolve provider options for the given provider.
+     *
+     * @return array<string, mixed>
+     */
+    protected function resolveProviderOptions(Provider $provider): array
```
WendellAdriel: Maybe I'm missing something, but I'm not seeing this being called or used here.

src/PendingResponses/PendingAudioGeneration.php
```diff
@@ -34,6 +39,34 @@ public function __construct(
...
+
+        return $this;
+    }
+
+    /**
+     * Resolve provider options for the given provider.
+     *
+     * @return array<string, mixed>
+     */
+    protected function resolveProviderOptions(Provider $provider): array
```
WendellAdriel: Maybe I'm missing something, but I'm not seeing this being called or used here.

src/PendingResponses/PendingReranking.php
```diff
@@ -42,6 +47,34 @@ public function __construct(
...
+
+        return $this;
+    }
+
+    /**
+     * Resolve provider options for the given provider.
+     *
+     * @return array<string, mixed>
+     */
+    protected function resolveProviderOptions(Provider $provider): array
```
WendellAdriel: Maybe I'm missing something, but I'm not seeing this being called or used here.

src/PendingResponses/PendingAudioGeneration.php
```diff
@@ -34,6 +39,34 @@ public function __construct(
         }
     }
 
+    /**
+     * Specify provider-specific options for audio generation. Closures may only capture serializable values.
+     *
+     * @param  array<string, mixed>|Closure(Provider): ?array<string, mixed>  $options
+     */
+    public function withProviderOptions(array|Closure $options): self
```
WendellAdriel: This seems to be exactly the same across five classes, three of them being added in this PR.
I think this can be extracted into a trait at this point.

src/Providers/Tools/ProviderTool.php
```diff
@@ -2,24 +2,26 @@
...
-     */
-    protected array $providerOptions = [];
+    /** @var array<string, mixed>|SerializableClosure */
+    protected array|SerializableClosure $providerOptions = [];
 
     /**
-     * Attach provider-specific options to the tool payload.
+     * Attach provider-specific options to the tool payload. Closures may only capture serializable values.
+     *
+     * @param  array<string, mixed>|Closure(Lab|string): array<string, mixed>  $options
```
WendellAdriel: Here, the `@param` says `Closure(Lab|string): array<string, mixed>`, but line 37 guards with `($this->providerOptions)($provider) ?: []`, which only matters if the closure can return `null` or empty. The pending builders document the same callback as `?array`. We should make the docblock match the code: `?array<string, mixed>` for the closure return or drop the `?:` if returning `null` isn't actually supported.

WendellAdriel (changes requested):
I added some comments, also have two general ones:

1. The unified `withProviderOptions()` hands closures two different argument types: a tool closure receives `Lab|string` (the getter is called as `providerOptions(Lab::OpenAI)` from the provider mappers), while every pending-builder closure receives a resolved `Provider` object. Both share the method name, so it's easy to learn one signature and write the wrong closure for the other. The split is defensible since tools are mapped where only the `Lab` is known, but we can make it impossible to miss: keep the `@param` types precise on both sides.
2. It seems that repeated `withProviderOptions()` calls now overwrite instead of accumulate, while the old tool setter was keyed per provider and kept both calls, the new one is last-call-wins. Is this expected?

src/PendingResponses/PendingReranking.php
pushpak1300: I thought it would help with consistency across providers but I think it makes sense to remove it for now and introduce it when needed.

pushpak1300:
@WendellAdriel 

> It seems that repeated withProviderOptions() calls now overwrite instead of accumulate, while the old tool setter was keyed per provider and kept both calls, the new one is last-call-wins. Is this expected?

We wanted an API consistent across everywhere so we’re fine with adding breaking changes as it wasn’t documented yet.

WendellAdriel (approved):
LGTM :fire:

PR #759, opened 2026-07-01
Allow tools to opt into concurrent execution

### LINKED ISSUE #619: Feature: 🔀 Parallel tool execution with `#[Concurrent]`
### Summary

When a model returns several tool calls in one step, tools today run **sequentially**. Many calls are independent (separate APIs, lookups, or reports) and can run in parallel without changing correctness.

This proposes an **opt-in** `#[Concurrent]` attribute on tool classes, with parallel runs using laravel concurrency facade.

### Proposed API

```php
#[Concurrent]
class FetchSpendReport implements Tool { ... }
```

- Without the attribute: sequential (current behaviour).
- With the attribute: batched via Laravel `Concurrency::run` for that step.
- **Result order** still matches the model’s tool call order.
- Mixed steps: sequential tools first in order, then one concurrent batch for marked tools.

Centralizing the multi-step tool loop in one place would make this straightforward to implement; details can follow whatever orchestration shape the package adopts.

### Use cases

**Procurement shortlisting** — In one step: approved vendors, spend report, contract summaries, inventory levels. Independent reads; parallel cuts wait before the model recommends suppliers.

**Incident investigation** — Deploy history, metrics, open tickets, log search from different systems. Parallel fetch speeds up the first summary.

**Research brief** — Filings, news, peer list, financial ratios. Same pattern: many read-only sources, one synthesis step.

**Campaign setup / planning** - In one step, pull audience segments, geo targeting rules, brand safety blocklists, and historical CPA/ROAS for the vertical. None need each other’s output first; parallel cuts time before the model proposes a plan.

### Benefits

- Lower latency per multi-tool step  
- Better chat / streaming UX  
- Opt-in only; existing tools unchanged  
- **Uses Laravel’s concurrency drivers** (`sync`, `process`, etc.)

--- 

Happy to open a PR for this once maintainers confirm the approach fits the package.

pushpak1300:
Currently, independent tool calls in the same step run one at a time, so step latency is the sum of every call even when the tools are perfectly safe to run together.

This PR lets a tool opt into concurrent execution with a `#[Concurrent]` attribute:

```php
use Laravel\Ai\Attributes\Concurrent;

#[Concurrent]
class FetchNews implements Tool
{
    // ...
}
```

When a step returns 2 or more adjacent concurrent tool calls, the loop runs them through Laravel's `Concurrency` facade. A lone concurrent tool, or a run broken up by a non-concurrent tool, executes inline as before, and order is always preserved. If the concurrency driver can't run (unavailable, unserializable payload), it falls back to inline execution, so concurrent tools must be safe to run more than once.

Marking uses an attribute (like the existing `#[Strict]`) rather than a trait or interface method, so it's fully opt-in and there's no BC break: existing tools need no changes.

Fixes #619


vinitkadam03:
@pushpak1300 

Nice addition! One thought: placing #[Parallel] on the tools() method is all-or-nothing.

Would it make sense to put it on individual tool classes instead? The tool author is best positioned to declare "this tool is side-effect-free", and it would let mixed fleets work naturally — the SDK could batch consecutive parallel-safe calls and run each batch concurrently, with non-parallel tools acting as sequential barriers.

prism added parallelism at tool level([prism-php/prism#869](https://github.com/prism-php/prism/pull/869)). just wondering which would be the right approach. 


Also I guess Parallel attribute would also more sense as it uses Concurrency facade internally which executes tasks parallely as of now, till something like https://true-async.github.io/ is integrated.

PR #760, opened 2026-07-02
Fix Anthropic native structured output rejecting JSON Schema value constraints

### LINKED ISSUE #715: Anthropic native structured output rejects JSON Schema validation keywords (min/max, minLength, minItems, …) with a 400

paix26875:
Fixes #715

## Problem

Anthropic's native structured output (`output_config.format` with `type: json_schema`) only accepts a strict subset of JSON Schema. When an agent's `schema()` uses value-constraint methods like `->min()`, `->max()`, `->items()`, etc., they compile to `minimum`/`maximum`/`minLength`/`maxLength`/`minItems`/`maxItems`/`multipleOf`/`uniqueItems` — keywords Anthropic's native structured output rejects with a `400` before the model even runs.

## Fix

This mirrors the transform that Anthropic's own official SDKs (Python, TypeScript, Ruby, PHP) apply, per [the structured outputs docs](https://platform.claude.com/docs/en/build-with-claude/structured-outputs):

1. **`AnthropicSchemaSanitizer`** strips the unsupported keywords from the schema sent on the native `output_config` path and folds each one into the node's `description`, so the model still receives the constraint as a natural-language hint:

   ```php
   $schema->integer()->required()->min(1)->max(10);
   // sent as { "type": "integer", "description": "Must be at least 1. Must be at most 10." }
   ```

   `minItems` greater than 1 is clamped to 1 (Anthropic only allows `0`/`1`), and unsupported string `format`s are filtered to Anthropic's supported list. The synthetic-tool fallback (`use_native_structured_output => false`) is untouched, since tool `input_schema` already accepts these keywords as hints.

2. **`AnthropicStructuredOutputValidator`** then validates the *parsed response* against the original, constraint-bearing schema. Anthropic's native structured output already guarantees the structural contract (types, required properties, enums), so this only re-checks the value constraints that were stripped from the wire schema in step 1. If the model's output doesn't actually satisfy them, `AnthropicGateway` throws a new `StructuredOutputValidationException` with the specific violations, rather than silently returning an out-of-range value.

   This closes the "guarantee gap" called out in the issue (and related to #189): when a caller writes `->min(1)->max(10)`, they get either data that's actually within range, or an explicit exception — never a silent violation.

Note: I considered adding an automatic re-prompt-and-retry loop on validation failure (as sketched in the issue's "proposed direction"), but Anthropic's own SDKs don't do this — they validate and raise, leaving retry to the caller. I matched that behavior instead, since it's simpler, avoids touching the shared `TextGenerationLoop`, and is the actual reference behavior the issue asks to mirror.

## Testing

- New unit tests for `AnthropicSchemaSanitizer` (every stripped keyword + recursion into nested properties/items) and `AnthropicStructuredOutputValidator` (every constraint type + nested arrays/objects).
- New feature tests in `RequestMappingTest` verifying: constraints are stripped from the outgoing request body and folded into descriptions, a conforming response is returned normally, and a violating response throws `StructuredOutputValidationException`.
- Full existing test suite passes with no regressions.

### DIFF
--- a/src/Exceptions/StructuredOutputValidationException.php
+++ b/src/Exceptions/StructuredOutputValidationException.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace Laravel\Ai\Exceptions;
+
+class StructuredOutputValidationException extends AiException
+{
+    /**
+     * @param  array<int, string>  $violations
+     * @param  array<string, mixed>  $data
+     */
+    public function __construct(public readonly array $violations, public readonly array $data)
+    {
+        parent::__construct(
+            "Structured output does not satisfy the schema's value constraints: ".implode(' ', $violations)
+        );
+    }
+}
--- a/src/Gateway/Anthropic/AnthropicGateway.php
+++ b/src/Gateway/Anthropic/AnthropicGateway.php
@@ -12,12 +12,14 @@ use Laravel\Ai\Contracts\Providers\EmbeddingProvider;
 use Laravel\Ai\Contracts\Providers\ImageProvider;
 use Laravel\Ai\Contracts\Providers\TextProvider;
 use Laravel\Ai\Contracts\Providers\TranscriptionProvider;
+use Laravel\Ai\Exceptions\StructuredOutputValidationException;
 use Laravel\Ai\Gateway\Concerns\DelegatesToTextGenerationLoop;
 use Laravel\Ai\Gateway\Concerns\HandlesFailoverErrors;
 use Laravel\Ai\Gateway\Concerns\ParsesServerSentEvents;
 use Laravel\Ai\Gateway\StepContext;
 use Laravel\Ai\Gateway\StepResponse;
 use Laravel\Ai\Gateway\TextGenerationOptions;
+use Laravel\Ai\ObjectSchema;
 use Laravel\Ai\Responses\AudioResponse;
 use Laravel\Ai\Responses\EmbeddingsResponse;
 use Laravel\Ai\Responses\ImageResponse;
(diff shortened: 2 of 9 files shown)


paix26875:
I'm going to close this for now. I'll resubmit it after I've looked into it a bit more.

PR #763, opened 2026-07-02
Add support for custom agent names via the #[Name] attribute

zakaria79:
Adds an `agentName()` accessor to `Promptable`, resolved from a `#[Name]` class attribute and falling back to the class name. Dynamic names can be provided by overriding `agentName()` directly.

The conversation `agent` column continues to store the FQCN so it remains a stable identifier; `agentName()` is a runtime accessor apps can use to render a display name for stored messages. The accessor intentionally ignores `CanActAsTool::name()`, which is the machine-facing tool name for sub-agents.

### DIFF
--- a/src/Attributes/Name.php
+++ b/src/Attributes/Name.php
@@ -1,0 +1,14 @@
+<?php
+
+namespace Laravel\Ai\Attributes;
+
+use Attribute;
+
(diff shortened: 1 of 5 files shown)


taylorotwell:
This doesn't seem to actually be incorporated anywhere in terms of actual usage?

PR #764, opened 2026-07-03
Feature: mistral file store support

dprinzensteiner:
Mistral now supports vector / file stores similar to how openai or google already does. For reference look at their API docs here: https://docs.mistral.ai/api/endpoint/beta/libraries

Together with claude, we added support for file storage and then usage via the FileSearch Tool.

I already ran the test suite with my mistral API key and they pass successfully.

I will also real world test this in an existing laravel application and add my thoughts in a comment here.


dprinzensteiner:
Update as promised: Just tried it out with a basic chat with RAG and the FileSearch Tool, works like expected.

pushpak1300:
I am closing this pull request because it lacks sufficient explanation, tests, or both. It is difficult for us to merge pull requests without these things because the change may introduce breaking changes to the this package.


Feel free to re-submit your change with a thorough explanation of the feature and tests. Thanks!

PR #765, opened 2026-07-03
Persist standalone reasoning content in conversations

dbpolito:
Current behavior:

- We already persist reasoning when it belongs to a tool call.
- That reasoning is stored inside `tool_calls` and replayed later.
- But reasoning summaries that are not tied to a tool call are currently lost.

Why this was added:

- Some providers return standalone reasoning blocks.
- Those blocks can be needed for later conversation turns.
- This adds storage for only those standalone blocks.
- It avoids duplicating reasoning that is already stored with tool calls.


dbpolito:
@pushpak1300 i know this is a substantial change, but i preferred opened a PR instead of an issue to show the intent...

Currently we're losing summaries that aren't related to tool calls, and i do think it's valuable data to save.

pushpak1300:
can you fix ci ?

dbpolito:
@pushpak1300 

Updated this to avoid adding a new column.

Since provider replay state is already stored in `meta['provider_content_blocks']` for approval resumes, standalone reasoning blocks now use the same path.

Tool-call reasoning still stays on `tool_calls`, and provider ownership is preserved via `meta['provider']`.

CI is passing.

AngadSethi:
Thanks for this, @dbpolito!

@pushpak1300, whenever you have a moment, can you please take a look?

AngadSethi:
@dbpolito, it looks like standalone reasoning may still be dropped from the `AgentResponse::$steps` representation.

`StepResponse` contains `providerContentBlocks`, but `TextGenerationLoop::buildStep()` does not pass them into `Step`, and `Step` currently has no corresponding property.

Would it make sense to add `providerContentBlocks` to `Step` and forward `$result->providerContentBlocks` in `buildStep()`? That would preserve standalone reasoning on the specific generation step that produced it and keep the messages and steps representations consistent

dbpolito:
@AngadSethi good catch!

I added `providerContentBlocks` to `Step` and now forward `$result->providerContentBlocks` from `TextGenerationLoop::buildStep()`, so the steps representation stays consistent with messages.

AngadSethi:
Thank you!

pushpak1300:
merged and alredy on 1.x

PR #738, opened 2026-06-23
Add web search and citation support to OpenRouterProvider

pietervanleuven:
Adds web search support to `OpenRouterProvider`, including handling of `url_citation` annotations from OpenRouter's response — both in streaming (via `CitationEvent`) and non-streaming (via `Meta`) responses.

See [OpenRouter web search docs](https://openrouter.ai/docs/guides/features/server-tools/web-search) for details on the annotations format.

PR #766, opened 2026-07-05
Fix OpenRouter test for new provider options API

mrdzen:
#738 landed a test using the two-argument
withProviderOptions('openrouter', [...]) signature that #753 (merged later) replaced with withProviderOptions(array|Closure), so the suite fails on 0.x with a TypeError. Update the call to the flat-array form, matching how #753 migrated the equivalent Anthropic and OpenAI web search tests.


pushpak1300:
Thanks.

Issue #770, opened 2026-07-06
🖱️Feature Request: Add support for client-executed tools (browser-side tools)


vinitkadam03:
Description:

Some tool calls can't or shouldn't run on the server — they need to interact with the browser directly, collect input from the user, or access device APIs. Today there's no way to express this in Laravel AI; those interactions must live outside the agentic loop.

Proposed design

A `#[ClientSideTool]` marker attribute on a tool class signals that the server should not execute it:

```php
use Laravel\Ai\Attributes\ClientSideTool;

#[ClientSideTool]
class GetUserLocationTool implements Tool
{
    public function handle(Request $request): string { return ''; }
    // ...
}
```

When the LLM calls a `#[ClientSideTool]` tool during generation, `TextGenerationLoop` skips execution, stops the loop, and surfaces the pending call in the response. The caller checks $response->hasPendingClientSideToolCalls(), ships the call to the browser, executes it client-side, then resumes the conversation on the next request by calling $store->storeToolResults($conversationId, $mergedResults) before calling prompt() again.

e.g.
The LLM sees this schema:
```json
{
  "name": "request_user_input",
  "parameters": {
    "type": "object",
    "properties": {
      "questions": {
        "type": "array",
        "items": {
          "type": "object",
          "properties": {
            "id":      { "type": "string" },
            "message": { "type": "string" },
            "type":    { "type": "string", "enum": ["text", "select", "confirm"] },
            "options": { "type": "array", "items": { "type": "string" } }
          },
          "required": ["id", "message", "type"]
        }
      }
    },
    "required": ["questions"]
  }
}
```
LLM tool call 
```json
{
  "name": "request_user_input",
  "arguments": {
    "questions": [
      {
        "id": "cabin_class",
        "message": "Which cabin class?",
        "type": "select",
        "options": ["Economy", "Business", "First"]
      },
      {
        "id": "meal_preference",
        "message": "Meal preference?",
        "type": "select",
        "options": ["Standard", "Vegetarian", "Vegan", "Halal"]
      },
      {
        "id": "special_requests",
        "message": "Any special requests?",
        "type": "text"
      }
    ]
  }
}
```

Flow

1. prompt() → LLM calls GetUserLocation → server skips execution
2. response->hasPendingClientSideToolCalls() === true
3. Frontend executes the tool (geolocation API, etc.)
4. $store->storeToolResults($conversationId, [...$serverResults, $clientResult])
5. prompt() on next request → loop resumes with complete tool results

[...]

PR #773, opened 2026-07-07
Add human-in-the-loop

pushpak1300:
This PR adds human-in-the-loop approval for tool calls with external side effects. An approvable tool can pause an agent before execution, expose the pending calls to the application, and resume from conversation history after a person approves, edits, or rejects them.

Approval works across `prompt()`, `stream()`, and `queue()`. Gated tools require a resumable `Conversational` agent because the paused tool calls are stored in its message history.

#### Gating a tool

Implement `Approvable` and use `InteractsWithApproval`. Tools are gated by default, while `needsApproval()` may decide per request:

```php
use Laravel\Ai\Approvals\Approval;
use Laravel\Ai\Concerns\InteractsWithApproval;
use Laravel\Ai\Contracts\Approvable;
use Laravel\Ai\Tools\Request;

class DeleteFile implements Tool, Approvable
{
    use InteractsWithApproval;

    protected function needsApproval(Request $request): bool|Approval
    {
        return str_starts_with($request['path'], 'storage/tmp/')
            ? false
            : Approval::required('Deletes a tracked project file');
    }

    public function handle(Request $request): string
    {
        Storage::delete($request['path']);

        return "Deleted {$request['path']}";
    }
}
```

The agent may override the tool's default:

```php
{
    return [
        (new DeleteFile)->withoutApproval(),
        (new SendEmail)->requireApproval('Review outgoing mail before it sends'),
    ];
}
```

The built-in `WriteFile`, `CopyFile`, and `DeleteFile` filesystem tools now require approval by default.

#### Pausing and resuming

A gated call is returned without being executed. Ungated calls in the same step still run.

```php
$response = $agent->forUser($user)->prompt('Delete config/app.php');

$response->awaitingApproval(); // true
$response->pendingApprovals;   // Collection<PendingApproval>
```

Resume the conversation with decisions keyed by tool call ID:

```php
use Laravel\Ai\Approvals\Decision;

$agent->continue($conversationId, as: $user)->prompt([
    'call_abc' => true,
    'call_def' => false,
    'call_ghi' => Decision::edit(['path' => '/tmp/x']),
    'call_jkl' => Decision::reject('Already deleted'),
]);
```

A rejection with a result continues the model loop. A bare rejection records the rejection and stops it. Every pending call needs a decision unless a `'*'` default is supplied:

```php
$agent->continue($conversationId, as: $user)->prompt(Decision::approveAll());

$agent->continue($conversationId, as: $user)->prompt([
    'call_abc' => Decision::edit(['path' => '/tmp/x']),
    '*' => false,
]);
```

Unknown, missing, or already-resolved IDs throw `ApprovalMismatchException`.

**Execution semantics.** Approved tool results are recorded durably before generation continues, so a retry never executes a resolved call twice — which also means a resume whose continuation fails has already consumed its decisions; continue with a new user prompt rather than retrying the decision map (a retry returns `409` with the current state). Ungated tool calls in a pausing step execute immediately — a pause is per-call, not per-step, so rejecting a gated call does not undo its ungated companions. Concurrent resumes may still reach a tool together, so external side effects should be idempotent on `$request->toolCallId()`.

#### Over HTTP

The package does not prescribe a request or response envelope. A continuation route can accept either a message or an ID-keyed decision map:

```php
Route::post('/chat/{conversation}', function (Request $request, string $conversation) {
    $response = (new FileAssistant)
        ->continue($conversation, as: $request->user())
        ->prompt(
            $request->filled('decisions')
                ? $request->collect('decisions')->all()
                : $request->input('message')
        );

    return $response->awaitingApproval()
        ? [
            'status' => 'awaiting_approval',
            'conversation_id' => $response->conversationId,
            'approvals' => $response->pendingApprovals,
        ]
        : $response->text;
});
```

Approval mismatches return `409` with the current pending approvals, allowing the client to refresh stale approval UI.

#### Streaming, queues, and broadcasting

The same decision map works for every delivery style:

| Method | Pause surfaces via |
|---|---|
| `prompt()` | `awaitingApproval()` and `pendingApprovals` |
| `stream()` | `tool_approval_request` stream event |
| `queue()` | `then()` callback or `ToolApprovalRequested` event |
| `broadcast()` / `broadcastOnQueue()` | `tool_approval_request` broadcast event |

```js
Echo.channel(`chat.${conversationId}`)
  .listen('.tool_approval_request', (event) => renderApprovals(event.approvals));
```

#### Vercel data stream protocol

Pauses emit the AI SDK's native `tool-approval-request` part after `tool-input-available`. Resumes emit `tool-output-available`, `tool-output-denied`, or `tool-output-error` before continuing the model stream.

```php
return (new FileAssistant)
    ->continue($conversation, as: $request->user())
    ->stream($request->input('decisions', []) ?: $request->input('message'))
    ->usingVercelDataProtocol(messageId: $request->input('message_id'));
```

Parsing `useChat` messages in the `approval-responded` state is intentionally left for a follow-up. Applications can currently extract those parts into the same ID-keyed decision map.

#### Why history and not an approvals table

Providers already require tool calls and results in conversation history when a run resumes. Storing approval state beside that history avoids maintaining a second copy. The paused assistant row records which calls remain pending, and after resolution each call's outcome stays readable from the row's tool results (rejections carry a `denied` flag), while `ToolApprovalRequested` can support a separate inbox view if needed.

> [!WARNING]
> Gated tools require a `Conversational` agent that remembers conversations for a participant. A gated tool on a non-resumable agent throws `ApprovalNotResumableException` instead of returning a pause that cannot be resumed.

#### Testing

Fakes can return pending approvals, and resumed prompts expose their decision map:

```php
FileAssistant::fake([
    AgentResponse::fakeAwaitingApproval([
        new PendingApproval('call_abc', 'DeleteFile', ['path' => 'config/app.php']),
    ]),
]);

FileAssistant::assertPrompted(fn (AgentPrompt $prompt) =>
    ($prompt->resume['call_abc'] ?? null)?->isApproved()
);
```

#### Breaking changes

Three changes affect an upgrading application. Everything else in this PR is additive.

**1. New `approval_state` column on the conversation messages table**

*What:* a new nullable `TEXT` column, `approval_state`, is added to the conversation messages migration.

*Why:* it marks which assistant turns are genuine approval pauses (so an ordinary max-steps truncation is not mistaken for one), and it holds the coordination state that records a resolved approval result durably before the run continues — the guarantee that a crash or retry cannot run an approved tool twice. The column is retained after resolution (its pending map empties), which bounds resume bookkeeping to turns that actually paused rather than every row in the conversation.

*Upgrade:* fresh installs get the column from the published migration. If you have already published and run the conversation migrations, add the column with a new migration and run `php artisan migrate`:

```php
use Illuminate\Database\Migrations\Migration;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Facades\Schema;

return new class extends Migration
{
    public function up(): void
    {
        Schema::table(config('ai.conversations.tables.messages', 'agent_conversation_messages'), function (Blueprint $table) {
            $table->text('approval_state')->nullable()->after('meta');
        });
    }

    public function down(): void
    {
        Schema::table(config('ai.conversations.tables.messages', 'agent_conversation_messages'), function (Blueprint $table) {
            $table->dropColumn('approval_state');
        });
    }
};
```

**2. `Agent` contract accepts `array|string`**

*What:* `Agent::prompt()`, `stream()`, `queue()`, `broadcast()`, `broadcastNow()`, and `broadcastOnQueue()` now accept `array|string` instead of `string`, where the array is a tool-call-id-keyed map of decisions.

*Why:* a resume reuses the same entry points, passing decisions where a message would go, so one method serves both starting and resuming a run across every delivery style.

*Upgrade:* nothing to do if your agents use the `Promptable` trait (the shipped path). If you implement `Laravel\Ai\Contracts\Agent` directly, widen those prompt parameters to `array|string` to match the contract.

**3. `ConversationStore` contract gains `storeApprovalResults()`**

*What:* the `ConversationStore` interface adds `storeApprovalResults(string $conversationId, string|int|null $participantId, array $toolResults): void`, and `storeAssistantMessage()` now returns `?string` — `null` when a resume produced nothing new to store.

*Why:* resolved approval results must be recorded durably on the paused turn before generation continues, which is what makes an approved tool safe against crash-retry double execution.

*Upgrade:* nothing to do if you use the shipped database store. If you bind a custom `ConversationStore`, implement `storeApprovalResults()` to merge the given results into the paused turn and throw `ApprovalMismatchException` when no paused turn matches. Existing `storeAssistantMessage()` implementations returning `string` continue to satisfy the widened return type.


vinitkadam03:
Great addition!

I was tinkering with something similar in Prism https://github.com/prism-php/prism/pull/932 (and using it in production for quite some time)

Separately was tinkering with DB persistence for paused state + wiring it through the Vercel AI SDK UI Data protocol here in Laravel AI SDK toofor implemting Client executed tools https://github.com/laravel/ai/pull/771, so it's great to see this land.

Since this establishes the pause/resume pattern, it'd be a natural fit for client-executed tools too - had opened #770 with a rough design.

src/Promptable.php
```diff
@@ -135,22 +166,41 @@ function () use ($providers, $prompt, $attachments, $resolvedTimeout, $invocatio
...
         );
     }
 
+    /**
+     * Split a prompt into its text and tool approval resume parts.
+     *
+     * @param  array<string, Decision|bool>|string  $prompt
+     * @return array{string, ?array<string, Decision>}
+     */
+    private function extractResume(array|string $prompt): array
```
WendellAdriel: This split gives `prompt()`, `stream()`, and `queue()` resume support, but `broadcast()`, `broadcastNow()`, and `broadcastOnQueue()` still type the prompt as `string` — so a paused broadcast conversation can't be resumed through the same delivery style. Let's widen those three to `array|string` too and add a test for a broadcast resume.

src/Promptable.php
```diff
@@ -226,6 +276,14 @@ private function getProvidersAndModelsForFailover(Lab|array|string|null $provide
         return $providers;
     }
 
+    /**
+     * Get the single provider / model pair a resume must run against, since a resume may not fail over to a different provider.
+     */
+    private function providersForResume(Lab|array|string|null $provider, ?string $model): array
```
WendellAdriel: Taking the first configured provider means a run that paused after failing over to `backup` resumes against `primary`, replaying `backup`'s raw `provider_content_blocks` to the wrong provider. Let's prefer the provider/model pair recorded on the paused turn's meta, or document that failover setups must pass an explicit provider on resume, with a regression test for the failover-then-resume case.

src/Gateway/TextGenerationLoop.php
```diff
@@ -45,13 +58,33 @@ public function generate(
         ?array $schema = null,
         ?TextGenerationOptions $options = null,
         ?int $timeout = null,
+        ?array $approval = null,
```
WendellAdriel, line 64: With these additions, `generate()` and `stream()` are up to 10–12 positional parameters and every call site carries trailing `null, null, null` padding. Since `$approval`, `$recordApprovalResults`, and `$validatedApproval` always travel together, we can group them into a small DTO for example.

src/Middleware/RememberConversation.php
```diff
@@ -36,7 +44,7 @@ public function handle(AgentPrompt $prompt, Closure $next)
             if (! $agent->currentConversation()) {
                 $conversationId = $this->store->storeConversation(
                     $agent->conversationParticipant()?->id,
-                    $this->generateTitle($prompt->prompt)
+                    $this->generateTitle($prompt->resume !== null ? 'Tool approval' : $prompt->prompt)
```
WendellAdriel: The `'Tool approval'` title branch is unreachable: a resume with no current conversation always throws `ApprovalMismatchException` before this callback runs. Let's drop the ternary and pass `$prompt->prompt` directly.

src/Middleware/RememberConversation.php
```diff
@@ -27,6 +27,14 @@ public function __construct(
      * Handle the incoming prompt.
      */
     public function handle(AgentPrompt $prompt, Closure $next)
+    {
+        return $this->remember($prompt, $next);
```
WendellAdriel: Do we need to wrap this in the `remember` method? It doesn't seem to be reused anywhere else, so we could just inline the logic here instead.

src/Contracts/Agent.php
```diff
@@ -42,7 +47,7 @@ public function stream(
      * Invoke the agent in a queued job.
      */
     public function queue(
-        string $prompt,
+        array|string $prompt,
```
WendellAdriel: Nit: docblock here was not updated like in `prompt()` or `stream()` for the `$prompt` change.

src/Storage/DatabaseConversationStore.php
```diff
@@ -117,6 +180,49 @@ public function storeAssistantMessage(string $conversationId, string|int|null $u
         return $messageId;
     }
 
+    /**
+     * Get every tool-result id already recorded in the conversation.
+     *
+     * @return array<int, string>
+     */
+    protected function existingToolResultIds(string $conversationId): array
```
WendellAdriel, line 140: This loads and decodes `tool_results` for every assistant row in the conversation on each resume persist, so the scan grows unbounded with conversation length. Let's bind it to the same window `getLatestConversationMessages()` uses, since only in-window rows can produce duplicates.

src/Prompts/AgentPrompt.php
```diff
@@ -63,6 +69,10 @@ public function append(string $prompt): AgentPrompt
      */
     public function revise(string $prompt, Collection|array|null $attachments = null): AgentPrompt
     {
+        if ($this->resume !== null) {
```
WendellAdriel: This guard is correct, but any existing middleware that unconditionally calls `$prompt->append()` or `revise()` will now throw on resume flows. Let's call that out in the docs so middleware authors know to skip prompts where `$prompt->resume !== null`.

WendellAdriel (changes requested):
Both of these are documentation gaps rather than code changes:

**Breaking changes**: the `ConversationStore` contract changes and the removed `AgentPrompt::withAttachments()` aren't listed as breaking changes, and UPGRADE.md covers none of the breaks. Let's add an upgrade section covering the `approval_state` column, the `Agent` contract widening, the store contract changes, and the removed method.

**Approval edge behaviors**: a failed resume consumes its decisions, so the conversation can only continue with a new user prompt, and ungated companion tools still execute in a pausing step. Let's document both alongside the `$request->toolCallId()` idempotency guidance.

src/Prompts/AgentPrompt.php
pushpak1300: I’m not throwing exceptions and still discarding middleware if it’s resuming.

src/Gateway/TextGenerationLoop.php
pushpak1300: The call sites don't actually carry null, null, null padding — there are only three, and the single null at the stream site is the pre-existing $schema param. The three approval params also only travel together in stream(), so I'd rather keep the signatures flat than introduce a DTO for two call sites.

src/Promptable.php
pushpak1300: Resuming on the first configured provider is intentional — it matches how every turn behaves after a failover (the next prompt always returns to the primary, replaying generic history cross-provider).

taylorotwell:
- Renamed the approval concern from `InteractsWithApproval` to `InteractsWithApprovals`.

```php
use Laravel\Ai\Concerns\InteractsWithApprovals;

class DeleteFile implements Tool, Approvable
{
    use InteractsWithApprovals;
}
```

- Replaced raw approval decision arrays with the new `Laravel\Ai\Approvals\Decisions` value object. Agent methods now accept `Decisions|string` instead of `array|string`.

```php
use Laravel\Ai\Approvals\Decision;
use Laravel\Ai\Approvals\Decisions;

$agent->prompt(Decisions::from([
    'call_abc' => true,
    'call_def' => Decision::reject('Already handled'),
]));
```

- Added chainable wildcard helpers for calls without explicit decisions.

```php
Decisions::from([
    'call_abc' => true,
])->rejectRemaining();

Decisions::from([
    'call_abc' => false,
])->approveRemaining();
```

- `Decision::approveAll()` and `Decision::rejectAll()` now return `Decisions`, so existing blanket-decision usage remains concise.

```php
$agent->prompt(Decision::approveAll());
$agent->prompt(Decision::rejectAll('Not now'));
```

- Replaced `AgentPrompt::$resume` and `QueuedAgentPrompt::$resume` with `$approvalDecisions`. Middleware and fake assertions can use `hasApprovalDecisions()` and `get()`.

```php
if ($prompt->hasApprovalDecisions()) {
    $decision = $prompt->approvalDecisions?->get('call_abc');
}
```

- Queue and broadcast jobs now carry a single `Decisions|string` prompt input rather than separate prompt and resume fields.

- Vercel approval-resume tool results are emitted without a tool call in the current stream only when an existing `messageId` is supplied. Results with neither a current-stream tool call nor an existing message are skipped to avoid orphaned `tool-output-*` parts.

```php
$response->usingVercelDataProtocol(
    messageId: $request->input('message_id'),
);
```

stuartbrameld:
just seen this, made my day, thanks

"laravel/ai v0.10.0 (released yesterday) ships first-party human-in-the-loop tool approval, which is the SDK-native version of the largest piece of custom harness you maintain."

🙏

PR #774, opened 2026-07-08
Let sub-agents continue the parent conversation

JVillator0:
Currently `AgentTool` always runs a delegated agent in isolation. It calls `->prompt($task)` with just the task string, so a sub-agent never sees the conversation it was called from (the default tool description even says so). If you want the delegated agent to keep talking in the same thread, remembering who the user is and what was already said, you have to re-feed that context by hand on every call.

This PR lets a sub-agent continue the parent's conversation when both sides remember conversations:

```php
class OrchestratorAgent implements Agent, HasTools, RemembersConversations
{
    use Promptable, RemembersConversations;

    {
        return [new SupportAgent]; // also implements RemembersConversations
    }
}
```

When the orchestrator is already in a conversation and delegates to `SupportAgent`, the sub-agent now continues that same conversation instead of starting cold, so it can read the history while handling its task.

Under the hood `AgentTool` takes the parent agent as an optional second argument, and the framework passes it in when it auto-wraps an agent returned from `tools()`. Right before the sub-agent runs, if both agents are `RemembersConversations` and the parent already has an active conversation, it calls `continue()` with the parent's conversation id and participant. Everything downstream already works: `GeneratesText` seeds `messages()` into the prompt and the `RememberConversation` middleware writes the sub-agent's turn back into the same conversation, so there's no new plumbing.

A sub-agent that doesn't remember conversations still runs in isolation, and `new AgentTool($agent)` keeps working, so nothing changes for existing code.

Went with an optional constructor arg on `AgentTool` rather than a new hook or contract change, so the wrapper stays the single place this behavior lives.

### Notes

* Sharing only kicks in when the parent already has a conversation with a participant. On a brand-new conversation the id isn't persisted until after the run, so on that first turn there's nothing to share yet and the sub-agent stays isolated.
* It's opt-in by capability: a sub-agent joins the parent thread only if it implements `RemembersConversations`, so history never leaks into agents meant to be stateless.

### DIFF
--- a/src/Providers/Concerns/GeneratesText.php
+++ b/src/Providers/Concerns/GeneratesText.php
@@ -125,7 +125,7 @@ trait GeneratesText
         }
 
         return array_map(
-            fn ($tool) => $this->resolveTool($tool),
+            fn ($tool) => $this->resolveTool($tool, $agent),
             [...$agent->tools()],
         );
     }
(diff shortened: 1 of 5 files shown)


JVillator0:
Hi @pushpak1300, would you be willing to review this? 😄 
I would love to be able to use this change

pushpak1300:
The sub-agent's turn should be written into the user's conversation thread. ← almost nobody wants this

pushpak1300:
Thanks for your pull request!

PR #778, opened 2026-07-08
Feature/citations and gemini config fix

Neunerlei:
Hey there.

I recently added laravel/ai to [our project](https://github.com/hawk-digital-environments/HAWKI/tree/feature/v2.5.0).
It works like a charm, but I noticed a three inconsistencies, this PR tries to resolve.

1. The Gemini Gateway tries to be smart by building the `generationConfig` object, but by doing so it locks away some options to forward to the provider, like [safetySettings](https://github.com/hawk-digital-environments/HAWKI/blob/feature/v2.5.0/app/Services/Ai/Providers/Adapters/Implementations/GeminiAdapter.php#L64). Also I was quite confused at first why it told me that "generationConfig" was not a valid config key, until I saw the implementation. To resolve this issue, I added a lean preprocessor step that hoists both values into the correct layer. 

2. I noticed that citations were only extracted while "not streaming". Since we are using streaming everywhere I could not access the groundingMetadata (for Google) or annotations (for OpenAi). This PR provides support for both.

3. I extended the UrlCitation object, to handle more complex cases. It should be completely backward compatible.
    1. groundingMetadata can contain citations that are relevant not for a single block of data, but for multiple. Therefore I extended UrlCitation globally to be able to hold more than one range, since I guess this can come handy in the future.
  
    2. Google gives us indexes for the citations as byte offsets, openAi for character offsets. To allow the frontend to determine the difference there is also a new isByteOffset flag on the range.

P.S. The main code was written by hand. The tests were written by Claude and reviewed by human eyes ;)

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -9,9 +9,13 @@
 /.github export-ignore
 /art export-ignore
 /tests export-ignore
+/workbench export-ignore
 .editorconfig export-ignore
 .gitattributes export-ignore
 .gitignore export-ignore
 CHANGELOG.md export-ignore
 phpunit.xml.dist export-ignore
+phpstan.neon.dist export-ignore
+rector.php export-ignore
+testbench.yaml export-ignore
 UPGRADE.md export-ignore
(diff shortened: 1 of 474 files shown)


pushpak1300:
Closing this as solving all of this issue individually.

PR #781, opened 2026-07-09
Trigger failover when a provider connection fails

### LINKED ISSUE #775: [Bug]: Automatic failover does not trigger on ConnectionException (e.g., connection refused)
When using the `laravel/ai SDK` with automatic failover (providing an array of fallback providers), the failover mechanism fails to trigger if the primary provider is completely unreachable (for example, if an Ollama instance is offline or the host is down).

In this scenario, Laravel's HTTP client throws an `Illuminate\Http\Client\ConnectionException`. However, the `HandlesFailoverErrors` trait currently only catches `Illuminate\Http\Client\RequestException`. Because `ConnectionException` does not inherit from `RequestException` in Laravel, the exception bubbles up, crashing the application instead of failing over to the fallback provider.

**Steps To Reproduce:**

1. Configure an agent with a primary provider pointing to an offline service (e.g., Ollama on a stopped local container).
2. Configure a valid fallback provider (e.g., OpenAI or Gemini).

```php
$agent->stream($prompt, provider: ['ollama', 'openai']);
```

3. Execute the prompt.
4. Instead of failing over to OpenAI, the application crashes with: `Connection refused for URI http://localhost:11434/api/chat`.
**Expected Behavior:** A connection failure (like timeouts or connection refused) should be treated as provider unavailability. The SDK should catch the `ConnectionException` and automatically failover to the next provider in the array.

**Suggested Fix:** In `Laravel\Ai\Gateway\Concerns\HandlesFailoverErrors`, add a catch block for `ConnectionException` so it throws a `ProviderOverloadedException` (which extends `FailoverableException`).

```php
    protected function withErrorHandling(string $providerName, Closure $callback): mixed
    {
        try {
            return $callback();
        } catch (\Illuminate\Http\Client\ConnectionException $e) {
            // Treat connection refused/timeouts as an overloaded or unavailable provider
            throw ProviderOverloadedException::forProvider(
                $providerName, $e->getCode(), $e
            );
        } catch (RequestException $e) {
            // ... existing RequestException handling ...
```

JVillator0:
Currently, automatic failover doesn't trigger when the primary provider is completely unreachable (host down, connection refused, an offline Ollama instance). In that case Laravel's HTTP client throws an `Illuminate\Http\Client\ConnectionException`, but `HandlesFailoverErrors` only catches `RequestException`. Since `ConnectionException` is a separate branch of the client exception hierarchy (it does not extend `RequestException`), it slips past the handler and bubbles up as a fatal error instead of failing over to the next provider.

This PR catches `ConnectionException` in `withErrorHandling` and rethrows it as a new `ProviderConnectionException`, which implements `FailoverableException` like the existing `RateLimitedException` and `ProviderOverloadedException`. The fix lives in the shared `HandlesFailoverErrors` trait that every gateway uses, so it applies to all providers at once, for both `prompt()` and `stream()`.

A single provider still surfaces the failure, now as a clear `ProviderConnectionException` instead of a raw `ConnectionException`, and the rest of the handler (429, 402, 503, quota patterns) is untouched, so nothing changes for existing code.

Fixes #775

### DIFF
--- a/src/Exceptions/ProviderConnectionException.php
+++ b/src/Exceptions/ProviderConnectionException.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace Laravel\Ai\Exceptions;
+
+use Throwable;
+
+class ProviderConnectionException extends AiException implements FailoverableException
+{
+    public static function forProvider(string $provider, int $code = 0, ?Throwable $previous = null): self
+    {
+        return new self(
+            'Could not connect to AI provider ['.$provider.'].',
+            $code,
+            $previous,
+        );
+    }
+}
--- a/src/Gateway/Concerns/HandlesFailoverErrors.php
+++ b/src/Gateway/Concerns/HandlesFailoverErrors.php
@@ -3,8 +3,10 @@
 namespace Laravel\Ai\Gateway\Concerns;
 
 use Closure;
+use Illuminate\Http\Client\ConnectionException;
 use Illuminate\Http\Client\RequestException;
 use Laravel\Ai\Exceptions\InsufficientCreditsException;
+use Laravel\Ai\Exceptions\ProviderConnectionException;
 use Laravel\Ai\Exceptions\ProviderOverloadedException;
 use Laravel\Ai\Exceptions\RateLimitedException;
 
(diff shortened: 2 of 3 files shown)


pushpak1300:
can you please fix conflicts ?

JVillator0:
Conflicts solved @pushpak1300 🙌

PR #789, opened 2026-07-11
Fix`TextDelta::combine` glues multi-step streamed text together mid-sentence

crishoj:
We run a multi-tool agent (using `gemini-3.5-flash`) that streams its replies over broadcast and persists conversations with the database store.

We noticed that whenever a reply spanned multiple generation steps — narration, a tool call, then more narration — the *persisted* message glued the step texts together mid-sentence:

> `…you can see this in the other guides too.Let's do that!Now that we are back in the editor…`

Live rendering was fine (the client sees the tool-call events between the deltas), but the _stored_ transcript reads as run-on self-talk.

Once a few of those turns accumulate in a conversation, the rehydrated history itself becomes a repetition attractor for the LLM, often causing verbatim phrase loops that only stop at the output-token cap:

image

## The problem

The cause is in `TextDelta::combine()`: it joins every text delta of the whole run with an empty string, blind to step boundaries.

## The fix

Every gateway mints a fresh `messageId` per stream call (i.e. per generation step), and each `TextDelta` carries it. `combine()` now groups deltas by `messageId`, joins each step's deltas exactly as before, drops whitespace-only steps, and joins the steps with a blank line — so the persisted text matches what the user actually saw: separate utterances.

### DIFF
--- a/src/Streaming/Events/TextDelta.php
+++ b/src/Streaming/Events/TextDelta.php
@@ -17,14 +17,22 @@ class TextDelta extends StreamEvent
 
     /**
      * Combine the text deltas in the given collection of events into a single string.
+     *
+     * Deltas from a multi-step generation carry a distinct message ID per step,
+     * and each step's text is a self-contained utterance (typically narration
+     * around a tool call). Steps are therefore joined with a blank line instead
+     * of being run together mid-sentence.
      */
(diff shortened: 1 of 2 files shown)


pushpak1300:
Thanks.

PR #793, opened 2026-07-14
Add stateless transcript prompting to the agent API

pushpak1300:
Currently there's no way to hand an agent a client-supplied conversation history at call time. `Conversational::messages()` is class-level only and conflicts with `RemembersConversations`, so a stateless backend (client posts the full transcript every request, like Vercel's `useChat`) has nowhere to plug in except the anonymous `agent()` helper.

This widens the `Promptable` verbs (`prompt`, `stream`, `queue`, `broadcast`, `broadcastNow`, `broadcastOnQueue`) to accept a full transcript in addition to a plain string:

```php
// before: no call-time transcript support on a named agent
$response = (new SalesCoach)->prompt('What about digital products?');

// after
$response = (new SalesCoach)->prompt([
    new UserMessage('What is our refund policy?'),
    new AssistantMessage('Our refund policy allows returns within 30 days...'),
    new UserMessage('What about digital products?'),
]);

// the plain array shape Message::tryFrom() already understands works too
$response = (new SalesCoach)->prompt($request->input('messages'));
```

The trailing user message is treated as "the prompt", everything before it is history. An explicit transcript overrides `Conversational::messages()` for that call, and throws a `LogicException` if the agent has an active `RemembersConversations` conversation (two owners of history is a footgun).

### DIFF
--- a/src/Concerns/InteractsWithFakeAgents.php
+++ b/src/Concerns/InteractsWithFakeAgents.php
@@ -84,7 +84,7 @@ trait InteractsWithFakeAgents
         ?string $message = null): self
     {
         $callback = is_string($callback)
-            ? fn ($prompt) => $prompt->prompt === $callback
+            ? fn ($prompt) => $prompt->text() === $callback
             : $callback;
 
         PHPUnit::assertTrue(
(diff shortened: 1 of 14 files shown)


pushpak1300:
Closing in favor of a smaller approach: a UiMessage converter that pairs useChat with the existing RemembersConversations/ConversationStore infrastructure, rather than widening the prompt surface.

PR #794, opened 2026-07-14
Support Vercel AI SDK UI messages, chat inputs, and ad-hoc history

pushpak1300:
Currently the package speaks Vercel's UI message stream on the way out (`usingVercelDataProtocol()`) but there is no way in. A `useChat` frontend POSTs its transcript in Vercel's parts-based `UIMessage` shape, and every app rewrites the same mapper to turn that into something an agent accepts, usually getting file and tool parts subtly wrong. There is also no way to hand an agent client-owned history at all, so even a correctly mapped transcript has nowhere to go.

After this PR, a stateless chat endpoint where the client owns the history is:

```php
use Laravel\Ai\Vercel\Vercel;

return (new ChatAgent)
    ->withMessages(Vercel::messagesFrom($request->input('messages')))
    ->stream(new UserMessage($text, $attachments))
    ->usingVercelDataProtocol();
```

**A `Vercel` converter for inbound messages.** Text parts become content, file parts on user messages become attachments (data URLs to `Base64*`, http URLs to `Remote*` by media type). Reasoning, `step-start`, `data-*` and tool parts are dropped the same way Vercel's own `convertToModelMessages()` drops them, and a `system` role is rejected so a client can't inject a system prompt through the transcript. An earlier cut taught `Message::tryFrom()` the shape directly, but that bakes one vendor's wire format into a core value object, so the translation lives in its own class, symmetric with the outbound side.

**Prompt inputs.** `prompt()`, `stream()`, `queue()` and the `broadcast` methods now accept `ChatInput|Decisions|UserMessage|string`. A `UserMessage` streams directly and its attachments merge in. `Contracts\ChatInput` is a two-method contract (`message()`, `decisions()`, resolved decisions first) meant for frontend request objects, so a follow-up `Vercel::chat($request)` can pass straight into `stream()`, and other frontend libraries implement the same contract instead of forking the entry points.

**`withMessages()` for client-owned history.** Accepts `Message` objects or raw `['role' => ..., 'content' => ...]` arrays, is sent ahead of the prompt, and serializes into queued jobs with the agent. It throws on `Conversational` agents. One history source at a time felt safer than silent precedence, open to relaxing that.

The stream protocol fixes that were previously part of this branch (finish part metadata, mid-stream error part) now live on the `vercel-stream-protocol` branch so they can be reviewed on their own. Tests cover each piece.

### DIFF
--- a/src/Contracts/ChatInput.php
+++ b/src/Contracts/ChatInput.php
@@ -1,0 +1,19 @@
+<?php
+
+namespace Laravel\Ai\Contracts;
+
+use Laravel\Ai\Approvals\Decisions;
+use Laravel\Ai\Messages\UserMessage;
+
+interface ChatInput
+{
+    /**
+     * Get the newest user message, if the input contains one.
+     */
+    public function message(): ?UserMessage;
+
+    /**
+     * Get the tool approval decisions, if the input contains any.
+     */
+    public function decisions(): ?Decisions;
+}
--- a/src/Files/File.php
+++ b/src/Files/File.php
@@ -59,44 +59,6 @@ abstract class File implements HasName, HasProviderOptions
         return $file;
     }
 
-    /**
-     * Create a file instance from a Vercel AI SDK UI message file part.
-     *
-     * @param  array<string, mixed>  $part
-     */
-    public static function fromUiMessagePart(array $part): ?File
-    {
-        $url = $part['url'] ?? '';
-        $mime = strtolower($part['mediaType'] ?? '');
-
-        $file = match (true) {
-            blank($url) => null,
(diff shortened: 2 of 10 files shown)


vinitkadam03:
This is awesome!! :tada: 

We've been maintaining a mapper for this - calling `convertToModelMessages` on the frontend before sending the request so UI messages map to model messages and are easier to parse into the Prism / AI SDK format. The downside is that collapsing on the client means the extra parts never reach the backend at all.

What I like here is that the raw `UIMessage` shape stays intact all the way to `tryFrom()`. It ignores the same parts as `convertToModelMessages` for now, but the full payload still lands server-side, which leaves room to handle more of it later.

PR #795, opened 2026-07-14
Store conversation participants polymorphically

### LINKED ISSUE #666: [Feature] [Feature Request] RemembersConversations Trait Does Not Support Different User Providers, Causing Conversation Leakage Between Users with Same ID

pushpak1300:
Resolves #666 by scoping conversations to a full participant morph (`participant_type` + `participant_id`), following the same pattern Sanctum uses for `personal_access_tokens` (`tokenable`) and the framework uses for database notifications (`notifiable`).

### Problem

Conversation ownership is keyed on a bare `user_id`. Any two models that share an id (e.g. `User` #1 and `Admin` #1, or apps with multiple auth providers) therefore collide, and conversation history leaks between them:

```php
(new SalesCoach)->forUser(User::find(1))->prompt('Hello from user');

// Admin #1 resolves the user's most recent thread, because it matches on id alone...
(new SalesCoach)->continueLastConversation(Admin::find(1));
```

### Approach

Conversations and messages now record a required `participant_type` + `participant_id` morph pair, derived automatically from the model instance — there is no config key and no "as type" API to remember. The identity travels with the object you pass, exactly like Sanctum's `$user->createToken()` or notifications' `$user->notify()`.

- The default migration creates `participant_type` + `participant_id` on both tables, indexed together with `updated_at`.
- `Conversation::participantType($participant)` resolves the discriminator via `getMorphClass()`, and `Conversation::participantKey($participant)` resolves the id via `getKey()` — so morph maps and custom primary keys both behave exactly as they do everywhere else in the framework. Plain non-Eloquent objects fall back to the class name and the `id` property.
- `continueLastConversation()` and the `RememberConversation` middleware thread the type through every store read and write.
- `HasConversations::conversations()` is a genuine `morphMany`, and `Conversation::participant()` exposes the inverse `morphTo`.

### Usage

Pass any model — the type is derived from the instance:

```php
// participant_id = 1, participant_type = App\Models\User
(new SalesCoach)->forUser(User::find(1))->prompt('Hello');

// participant_id = 1, participant_type = App\Models\Team — no collision
(new SalesCoach)->forParticipant(Team::find(1))->prompt('Hello');

// Resolves the team's latest conversation, never the user's...
(new SalesCoach)->continueLastConversation(Team::find(1))->prompt('Continue');

// Continuing by id works for any participant type...
(new SalesCoach)->continue($conversationId, Team::find(1))->prompt('Pick up where we left off');
```

`forParticipant()` is the new canonical method; `forUser()` is retained as an alias since a user remains the common case.

Listing works through the morph in both directions:

```php
class Team extends Model
{
    use HasConversations; // $team->conversations is a morphMany, scoped by type
}

$conversation->participant; // morphTo back to the owning model
```

Pinning morph types with `Relation::enforceMorphMap([...])` is recommended so stored types survive class renames.

Note that `continue()` intentionally trusts the given conversation id and does not assert ownership — authorization belongs in the application (e.g. `$team->conversations()->findOrFail($id)` or a policy), consistent with route model binding.

### Breaking changes

- **`ConversationStore` contract:** all four write/read methods now receive the participant's morph type as a new, required argument:

  ```php
  // Before...
  public function latestConversationId(string|int $userId): ?string;
  public function storeConversation(string|int|null $userId, string $title): string;
  public function storeUserMessage(string $conversationId, string|int|null $userId, AgentPrompt $prompt): string;
  public function storeAssistantMessage(string $conversationId, string|int|null $userId, AgentPrompt $prompt, AgentResponse $response): string;

  // After...
  public function latestConversationId(string $participantType, string|int $participantId): ?string;
  public function storeConversation(string $participantType, string|int $participantId, string $title): string;
  public function storeUserMessage(string $conversationId, string $participantType, string|int $participantId, AgentPrompt $prompt): string;
  public function storeAssistantMessage(string $conversationId, string $participantType, string|int $participantId, AgentPrompt $prompt, AgentResponse $response): string;
  ```

  Custom store implementations must adopt the new signatures; stores that don't care about types may simply ignore the argument. The participant id is also no longer nullable — the middleware only runs when a participant is present, so the nullable case was dead weight in the contract.

- **`HasConversations::conversations()`** now returns `MorphMany` instead of `HasMany`. Only affects code that type-hinted the relationship.
- **Fresh installs** get `participant_type`/`participant_id` columns instead of `user_id` (the create migration was updated in place).

### Upgrading existing applications

The store reads and writes the participant columns exclusively, so existing installations must add the new columns and backfill them:

```php
Schema::table('agent_conversations', function (Blueprint $table) {
    $table->string('participant_type')->nullable()->after('user_id');
    $table->renameColumn('user_id', 'participant_id');
    $table->index(['participant_type', 'participant_id', 'updated_at'], 'participant_updated_at_index');
});

Schema::table('agent_conversation_messages', function (Blueprint $table) {
    $table->string('participant_type')->nullable()->after('user_id');
    $table->renameColumn('user_id', 'participant_id');
    $table->index(['participant_type', 'participant_id'], 'participant_index');
});

// Backfill so existing conversations remain visible to their owners...
DB::table('agent_conversations')
    ->whereNotNull('participant_id')
    ->update(['participant_type' => 'App\Models\User']);

DB::table('agent_conversation_messages')
    ->whereNotNull('participant_id')
    ->update(['participant_type' => 'App\Models\User']);
```

If you use a morph map, backfill with your map's alias instead of the class name. Skipping the backfill doesn't lose data, but unbackfilled conversations become invisible to `continueLastConversation()` and the `conversations()` relationship, since lookups now match on the participant's morph class.


pushpak1300:
Related to https://github.com/laravel/ai/pull/749

vinitkadam03:
This is a great addition!

Since we're already breaking the contract and touching the migration, would it be worth folding in message branching too? The use case I have in mind is editing an earlier user message and continuing the chat from there while keeping the old path around, the way ChatGPT / Claude / Open WebUI let you scroll between variants.

Mechanically I think that's a nullable `parent_id` on the messages table (so history becomes a tree instead of a flat list), plus something on the conversation to track the active leaf/branch so you know which path to replay.

The one wrinkle: branching complicates partial history loading. Today "give me the last N messages" is a simple `ORDER BY created_at LIMIT N`, but with a tree that's ambiguous, since N could follow any branch. You'd have to resolve the active leaf and walk `parent_id` upward N times (a recursive CTE), which can't ride a plain `(conversation_id, created_at)` index.

From what I can tell, ChatGPT and Open WebUI sidestep this entirely: they load the whole per-conversation tree and reconstruct the active thread client-side by walking from a leaf pointer (Open WebUI calls it `currentId`) up through `parentId`, paginating at the conversation-list level rather than within a conversation. So message-level partial loading and branching are a bit at odds unless we adopt that same full-tree-per-conversation model, which is probably worth deciding explicitly.

peterjthomson:
We love ya'll and opted in to this early with our eyes open, but from what I can see this release needs:

1. A new migration to shim to match the vendor migration. Including a data-migration.
2. An update to the index on the table.
3. Updates to all application code that was relying on the relationship. 

Absolutely no problem doing so. Just want to make sure I haven't misunderstood the upgrade path.

PR #797, opened 2026-07-16
Execute `then(...)` callback when faking the queued generation of transcription/image/audio/embeddings

gdebrauwer:
In a project, I have the following code:

```php
class MeetingController
{
    public function update(Meeting, UpdateMeetingRequest $request)
	{
		Transcription::fromStorage($request->audioFile())
			->queue()
			->then(function ($transcription) use ($meeting) {
				$meeting->update(['audio_transcription' => $transcription]);
			});
	}
}
```

In a test of this api call, I discovered that I could define `Transcription::fake()` but the `then()` callback would never be executed. That means I can't really test the logic that happens in that callback.

This PR fixes that:

```php
Transcription::fake(['foo bar baz']);

$this->putJson('/meeting/{$this->meeting->id', ['audio' => $this->audioFile])->assertSuccessful();

Transcription::assertQueued(fn (QueuedTranscriptionPrompt $prompt): bool => $prompt->audio->path === $this->audioFile);

// The 'then' callback is executed
$this->assertSame('foo bar baz', $this->meeting->audio_transcription);
```

If you still want the previous logic and don't want to execute the `then()` callback, then that is still possible by combining `Transcriptions::fake()` and `Queue::fake()`

```php
Transcription::fake(['foo bar baz']);
Queue::fake([GenerateTranscription::class]);

$this->putJson('/meeting/{$this->meeting->id', ['audio' => $this->audioFile])->assertSuccessful();

Transcription::assertQueued(fn (QueuedTranscriptionPrompt $prompt): bool => $prompt->audio->path === $this->audioFile);
Queue:assertPushed(GenerateTranscription::class);

// The 'then' callback is not executed
$this->assertNotSame('foo bar baz', $this->meeting->audio_transcription);
```

I fixed this also for faking queued image generation, queued embeddings generation, and queued image generation.

---

### Update

Agents now get the same treatment: `Agent::fake()` combined with `queue()->then(...)` or `broadcastOnQueue()` executes the callbacks as well. Since no code path returns `FakePendingDispatch` anymore, the class and the `instanceof` guards in `HasQueuedResponseCallbacks` have been removed.

### ⚠️ Behavior change

Previously, faking a capability meant `queue()` dispatched nothing. Queued fakes now dispatch the real job (the fake intercepts at the gateway layer), which changes test behavior in two ways:

1. **`then()` / `catch()` callbacks now execute** under fakes on a sync queue. Tests that relied on those callbacks *not* running (e.g. callbacks that mutate state) may need updating.
2. **Test suites where `QUEUE_CONNECTION` is not `sync`** will now push real jobs to the configured queue where previously nothing was dispatched.

To restore the old behavior in either case, combine the capability fake with `Queue::fake()`.


pushpak1300:
AI-REVIEW: _Automated review — Claude Opus 4.8 running `/code-review` (high) plus a multi-axis pass. The branch had merge conflicts with `0.x` (multimodal embeddings) which I resolved and pushed; a resulting test regression is fixed in the commit noted below._

The core change is **sound**: `then()`/`catch()` only register on a real `PendingDispatch` (they skip `FakePendingDispatch` in `HasQueuedResponseCallbacks`), so the callback now correctly fires on the sync connection and is correctly skipped under `Queue::fake()`. The audio/image/transcription/embeddings `then()`-callback tests all pass.

**Was blocking — now fixed (pushed to this branch)**

- The new `Ai::embeddingsAreFaked() ||` cache guard in `generateFromCache()` / `cacheEmbeddings()` (`src/PendingResponses/PendingEmbeddingsGeneration.php`) over-reached: it bypassed the cache for **all** faked embeddings, including explicit `->cache(60)`. After merging `0.x`, that broke two pre-existing tests that pass on `0.x` — `EmbeddingsFakeTest` "cached embeddings with media inputs use content hashes" (expected 2 calls, got 3) and "cached remote embeddings do not fetch remote metadata" (expected 1, got 2). Your own `EmbeddingsCacheTest` "faked embeddings bypass the cache" (global-config caching) and those two tests (explicit `->cache()`) route through the same `shouldCache()` path with opposite expectations.
  - **Fix pushed:** scoped the bypass to config-driven caching only — `(Ai::embeddingsAreFaked() && is_null($this->shouldCache)) || ! $this->shouldCache()`. A fresh fake still bypasses the global-config cache (your test stays green), but explicit `->cache()` under a fake now caches again (the two `0.x` tests pass). Full suite: **1889 passed, 0 failed** locally.

**Should-fix / decide** (behavior changes inherent to making `then()` fire — no other tests broke, but worth documenting)

1. Removing the early `return new Queued*Response(new FakePendingDispatch)` means a faked `queue()` now dispatches the real job, which runs `generate()` immediately on the sync connection. So a faked `queue()` now also records a synchronous generation and fires `Generating*`/`*Generated` events (via the `Generates*` concerns, e.g. `src/Providers/Concerns/GeneratesAudio.php:31`). Net: `assert*Generated` now passes after a faked `queue()`, and `assertNo*Generated` would now fail. Intended? Worth a line in the upgrade notes.
2. The fake-response queue is now consumed by the queued path (advances `currentResponseIndex`). A test that fakes `queue()` then `generate()` gets the *second* fake, not the first; and `preventStray*()` throws if a queued fake has no configured response.

**Nits** — none. AGENTS.md/CLAUDE.md compliance clean (docblocks and comments fine).

taylorotwell:
I think this is good just need to fix conflicts.

PR #801, opened 2026-07-18
Optimize object branch resolution in SchemaNormalizer::mergeUnion()

awesomecoderdev:
### Summary

`mergeUnion()` previously determined whether a union's resolved branches contained object variants by calling `mergeObjectVariants()`, which filtered `$resolved` for object branches. When that returned `null` (fewer than two object branches), the fallback path called `firstObjectBranch()`, which filtered the same `$resolved` array a second time using the identical predicate.

This change resolves object branches once and reuses that result for both the count check and the fallback selection, removing the duplicate pass.

### Changes

- Filter `$resolved` for object branches a single time in `mergeUnion()`.
- Reuse the filtered result for the `count($objects) >= 2` check that decides whether to merge multiple object variants.
- Reuse the same result for fallback branch selection when fewer than two object branches are present.
- Remove `mergeObjectVariants()` and `firstObjectBranch()`, whose logic is now inlined.

### Behavior

No functional changes. Output is identical for all union shapes — scalar unions, single-object unions, multi-object unions, and mixed scalar/object unions — since the merge and fallback logic is preserved exactly, only de-duplicated.

### Testing

- Existing test suite passes.
- Verified fallback paths for 0-object and 1-object resolved branches produce the same output as before the change.

### DIFF
--- a/src/Schema/SchemaNormalizer.php
+++ b/src/Schema/SchemaNormalizer.php
@@ -10,20 +10,50 @@ class SchemaNormalizer
      * Type-specific constraint keywords the deserializer rejects on a multi-type union.
      */
     private const TYPE_KEYWORDS = [
-        'minLength', 'maxLength', 'pattern', 'format',
-        'minimum', 'maximum', 'multipleOf',
-        'items', 'minItems', 'maxItems', 'uniqueItems',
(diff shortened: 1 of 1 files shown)


pushpak1300:
Thanks for your pull request!

PR #804, opened 2026-07-19
Add custom HTTP headers for agents and generation builders

### LINKED ISSUE #290: Support container (Claude Skills) and per-request anthropic-beta headers
### Description
 Two related gaps in the Anthropic provider prevent using Claude-hosted Skills:

  1. The `container` field is not forwarded to the Anthropic API from `providerOptions`
  2. The `anthropic-beta` header can only be set globally in config, not per-request via `providerOptions`
  Any other keys (including `container`) are silently dropped and never sent to Anthropic's API.

 Claude Skills require **both** — the `container.skills` payload field **and** the beta headers `code-execution-2025-08-25,skills-2025-10-02` on the same request.

  ### Problem 1: `container` is silently dropped

  `buildHttpRequestPayload` in `Providers/Anthropic/Handlers/Text.php` only maps a fixed set of keys
  from `providerOptions`. The `container` key is never included in the outgoing request body.

###  Problem 2: anthropic-beta header is global only
                                                                                                                                                         
  Anthropic.php reads betaFeatures from the constructor (set in config) and passes it as
  anthropic-beta. There is no way to append or override beta flags at request time.                                                                      
                                                                                                                                                         
  This is a problem because beta features are often capability-specific (e.g. skills, computer use,                                                      
  code execution) and should be enabled per-agent, not globally for all requests to the provider.

  ---                                                                                                                                                    
  Use Case                                                        
          
  Using a Claude-hosted skill with a Laravel AI agent via HasProviderOptions:
                                                                                                                                                         
```
  {                                                                                                                                                      
      return match ($provider) {                                  
          Lab::Anthropic => [                                                                                                                            
              'beta' => 'code-execution-2025-08-25,skills-2025-10-02',
              'container' => [                                                                                                                           
                  'skills' => [                                                                                                                          
                      [
                          'type'     => 'custom',                                                                                                        
                          'skill_id' => 'sk-...',                 
                          'version'  => 'latest',                                                                                                        
                      ],
                  ],                                                                                                                                     
              ],                                                  
          ],
          default => [],
      };                                                                                                                                                 
  }
```
                                                                                                                                                         
  Anthropic API Reference                                                                                                                                
   
  - Using Agent Skills with the API: https://platform.claude.com/docs/en/build-with-claude/skills-guide

emaadali:
There's currently no way to attach custom HTTP headers to provider requests. That makes it hard to sit the SDK behind AI proxies/gateways that authenticate or tag traffic via headers (for example [Cloudflare AI Gateway](https://developers.cloudflare.com/ai-gateway/)), which uses headers for gateway auth and request metadata/cost tags.

This PR adds opt in header support on the same shapes already used for provider options:

- Agents: `HasHeaders` with `headers(Lab|string $provider): array`
- Builders: `withHeaders(array|Closure)` on embeddings, transcription, image, audio, and reranking

Headers are sent on the HTTP request (or Bedrock `@http`) and are never merged into the JSON body.

```php
class ResearchAgent implements Agent, HasHeaders
{
    use Promptable;

    public function headers(Lab|string $provider): array
    {
        return [
            'cf-aig-authorization' => 'Bearer '.$this->gatewayToken,
            'cf-aig-metadata' => json_encode(['tenant' => $this->tenantId]),
        ];
    }
}
```

```php
Embeddings::for(['Hello'])
    ->withHeaders([
        'cf-aig-authorization' => 'Bearer '.$token,
        'cf-aig-metadata' => json_encode(['feature' => 'search']),
    ])
    ->generate();
```

```php
Image::of('A red apple')
    ->withHeaders(fn (Provider $provider) => [
        'X-Custom-Header' => $provider->driver(),
    ])
    ->generate();
```

Notes
- Closures receive the resolved Provider, matching withProviderOptions
- Queued prompts store array headers only (closures are dropped), same as queued provider options
- Embeddings cache keys include a headers fingerprint so different header sets do not collide


pushpak1300:
like the #838 approach tackling it over there.

emaadali:
@pushpak1300 The approach in #838 only supports headers configured via the config file, but I need dynamic headers (as does anyone who is going to be using the Cloudflare AI Gateway) which is why my implementation allowed users to configure headers as methods inside each agent class. I'll just use my fork of the AI SDK but I wanted to leave that bit of feedback for you.

emaadali:
@pushpak1300 I see you reopened this PR, do you want me to resolve conflicts?

pushpak1300:
Yes @emaadali

emaadali:
@pushpak1300 Resolved the conflicts against latest 0.x.

pushpak1300:
Trackign it into #904

PR #805, opened 2026-07-20
Add searchable tool catalogs

pushpak1300:
Every tool given to an agent normally ships its full definition to the provider on every request, so large catalogs spend context on tools the model may never call.

This PR adds `ToolSearch::for(...)`, which keeps a catalog of tools out of the provider payload and exposes it through two meta-tools instead, mirroring the searchable tool catalogs in laravel/mcp#317:

```php
{
    return [
        new SendEmail,
        ToolSearch::for([
            'orders' => [new LookupOrder, new FetchShipment],
            'github' => Mcp::client('github')->tools(),
            new SearchDocs,
        ])->maxToolCalls(25)->maxOutputBytes(65_536),
    ];
}
```

String keys group their tools: iterable values flatten into the catalog under that group, group names are listed in the `search_tools` description so the model knows the domains upfront, returned on each search result, and ranked alongside tool names — searching "github" surfaces that group's tools first. Ungrouped entries work as before.

The catalog is represented to the provider by `search_tools` and `execute_tools`:

- `search_tools` ranks the catalog by lexical matches over group names, tool names, descriptions, and nested schema fields — whole-token matches outrank substring matches — returning exact names, descriptions, and complete argument JSON Schemas. An empty query browses the catalog. Results are capped by both item count and response bytes.
- `execute_tools` runs multiple catalog calls synchronously in order, stopping on the first error so later calls never run against a failed state — a batch of lookups costs one provider round trip. When a call depends on a previous result, the model calls `execute_tools` again.

Nested executions flow through the generation loop, so `InvokingTool` and `ToolInvoked` events fire per catalog call, streamed responses emit `ToolResultEvent`s for nested results, and the response transcript records nested tool calls and results with stable `parent:index` tool call IDs. The loop snapshots and restores its nested-call records around every generation, so a catalog tool that prompts another agent on the same provider cannot clobber the parent transcript.

Execution enforces a configurable per-invocation call limit and a byte cap on every model-facing result. Approval-gated tools are rejected when the catalog is built — pass those to the agent directly. Duplicate tool names within a catalog, and provider-name collisions across expanded tools, fail fast with actionable exceptions.


taylorotwell:
Going to skip this one and just have provider hosted.

PR #806, opened 2026-07-21
[0.x] Update .gitattributes

jackbayliss:
This adds the below to export ignore 
- /workbench -
- `phpstan.neon.dist`
-  `rector.php`
- `testbench.yaml`

We don't need them exported:
Screenshot 2026-07-21 at 17 30 24

This means, end users dont end up with them in their vendor..
Saves a bunch of data cos the workbench folder 

HYB


pushpak1300:
Thanks. HYB!

Issue #170, opened 2026-02-15
Support for thinking output


rajabilal555:
Do we have a plan to support thinking output in the future?

Like the toolcalls and citations there is also thinking_delta and such for models that support thinking output

GregPeden:
I agree, I tried to get it to export thinking activity but I couldn't figure it out (using openAI agents).

meirdick:
For what it's worth, Prism already supports this at the provider level — the `additionalContent` field on `Prism\Prism\Text\Step` can carry thinking/reasoning data, and the xAI provider already extracts `reasoning_content` into streaming events (`ThinkingStartEvent`, `ThinkingEvent`, `ThinkingCompleteEvent`).

I've implemented the same pattern in [meirdick/prism-workers-ai](https://github.com/meirdick/prism-workers-ai) v0.3.0 for Workers AI reasoning models (e.g. Kimi K2.5). Non-streaming responses surface thinking in `$response->steps[0]->additionalContent['thinking']`, and streaming emits the standard Prism thinking events.

The gap is that `Laravel\Ai\Responses\Data\Step` doesn't expose `additionalContent` from the underlying Prism step, so the thinking data is lost when going through `agent()->prompt()`. The data is there at the Prism layer — it just needs to be passed through in the SDK's response wrapper.

mikescott:
Similar problem with web search. The results appear to be available in `additionalContent` but that is not persisted even with `RemembersConversations` - this results in the agent repeating web searches in subsequent conversation turns because it doesn't have the results in its context.

GregPeden:
I did get thinking output to work, you just have to specify a reasoning level. The agent companies don't provide the reasoning messages unless you specify that.

meirdick:
@GregPeden Good point on needing to specify a reasoning level — we just shipped [v0.4.0](https://github.com/meirdick/prism-workers-ai/releases/tag/v0.4.0) of `meirdick/prism-workers-ai` which now forwards provider options (like `reasoning_effort`) directly into the API request payload.

```php
Prism::text()
    ->using('workers-ai', 'workers-ai/@cf/moonshotai/kimi-k2.5')
    ->withPrompt('Explain quantum entanglement')
    ->withProviderOptions(['reasoning_effort' => 'high'])
    ->asText();
```

So the full reasoning flow is now:
- **Request side**: `withProviderOptions(['reasoning_effort' => 'high'])` gets forwarded to the Workers AI payload (v0.4.0)
[...]

PR #813, opened 2026-07-23
Emit and replay reasoning on the Chat Completions providers

helitik:
The OpenRouter streaming gateway reads `delta.content` but never looks at the reasoning fields, so when a model returns a chain-of-thought the reasoning text is silently dropped — even though its `reasoning_tokens` are already counted in `Usage`. The `Reasoning*` stream events (and their Vercel `reasoning-start`/`-delta`/`-end` parts) already exist and are emitted by the DeepSeek, Gemini, xAI, Anthropic, OpenAI and Bedrock gateways; OpenRouter just wasn't wired up.

This emits `ReasoningStart` / `ReasoningDelta` / `ReasoningEnd` while streaming, and replays the captured reasoning on follow-up requests so a tool loop doesn't lose the thinking that produced its tool calls. When a stream carries no reasoning fields the output is identical to before, so nothing changes for existing code.

## Reading reasoning

Reasoning is read from whichever field the upstream uses: `reasoning` (OpenRouter's native field, see the [reasoning tokens docs](https://openrouter.ai/docs/use-cases/reasoning-tokens)), `reasoning_content` (the alias used by OpenAI-compatible upstreams such as LiteLLM and vLLM), or the structured `reasoning_details` array that models like Claude return. Only one source is read per stream, since OpenRouter mirrors the same text into more than one of them.

The block state machine mirrors the one added for DeepSeek in #534: the block opens on the first reasoning delta and closes as soon as answer text or tool calls begin — or when the stream ends. It now lives in a shared `ChatCompletionReasoning` class rather than inline in each gateway, which also fixes a DeepSeek ordering bug along the way: a chunk carrying both the last reasoning token and the first answer token used to close the block before consuming its own reasoning, opening a second block for one token.

## Replaying reasoning

[OpenRouter accepts three inbound shapes](https://openrouter.ai/docs/use-cases/reasoning-tokens): `reasoning_details` verbatim, or the plain text as either `reasoning` or its documented alias `reasoning_content`. `reasoning_details` is preferred because it carries the signatures and encrypted payloads that thinking models require back unmodified — flattening those to a plain string is exactly the modification the docs warn against:

> the entire sequence of consecutive reasoning blocks must match the outputs generated by the model during the original request; you cannot rearrange or modify the sequence of these blocks

Because the whole sequence has to match, a reasoning block that reached us without its signature invalidates every block beside it, not just itself — so a sequence containing one falls back to replaying the plain text instead of being sent with a gap in it. That covers both Anthropic, which rejects unsigned thinking outright, and Gemini, which reports a corrupted thought signature.

An empty `reasoning_details` array is replayed as an empty array rather than dropped: DeepSeek V4 answers with one on turns it produced no visible reasoning for, and requires it back.

Detail fragments streamed under one block identity are merged back into a single block, concatenating in the order OpenRouter documents, so an incremental upstream ends up with one replayable block and counts its reasoning once. Fragments carrying neither an `id` nor an `index` are treated as separate blocks, since there is nothing to merge them on. Encrypted blocks emit no `ReasoningDelta` — there is nothing to render — but are still captured so they survive the round trip.

Reasoning is only replayed alongside tool calls, never to pad a finished turn, and another provider's replay state is already stripped upstream by `withoutForeignProviderContentBlocks()`.

## Scope

Groq and the configurable OpenAI-compatible gateway share the same parser and mapper, so they capture and replay reasoning here too rather than inheriting the shared code as dead weight.

Which field the plain text is replayed under is recorded at capture time and reused, rather than fixed per gateway. That is what Groq needs, since it answers on `reasoning`, and it is what [current vLLM](https://docs.vllm.ai/en/latest/features/reasoning_outputs/) needs too — it renamed `reasoning_content` to `reasoning` and documents the old name as legacy. Answering on the field the upstream used avoids guessing which generation of an OpenAI-compatible server is on the other end.

DeepSeek is a special case: [its thinking mode](https://api-docs.deepseek.com/guides/thinking_mode/) returns a 400 unless the reasoning of every prior assistant turn comes back once a request carries `tools` — including turns that made no tool call, and including empty reasoning. Its mapper now replays on every assistant message when tools are present, and on tool-call messages regardless.

Mistral is untouched: it shares the message mapper but has its own response parser and stream handler, so it captures no reasoning and the replay is a no-op for it. Ollama is left out as well: it uses its own message mapper and a differently shaped `thinking` field.

Note that `providerContentBlocks` on `StepResponse`, `AssistantMessage` and `ToolApprovalRequest` is now documented as `array<array-key, mixed>` rather than `array<int, array<string, mixed>>`. Anthropic and Bedrock store a list of raw blocks while the Chat Completions providers store a keyed map, and the docblocks now say so.

Contributes to #170.

### DIFF
--- a/src/Gateway/Bedrock/BedrockTextGateway.php
+++ b/src/Gateway/Bedrock/BedrockTextGateway.php
@@ -687,7 +687,7 @@ class BedrockTextGateway implements EmbeddingGateway, StepTextGateway
      * Build the assistant conversation message block combining text and tool calls.
      *
      * @param  array<ToolCall>  $toolCalls
-     * @param  array<int, array<string, mixed>>  $providerContentBlocks
+     * @param  array<array-key, mixed>  $providerContentBlocks
      */
     protected function buildAssistantConversationMessage(string $text, array $toolCalls, array $providerContentBlocks = []): array
     {
--- a/src/Gateway/DeepSeek/Concerns/HandlesTextStreaming.php
+++ b/src/Gateway/DeepSeek/Concerns/HandlesTextStreaming.php
@@ -4,15 +4,13 @@ namespace Laravel\Ai\Gateway\DeepSeek\Concerns;
 
 use Generator;
 use Illuminate\Support\Str;
+use Laravel\Ai\Gateway\OpenAiCompatible\ChatCompletionReasoning;
 use Laravel\Ai\Gateway\StepResponse;
 use Laravel\Ai\Providers\Provider;
 use Laravel\Ai\Responses\Data\Meta;
 use Laravel\Ai\Responses\Data\ToolCall;
 use Laravel\Ai\Responses\Data\Usage;
 use Laravel\Ai\Streaming\Events\Error;
-use Laravel\Ai\Streaming\Events\ReasoningDelta;
-use Laravel\Ai\Streaming\Events\ReasoningEnd;
-use Laravel\Ai\Streaming\Events\ReasoningStart;
 use Laravel\Ai\Streaming\Events\StreamEvent;
 use Laravel\Ai\Streaming\Events\StreamStart;
 use Laravel\Ai\Streaming\Events\TextDelta;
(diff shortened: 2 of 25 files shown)


pushpak1300:
Solving this in #1007

PR #821, opened 2026-07-25
Fix Gemini file import operation handling

### LINKED ISSUE #817: Gemini addFile treats importFile Operation as document id
https://github.com/laravel/ai/blob/18d574e859d3eaebb73244c4774f1ecce1e59fb8/src/Gateway/Gemini/GeminiStoreGateway.php#L75

## Finding

`GeminiStoreGateway::addFile` POSTs to `fileSearchStores/{store}:importFile` and returns `basename($response->json('name'))` without polling. Gemini’s File Search API returns a long-running **Operation** (`name` like `fileSearchStores/{store}/operations/{opId}`, with `done` false until indexing finishes), not a Document resource. The package never calls operations.get, never waits for `done`, and never reads the completed operation’s document name. Callers therefore receive an operation id, treat it as a document id, and later `removeFile` builds `.../documents/{opId}` which is the wrong resource. `createStore` with initial `fileIds` uses the same broken path.

## Violated Invariant Or Contract

`StoreGateway::addFile` must return a document identity usable with `removeFile` / store membership. Returning an unfinished operation name violates the gateway contract and the dual-id model used by `Store::add`.

## Oracle

- Gemini API: `fileSearchStores.importFile` response body is an `Operation`; clients poll until `done`.
- Official samples: `while not operation.done: operation = client.operations.get(operation)`.
- Package unit tests mock a document-shaped `name` (`.../documents/doc456`), which does not match the live Operation contract.
- `removeFile` deletes `.../{documentId}` under the store’s documents collection.

## Counterexample

```php
$store = Stores::create('docs', provider: 'gemini');
$doc = $store->add(Document::fromPath('manual.pdf'));
// $doc->id() is operation basename, not documents/...
$store->remove($doc); // DELETE .../documents/{operationId} → wrong target
// Immediate file search may miss still-indexing content
```

## Why It Might Matter

Store add/remove lifecycle is broken for the Gemini provider: apps cannot reliably track or delete documents; search may run against incomplete imports; resource cleanup fails. Distinct from the generic `Store::remove(deleteFile)` file-id issue (document vs file service id).

## Proof

- Control-flow: importFile → return basename(name) with no poll.
- Contract mismatch: Operation vs Document per Gemini File Search API.
- Cross-entry: tests mock document names; production returns operations.

## Counterevidence Checked

- If a future API returned document names synchronously, basename would work—but documented contract is Operation.
- Strongest false-positive reason: “SDK users always poll themselves.” Counter: the gateway hides HTTP and returns a string typed as document id with no operation handle exposed.

## Suggested Next Step

Poll the operation until `done`, then return the document resource name from the completed response (or error if failed). Surface pending state if the public API should remain async.

drewmt:
`fileSearchStores.importFile` returns a long-running operation, but the Gemini store gateway currently treats that operation name as the imported document ID. This makes `Store::add()` return an operation ID and causes later document operations to target the wrong resource.

This change waits for the import operation to finish, surfaces operation failures and invalid responses, and returns the completed `response.documentName`. Polling is bounded and uses Laravel's sleep utility so the behavior remains testable.

Fixes #817

### Testing

- `vendor/bin/pest tests/Feature/Providers/Gemini/StoreGatewayTest.php --compact` — 15 passed, 32 assertions
- `php -d memory_limit=-1 vendor/bin/pest --compact --testsuite=Unit,Feature` — 1,883 passed, 24 skipped, 3,928 assertions
- `vendor/bin/pint --test src/Gateway/Gemini/GeminiStoreGateway.php tests/Feature/Providers/Gemini/StoreGatewayTest.php`
- `vendor/bin/phpstan analyse --memory-limit=-1`

### DIFF
--- a/src/Gateway/Gemini/GeminiStoreGateway.php
+++ b/src/Gateway/Gemini/GeminiStoreGateway.php
@@ -4,9 +4,12 @@ namespace Laravel\Ai\Gateway\Gemini;
 
 use DateInterval;
 use Illuminate\Http\Client\PendingRequest;
+use Illuminate\Http\Client\Response;
 use Illuminate\Support\Collection;
+use Illuminate\Support\Sleep;
 use Laravel\Ai\Contracts\Gateway\StoreGateway;
 use Laravel\Ai\Contracts\Providers\StoreProvider;
+use Laravel\Ai\Exceptions\AiException;
 use Laravel\Ai\Gateway\Concerns\CreatesClient;
 use Laravel\Ai\Gateway\Concerns\HandlesFailoverErrors;
 use Laravel\Ai\Providers\Provider;
@@ -17,6 +20,10 @@ class GeminiStoreGateway implements StoreGateway
 {
     use CreatesClient;
     use HandlesFailoverErrors;
+
+    protected const IMPORT_OPERATION_MAX_ATTEMPTS = 60;
+
+    protected const IMPORT_OPERATION_POLL_INTERVAL_SECONDS = 5;
 
     /**
      * Get a vector store by its ID.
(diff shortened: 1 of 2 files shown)


saifulferoz:
I went through this against #817 and it addresses all four points in the report: `addFile` now returns a real document id, `removeFile` therefore targets `.../documents/{documentId}` correctly, and `createStore` is covered too since it delegates to `addFile` per file id.

I checked out the branch and ran it: 15 passed on `StoreGatewayTest`, and 2152 passed / 24 skipped on `--testsuite=Unit,Feature`, matching your numbers. Reverting just `GeminiStoreGateway.php` and keeping the tests fails 4 of them, so the new cases do pin the actual behaviour rather than only documenting it. The branch still applies cleanly to `0.x` at 2142b6e.

One thing worth adding to the record, since every test here mocks the payload: the `response.documentName` path is confirmed by the official SDK, not just inferred. In `googleapis/python-genai`, `_ImportFileOperation_from_mldev` delegates to `_ImportFileResponse_from_mldev`, which reads `documentName` nested under `response` and maps it to `document_name`. The REST reference documents the Operation envelope (`name` / `done` / `error` / `response`) but not the contents of `response`, so the SDK is the only public confirmation I could find that the field name is right. It is.

Two things I would want a maintainer to weigh before this merges.

**1. `addFile` becomes a blocking call of up to five minutes.**

`IMPORT_OPERATION_MAX_ATTEMPTS = 60` at a five second interval means a single `Store::add()` can now occupy a request or worker for five minutes where it previously returned in one round trip. There is no other polling loop in `src/`, so this is the first place the package waits on a provider, and `OpenAiStoreGateway::addFile` returns `$response->json('id')` immediately without waiting for indexing.

That asymmetry is not obviously wrong — OpenAI's endpoint hands back a real id synchronously and Gemini's genuinely does not, so something has to give. But it does mean the same interface method now has very different timing characteristics per provider, and the store already models readiness separately (`ready:` on `Store`, `pendingDocumentsCount` on `StoreFileCounts`). The issue raised the same fork explicitly: "Surface pending state if the public API should remain async."

So the question is whether the gateway should block, or return the operation and let the caller resolve it. Blocking is the smaller change and makes the return value honest, which is probably the right call for now — but it is a behavioural change to a public method that is worth being deliberate about rather than arriving at by way of a bug fix. Might be worth a line in the docs either way.

**2. The timeout error drops the operation name.**

```php
throw new AiException('Gemini file import operation timed out.');
```

When this fires the import is still running server side and will most likely produce a document. The caller has no handle to it: no operation name, no document id, so the document cannot be found or deleted later. Including `$operationName` in the message would make that recoverable, and costs a `sprintf`.

Neither of these blocks the fix as far as I am concerned — the current behaviour returns an operation id where a document id is documented, which is worse than either option here.

PR #822, opened 2026-07-25
Support configured OpenRouter plugins

### LINKED ISSUE #776: [bug] OpenRouter gateway: support provider-level plugins configuration
**Problem:** I was trying to attach files to an OpenRouter Request using the file url method. The OpenRouter documentations states that it supports request-level plugins in my case,  file-parser for PDF processing via mistral-ocr, but it seems that the plugin iformation was ignored causing a 500 error.

**Example:** The following code will return a 500 error.
Example code:
```
$attachments = collect($this->queuedFiles)
            ->map(fn ($file) => Document::fromUrl($file['url'])->as($file['name']))
            ->all();

...
response = $agent->forUser($user)->prompt($prompt, $attachments);
```

**Proposed fix:** In file vendor/laravel/ai/src/Gateway/OpenRouter/Concerns/BuildsTextRequests.php, BuildsTextRequests::buildTextRequestBody(), after the existing providerOptions merge (line ~54), read plugins from the provider's additional configuration and append it to the body:

```
$plugins = $provider->additionalConfiguration()['plugins'] ?? null;

if (filled($plugins)) {
    $body['plugins'] = $plugins;
}
```
This allows the ai.php provider config to have the plugins array:
```
'openrouter' => [
    'driver' => 'openrouter',
    'key' => env('OPENROUTER_API_KEY'),
    'plugins' => [
        ['id' => 'file-parser', 'pdf' => ['engine' => 'mistral-ocr']],
    ],
],
```

I have tested this fix and works for me. I can now pass files to the provider as URL's instead of embedded in the context.

I am using Laravel AI v0.8.1

drewmt:
OpenRouter accepts a top-level `plugins` array for features such as PDF parsing, but plugins configured under `ai.providers.openrouter` are currently ignored. This prevents applications from defining a shared plugin policy for all agents using that provider.

This change includes non-empty configured plugins in text requests while keeping per-agent provider options authoritative, so an agent can still override the global default for a specific request.

Fixes #776

### Testing

- Failing-first regression confirmed configured plugins were absent before the change
- OpenRouter provider suite: 137 passed, 226 assertions
- Unit and Feature suites: 1,882 passed, 24 skipped, 3,921 assertions
- Pint and PHPStan pass

### DIFF
--- a/src/Gateway/OpenRouter/Concerns/BuildsTextRequests.php
+++ b/src/Gateway/OpenRouter/Concerns/BuildsTextRequests.php
@@ -68,6 +68,12 @@ trait BuildsTextRequests
             'top_p' => $options?->topP,
         ]));
 
+        $plugins = $provider->additionalConfiguration()['plugins'] ?? null;
+
+        if (filled($plugins)) {
+            $body['plugins'] = $plugins;
+        }
+
         $providerOptions = $options?->providerOptions($provider->driver());
 
         if (filled($providerOptions)) {
--- a/tests/Feature/Providers/OpenRouter/ProviderOptionsTest.php
+++ b/tests/Feature/Providers/OpenRouter/ProviderOptionsTest.php
@@ -31,6 +31,23 @@ test('provider options are included in openrouter request body', function (): vo
     });
 });
 
+test('provider plugins configuration is included in openrouter request body', function (): void {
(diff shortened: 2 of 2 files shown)


pushpak1300:
Closing because of this https://github.com/laravel/ai/issues/776#issuecomment-5077666415

PR #823, opened 2026-07-25
Expose the persisted assistant message ID on AgentResponse

tristan-ch:
When conversation persistence is enabled, `RememberConversation` stores the assistant
message via `ConversationStore::storeAssistantMessage()`, which returns the ID of the
newly persisted message — but the middleware currently discards it. The response
exposes `conversationId`, yet not the ID of the message that was just written.

Applications that want to attach additional data to that exact message after the turn
(citations/sources, attachments, scores, metadata) currently have to guess "the latest
assistant message in the conversation" or subclass the conversation store just to
intercept the return value.

This PR keeps the ID and passes it through `withinConversation()` into a new nullable
`assistantMessageId` property on `AgentResponse` (inherited by `StreamedAgentResponse`,
and propagated through `StreamableAgentResponse` for the streaming path):

    // Non-streamed
    $response = $agent->prompt('...');
    $response->assistantMessageId; // ID of the assistant message just persisted, or null

    // Streamed
    $stream = $agent->stream('...');
    foreach ($stream as $event) { /* ... */ }
    $stream->assistantMessageId;   // same, once the stream has been consumed

`null` remains a legitimate state: when persistence is not enabled, or when the store
itself returns `null` (approval-resume turns that produce no text, no tool calls, and
no new tool results insert no row).

Fully backward compatible: optional trailing parameters, nullable properties, no
behavior change anywhere else. Covered by new and extended tests in
`AgentMiddlewareTest` (non-streamed, streamed, not-remembered) and
`AgentStreamFailoverTest` (ID survives failover and `then()` callbacks).

### DIFF
--- a/resources/boost/skills/ai-sdk-development/SKILL.md
+++ b/resources/boost/skills/ai-sdk-development/SKILL.md
@@ -143,6 +143,7 @@ class SalesCoach implements Agent, Conversational
 // Start a new conversation
 $response = (new SalesCoach)->forUser($user)->prompt('Hello!');
 $conversationId = $response->conversationId;
+$assistantMessageId = $response->assistantMessageId;
 
 // Continue an existing conversation
 $response = (new SalesCoach)->continue($conversationId, as: $user)->prompt('Tell me more.');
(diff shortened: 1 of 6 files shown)


pushpak1300:
close becuase of #961

PR #828, opened 2026-07-26
Preserve input boundaries in embeddings cache key

### LINKED ISSUE #826: Embedding cache keys collide for distinct string batches
## Environment

- `laravel/ai` v0.10.1 (`47f0ffdb86639f710eb93fcdf1bd07635599f270`)

## Problem

`PendingEmbeddingsGeneration::cacheKey()` builds the key for all-string inputs with:

```php
implode("-", $this->inputs)
```

The input boundaries are not preserved, so distinct batches can share a cache entry:

```php
["a-b", "c"]
["a", "b-c"]
```

Both produce the same key material. A cached response for the first batch can therefore be returned for the second batch, yielding incorrect embeddings.

## Expected behavior

The key should encode the ordered input list unambiguously, for example by hashing a JSON structure that contains the inputs array alongside provider, model, dimensions, and options.

The non-string input branch already uses structured JSON and does not have this ambiguity.

lazerg:
The all-string branch of `cacheKey()` joined inputs with `implode('-', ...)`, which loses the boundaries between items. Batches like `['a-b', 'c']` and `['a', 'b-c']` produced the same key, so a cached response for one could be handed back for the other. Encoding the inputs as JSON keeps each item distinct.

Fixes #826

### DIFF
--- a/src/PendingResponses/PendingEmbeddingsGeneration.php
+++ b/src/PendingResponses/PendingEmbeddingsGeneration.php
@@ -213,7 +213,7 @@ class PendingEmbeddingsGeneration
         if (array_all($this->inputs, fn ($input) => is_string($input))) {
             return 'laravel-embeddings:'.hash(
                 'sha256',
-                $provider->driver().'-'.$model.'-'.$dimensions.'-'.$optionsFingerprint.'-'.implode('-', $this->inputs),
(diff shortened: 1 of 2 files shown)


lazerg:
Hey @pushpak1300, take a look when you have time.

pushpak1300:
closing because of #832

Issue #816, opened 2026-07-24
Gemini sanitizeRequestParts drops thoughtSignature required for tool loops


bnomei:
https://github.com/laravel/ai/blob/18d574e859d3eaebb73244c4774f1ecce1e59fb8/src/Gateway/Gemini/Concerns/ParsesTextResponses.php#L74

## Finding

When parsing Gemini text steps, `providerContentBlocks` are set via `sanitizeRequestParts(excludeThinkingParts($parts))`. `sanitizeRequestParts` rebuilds every `functionCall` with only `name` and optional `args`, dropping `thoughtSignature` (and other fields such as `id`). On the next tool-loop step, `MapsMessages::mapAssistantMessage` replays those blocks verbatim. Gemini 3 / thinking models require the signature on function-call history for multi-turn tool use; omitting it yields HTTP 400 (`Function call is missing a thought_signature in functionCall parts`) or degraded tool continuity. Default text model is `gemini-3.6-flash`.

## Violated Invariant Or Contract

Provider content blocks meant for exact replay must preserve fields the provider requires on subsequent turns. Gemini documents thought signatures as mandatory for function-calling multi-turn on Gemini 3 (and important for 2.5).

## Oracle

- Official Gemini thought-signature docs: function calling requires echoing signatures; missing → 400 on Gemini 3.
- Code comment on `sanitizeRequestParts`: “so they can be sent back to Gemini as conversation history.”
- Default model `GeminiProvider::defaultTextModel()` → `gemini-3.6-flash`.
- Stream path uses the same sanitizer (`HandlesTextStreaming`).

## Counterexample

1. Agent with any tool, `provider: 'gemini'` (default model).
2. Model returns parts including `functionCall: { name, args, thoughtSignature: "..." }`.
3. Package stores sanitized blocks without `thoughtSignature`.
4. Loop sends tool results + model history → Gemini 400: missing thought_signature on functionCall parts.

## Why It Might Matter

Catastrophic availability for the default Gemini agent tool path: multi-step agents cannot complete tool use. This is not a rare edge model—the package default is a Gemini 3-era flash model.

## Proof

- Dataflow: response parts → sanitize strips signature → providerContentBlocks → next request contents.
- Contract mismatch: “replay history” vs field-stripping sanitizer.
- Counterexample: multi-step tool agent on default Gemini model.

## Counterevidence Checked

[...]

PR #834, opened 2026-07-30
Add Gemini tool-loop coverage for thought signature preservation

sulimanbenhalim:
Adds a couple of tests pinning that the `thoughtSignature` on a Gemini function-call part survives the tool-call continuation, on both the non-streaming and streaming paths.

Why it matters: on the `generateContent` API the signature comes back as a sibling of `functionCall`, and Gemini 3 rejects tool history whose functionCall part is missing it (`Function call ... is missing a thought_signature`). The default text model is `gemini-3.x`, so this is the default agent tool path, and it wasn't covered. The gateway already preserves it (`sanitizeRequestParts` only rebuilds the inner `functionCall` and keeps the rest of the part), these tests just lock that in so a future refactor can't quietly drop it.

Tests only, no behavior change. Both pass on `0.x`, and I checked they go red if the sanitizer stops preserving the sibling.

Came out of the discussion on #816.

### DIFF
--- a/tests/Feature/Providers/Gemini/StreamingTest.php
+++ b/tests/Feature/Providers/Gemini/StreamingTest.php
@@ -166,6 +166,49 @@ describe('tool calls', function (): void {
             }
         }
     });
+
+    test('streaming preserves the thought signature across the tool call continuation', function (): void {
+        Http::fake([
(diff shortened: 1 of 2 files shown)


pushpak1300:
Thanks

PR #835, opened 2026-07-30
Pass provider to the agent tools method

### LINKED ISSUE #281: Feat: providerTools to conditionally add provider tools
Case:
 - Agent with default Anthropic model, failover OpenAI
 - Tools include WebFetch
 - When using OpenAi, webfetch is not available -> crashes

So we can't use the failover, or we can't use the webfetch tool. Ideally we should use webfetch when available.

Possible solution:

Add `hasProviderTools` similar to `hasProviderOptions`, which adds a `providerTools` method, which adds tools based on the provider.

```php
/**
 * Get provider-specific tools.
 */
public function providerTools(Lab|string $provider): array
{
    return match ($provider) {
        Lab::Anthropic => [
            new WebFetch,
        ],
        default => [],
    }
}
```

Alternatively, add the `$provider` argument to the `tools()` option directly. (Edit: But this would be a BC break)

KeironLowePatchWorks:
Agents that fail over currently expose the same tools to every provider. A provider-specific tool such as Anthropic's `WebFetch` can therefore be sent to OpenAI during failover, causing the fallback request to fail too.

This passes the active provider to `tools()` so each provider receives only the tools it supports:

```php
use Laravel\Ai\Enums\Lab;

public function tools(Lab|string $provider): iterable
{
    return match ($provider) {
        Lab::Anthropic => [new SearchKnowledgeBase, new WebFetch],
        default => [new SearchKnowledgeBase],
    };
}
```

Tools are resolved again for every provider attempted during failover. Standard and aliased connections receive their provider `Lab`; named OpenAI-compatible connections receive the connection name so separate endpoints can expose different tools.

This changes the `HasTools` contract, so existing agents must add the `Lab|string $provider` parameter even when they don't need provider-specific tools. The upgrade guide and generated agent stubs document the new signature.

Tests cover provider selection, aliased connections, failover, and named OpenAI-compatible connections.

Fixes #281.

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,39 @@
 # Upgrade Guide
 
+## Upgrading To 1.0 From 0.x
+
+### Provider Aware Agent Tools
+
+**Likelihood Of Impact: High**
+
+The `HasTools` contract now receives the provider that is being invoked, allowing an agent to offer a different set of tools to each provider. Agents implementing `HasTools` should add the `$provider` argument to their `tools` method:
+
+```php
+use Laravel\Ai\Enums\Lab;
+
+/**
+ * Get the tools available to the agent for the given provider.
+ */
+public function tools(Lab|string $provider): iterable
+{
+    return [new SearchKnowledgeBase];
+}
+```
+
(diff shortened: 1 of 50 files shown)


pushpak1300:
Can you fix tests and fix merge conflicts ?

pushpak1300:
Thanks for your pull request!

PR #837, opened 2026-07-31
Preserve Gemini thought signatures across persisted tool conversations

sulimanbenhalim:
Fixes a Gemini 3 break for apps that use conversation memory with tools.

When a completed tool turn is persisted (`RemembersConversations`) and the conversation continues in a later request, the model's function call gets rebuilt from the stored `ToolCall` DTOs, which didn't carry the `thoughtSignature`. So the replayed history goes back without it, and Gemini 3 rejects that with `400 ... Function call is missing a thought_signature in functionCall parts`. The default text model is `gemini-3.x`, so it hits the default tool path on the second turn. The in-flight loop was already fine (it replays the raw provider blocks); only the persisted path dropped it.

The fix follows how OpenAI/xAI already carry `reasoning_encrypted_content` on the tool call. `ToolCall` gets a `thoughtSignature` field, the Gemini parser attaches the sibling signature to the call, and the mapper emits it back when it rebuilds a rehydrated function-call part. Storage round-trips it for free through the existing `toArray`/`fromArray`.

Added a persist-then-continue test that checks the signature is stored on the tool call and replayed on the next turn, a streaming test that pins the signature onto the emitted `ToolCall`, a streaming test that pins a signature-only part surviving into the replayed model parts, and a mapping test that pins the part-level replay. Other providers are unchanged, the new field just defaults to null.

Related to #816, which misdiagnoses the cause. `thoughtSignature` is a [Part-level field](https://ai.google.dev/api/generate-content), a sibling of `functionCall`, not a key inside it — `sanitizeRequestParts` only rewrites `$part['functionCall']` and leaves the signature intact. The loss is on this persisted-continuation path.

Still open, out of scope here: persisted text-only turns drop the final-part signature, and history rebuilt from the client — `Vercel::fromUiMessage()` and `AgentUserInteraction::toolCallsFrom()` — has no slot to carry the signature back. `excludeThinkingParts` dropping `thought` parts is fine: the rule that thought blocks must be resent unmodified is [documented for the Interactions API](https://ai.google.dev/gemini-api/docs/thinking), which `generateContent` has no equivalent of — there, signatures are part metadata, and the ones that matter ride the `functionCall` parts this PR preserves.

No `UPGRADE.md` entry needed: the new field is nullable and last, and `fromArray` defaults it, so stored rows rehydrate unchanged.


sulimanbenhalim:
synced with latest 0.x. only conflict was an overlapping gemini streaming test, kept both. green now.

PR #841, opened 2026-08-02
Add embeddings support to the OpenAI-compatible provider

### LINKED ISSUE #840: it is not possible to use embeddings with the openai-compatible driver
Hi,
get the error that embedding is not supported with the openai-compatible driver.
lgA

ademola-emmanuel:
Fixes #840.

The `openai-compatible` driver only implements the text contract, so calling `embeddings()` on it fails even though OpenAI-compatible servers such as LM Studio, vLLM, and LiteLLM expose the standard `/v1/embeddings` endpoint.

This PR wires the driver into the existing embeddings architecture:

- `OpenAiCompatibleProvider` now implements `EmbeddingProvider` through the shared `GeneratesEmbeddings` and `HasEmbeddingGateway` concerns, so caching, faking, and events work unchanged.
- `OpenAiCompatibleGateway` gains a `generateEmbeddings()` method that posts to the configured endpoint using the OpenAI wire format. It reuses the gateway's existing client (base URL, optional bearer token, failover error handling) and response validation, and mirrors the OpenRouter gateway's implementation.
- Since arbitrary endpoints have no known models, the default embeddings model must be configured, following the same pattern the driver already uses for `defaultTextModel()`. Dimensions are optional: when configured (or passed explicitly) they are sent with the request, and when omitted the request leaves out the `dimensions` parameter so the model's native dimensions are used. Backends that reject `dimensions` for fixed-dimension models never receive it unless the user opts in. This mirrors how the Ollama gateway filters the parameter.

```php
'openai-compatible' => [
    'url' => env('OPENAI_COMPATIBLE_URL'),
    'key' => env('OPENAI_COMPATIBLE_API_KEY'),
    'models' => [
        'embeddings' => [
            'default' => 'text-embedding-qwen3-embedding-0.6b',
            'dimensions' => 1024, // optional
        ],
    ],
],
```

Tests mirror the OpenRouter embeddings suite: request formatting, response parsing, dimensions sent when configured and omitted when not, auth header present and absent, provider options passthrough, missing-model exception, and rate limit, overload, and error-in-200 handling.

### DIFF
--- a/resources/boost/skills/ai-sdk-development/SKILL.md
+++ b/resources/boost/skills/ai-sdk-development/SKILL.md
@@ -407,23 +407,35 @@ $store->assertAdded('file_id');
 
 ## OpenAI-Compatible Provider
 
-Point the SDK at any OpenAI Chat Completions endpoint (LM Studio, vLLM, Together, etc.) with the config-driven `openai-compatible` driver. Define named instances in `config/ai.php`, no code required:
+Point the SDK at any OpenAI-compatible endpoint (LM Studio, vLLM, Together, etc.) with the config-driven `openai-compatible` driver. Define named instances in `config/ai.php`, no code required:
 
 ```php
 'my-llm' => [
     'driver' => 'openai-compatible',
     'url' => env('MY_LLM_URL'),        // required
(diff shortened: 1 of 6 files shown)


pushpak1300:
Thanks.

PR #843, opened 2026-08-02
Azure OpenAI - Add web search support

jesse-bos:
Azure's OpenAI Responses API serves the same `web_search` tool as OpenAI, so this implements `SupportsWebSearch` on `AzureOpenAiProvider` with the same option mapping, keyed to `Lab::Azure`.

Docs: [Web search with the Responses API](https://learn.microsoft.com/en-us/azure/foundry/openai/how-to/web-search)


pushpak1300:
Thanks

PR #845, opened 2026-08-03
Add Vercel protocol coverage for text and reasoning stream events

dfinchenko:
`VercelProtocolStreamTest` currently covers only the tool-related parts of the Vercel AI SDK protocol — tool input/output, approval requests, and the start / finish frames. The `toVercelProtocolArray()` mappings for the text and reasoning lifecycle events had no coverage at all, even though every streaming gateway emits them: `TextStart`/`TextEnd` come from the fake gateway plus the Gemini, Groq, OpenRouter and OpenAI-compatible streams, and `ReasoningStart`/`Delta`/`End` come from Anthropic, Gemini, Bedrock, DeepSeek and xAI.                                                                                                                                                                                
                                                                                                                                                                                                                                         
This adds two tests that assert the full frame sequence for each: a text stream emitting `text-start` / `text-delta` / `text-end`, and a reasoning stream emitting `reasoning-start` / `reasoning-delta` / `reasoning-end`. Both reuse the existing `vercelProtocolParts()` helper and follow the style of the surrounding tests.                                                                                                                                                          
                                                                                                                                                                                                                                         
Tests only, no source changes.

### DIFF
--- a/tests/Feature/VercelProtocolStreamTest.php
+++ b/tests/Feature/VercelProtocolStreamTest.php
@@ -4,9 +4,14 @@ use Laravel\Ai\Approvals\PendingApproval;
 use Laravel\Ai\Responses\Data;
 use Laravel\Ai\Responses\Data\Usage;
 use Laravel\Ai\Responses\StreamableAgentResponse;
+use Laravel\Ai\Streaming\Events\ReasoningDelta;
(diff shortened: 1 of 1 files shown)


pushpak1300:
Thanks

PR #846, opened 2026-08-04
Fix OpenAI gateway dropping cache write tokens from usage

aqu1les:
The OpenAI gateway hardcodes `cacheWriteInputTokens` to `0` and never reads `cache_write_tokens` from the response, so cache writes get counted as plain input.

That was harmless while writes were priced like normal input. The `gpt-5.6` family charges a premium to write the prefix into cache, so any cost derived from `Usage` now comes out under the real spend.

Same ~18k prefix sent twice to `gpt-5.6-luna`:

```jsonc
// write
"input_tokens": 18014,
"input_tokens_details": { "cache_write_tokens": 18011, "cached_tokens": 0 }

// read
"input_tokens": 18014,
"input_tokens_details": { "cache_write_tokens": 0, "cached_tokens": 18011 }
```

The field is under `input_tokens_details` (Responses API), not `prompt_tokens_details` where the OpenRouter gateway reads it.

`input_tokens` is the total and both detail fields are slices of it: `18014 = 3 + 18011 + 0` on the write, `3 + 0 + 18011` on the read. The premium is charged **in place of** the input rate, not on top of it, so leaving the write inside `promptTokens` would bill the same slice twice. Subtracting it is the same treatment `cached_tokens` already gets in these two methods, and what the Anthropic gateway does.

## Tests

Two cases pin the write, non-streaming and streaming, and both fail on `0.x` before the change. A third covers the read path so the new subtraction can't break it. `OpenAiHelpers` needed `cache_write_tokens` and a model override threaded through; existing callers keep their defaults.

Also checked live against `gpt-5.6-luna`, a cold prefix reports 26,111 write, 0 read, 3 prompt.

[prompt caching docs](https://developers.openai.com/api/docs/guides/prompt-caching)

### DIFF
--- a/src/Gateway/OpenAi/Concerns/HandlesTextGeneration.php
+++ b/src/Gateway/OpenAi/Concerns/HandlesTextGeneration.php
@@ -267,11 +267,14 @@ trait HandlesTextGeneration
                 $responseId = $response['id'] ?? $responseId;
                 $responseUsage = $response['usage'] ?? [];
 
+                $cachedTokens = $responseUsage['input_tokens_details']['cached_tokens'] ?? 0;
+                $cacheWriteTokens = $responseUsage['input_tokens_details']['cache_write_tokens'] ?? 0;
+
                 $usage = new Usage(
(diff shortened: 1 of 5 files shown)


pushpak1300:
thanks.

PR #847, opened 2026-08-04
Add AgentFailed event when an agent prompt fails

### LINKED ISSUE #782: [Feature Request] Add `AgentFailed` event when an agent prompt fails
## Summary

Currently, the Laravel AI SDK dispatches `PromptingAgent` when an agent starts and `AgentPrompted` when it completes successfully. However, there is no event dispatched when an agent call fails (e.g. due to an unrecoverable exception, a provider error, or a timeout).

This gap makes it impossible for third-party packages or application-level listeners to reliably track failed AI calls — records created on `PromptingAgent` remain in a "processing" state indefinitely with no way to transition them to a "failed" state automatically.

## Proposed Solution

Add a new `AgentFailed` event dispatched whenever an agent prompt results in a terminal failure (i.e. not a failover — those are already covered by `AgentFailedOver`).

```php
namespace Laravel\Ai\Events;

use Laravel\Ai\Prompts\AgentPrompt;
use Throwable;

class AgentFailed
{
    public function __construct(
        public string $invocationId,
        public AgentPrompt $prompt,
        public Throwable $exception,
        public ?object $usage = null, // token usage if partially available
    ) {}
}
```

## Why Include Token Usage

Some providers return partial token usage even on failed requests (e.g. when a request is rejected after being processed, or when a structured output validation fails after the model has already generated a response). Including a nullable `$usage` on the event would allow listeners to log those tokens accurately rather than discarding them.

> **Note:** #307  touches on exposing token/meta data from the SDK, and while there is some overlap in motivation, this request is specifically about the **failure lifecycle** — the need for an event that fires when a prompt fails terminally, regardless of whether token data is available. Maybe can they be grouped together ? 

## Current Workaround

Without this event, the only viable workaround is a scheduled command that marks stale `processing` records as `failed` after a configurable timeout — which is fragile and introduces unnecessary latency in failure reporting.

## Event List Impact

The new event would fit naturally alongside the existing lifecycle events:

| Event | When |
|---|---|
| `PromptingAgent` | Agent prompt starts |
| `AgentPrompted` | Agent prompt completes successfully |
| **`AgentFailed`** | **Agent prompt fails terminally** |
| `AgentFailedOver` | Provider failed but a fallback succeeded |

## Use Cases

- Logging packages that track AI usage and need to mark failed calls
- Alerting / monitoring on AI failure rates
- Retrying failed calls from a queue listener

fissible:
- **Add event for terminal agent failures**
- **Preserve processed prompts on agent failures**

Closes #782

### DIFF
--- a/src/Events/AgentFailed.php
+++ b/src/Events/AgentFailed.php
@@ -1,0 +1,17 @@
+<?php
+
+namespace Laravel\Ai\Events;
+
+use Laravel\Ai\Prompts\AgentPrompt;
+use Laravel\Ai\Responses\Data\Usage;
+use Throwable;
+
+class AgentFailed
+{
+    public function __construct(
+        public string $invocationId,
+        public AgentPrompt $prompt,
+        public Throwable $exception,
+        public ?Usage $usage = null,
+    ) {}
+}
--- a/src/Promptable.php
+++ b/src/Promptable.php
@@ -355,7 +355,7 @@ trait Promptable
      */
     private function recordAgentFailure(string $invocationId, AgentPrompt $prompt, Throwable $exception): void
     {
-        event(new AgentFailed($invocationId, $prompt, $exception));
+        event(new AgentFailed($invocationId, $prompt->processed(), $exception));
     }
 
     /**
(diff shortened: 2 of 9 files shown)


pushpak1300:
closing because of #848

PR #848, opened 2026-08-04
Add step, failure, and sub-agent correlation events for agent runs

### LINKED ISSUE #782: [Feature Request] Add `AgentFailed` event when an agent prompt fails

pushpak1300:
Right now a run gives you two events: one when it starts, one when it finishes. A five step tool calling run still gives you those same two. And if it throws, you get nothing at all, so anything you opened at the start never closes.

This adds events for each step, and for failures.

```php
// fires after every provider call, not just once at the end
Event::listen(StepCompleted::class, function (StepCompleted $event) {
    Log::info("step {$event->stepNumber}", [
        'in' => $event->response->usage->promptTokens,
        'out' => $event->response->usage->completionTokens,
        'finish' => $event->response->finishReason->value,
        'ms' => $event->time,
    ]);
});

// never fired before, the run just died quietly
Event::listen(AgentFailed::class, function (AgentFailed $event) {
    report($event->exception);
});
```

New events: `StartingStep`, `StepCompleted`, `StepFailed`, `AgentFailed`, `ToolFailed`.

`StartingStep` carries what is being sent, so a span opened there can record the request rather than reconstructing it from the agent:

```php
Event::listen(StartingStep::class, function (StartingStep $event) {
    $event->messages;                // including the tool results of the steps before it
    $event->options?->temperature;   // resolved for this step, not the agent's own
});
```

> [!NOTE]
> These events carry the run's live state and are meant for synchronous listeners. `StartingStep` holds the whole message history, so a `ShouldQueue` listener serializes it — attachments and all — once per step. If you need to defer work, extract what you need in a synchronous listener and queue that.

`StepCompleted` carries the step response whole, the same way `AgentPrompted` carries its own, so the text and usage of an individual step are both available and new response fields don't need a constructor change here:

```php
Event::listen(StepCompleted::class, function (StepCompleted $event) {
    $event->response->text;
    $event->response->usage;
    $event->response->meta->model;   // what answered, which may differ from $event->model
    $event->response->toolCalls;
});
```

Everything in one run now shares the same `invocationId`, including across failover, so you can group events by run:

```php
Event::listen(StepCompleted::class, fn ($e) => Cache::increment("run:{$e->invocationId}:steps"));
```

Sub-agents called as tools are linked to their parent as well, so a delegated run can be traced back to the tool call that started it:

```php
Event::listen(PromptingAgent::class, function (PromptingAgent $event) {
    $event->prompt->parentInvocationId;     // the run that delegated
    $event->prompt->parentToolInvocationId; // the tool call that delegated
});
```

Step and tool events carry the wall time spent in the call they report, as `time`, in milliseconds — matching `QueryExecuted::$time`.

### A stream error now throws instead of ending the stream quietly

This is the one behavior change in the PR, and it is deliberate.

When a provider reported an error **inside** the stream body rather than as an HTTP error — a 200 response whose SSE payload contains `{"error": {...}}`, which is how most providers signal a mid-stream rate limit or overload — the loop used to `break`. That path suppressed the terminal `StreamEnd` entirely, so the consumer received the `Error` event and then a stream that simply stopped: no finish reason, a partial `->text`, and a run that had failed reporting no failure. There was no event to hang a closing span on because, as far as the library was concerned, nothing had gone wrong.

The step now throws `Laravel\Ai\Exceptions\StreamErrorException` (extends `AiException`), after dispatching `StepFailed` and `AgentFailed`.

```php
use Laravel\Ai\Exceptions\StreamErrorException;

try {
    foreach ($agent->stream('Hi') as $event) {
        // The provider's Error event still reaches you first, unchanged.
    }
} catch (StreamErrorException $e) {
    $e->error?->type;      // 'rate_limit_exceeded' — the provider's own classification
    $e->error?->message;
    $e->error?->metadata;
}
```

**What you need to do.** If you consume `->stream()` and relied on the loop ending quietly on a stream-body error, wrap the iteration and handle `StreamErrorException`. Everything you had before is still available: the `Error` event is yielded to the consumer *before* the throw, and the same instance is on `$e->error`.

**Two consequences worth knowing.**

- Over SSE via `toResponse()`, the `data:` frame for the `Error` event is flushed, then the exception aborts the response callback before `data: [DONE]`. Clients keying off the error frame are unaffected; clients waiting for `[DONE]` will see a truncated response.
- A stream-body error still does not trigger provider failover, exactly as before. `HandlesFailoverErrors` maps `RequestException` by HTTP status (429/402/503), and these arrive with status 200, so they never reached that mapper. Making them failoverable needs a real recoverability signal — `Error::$recoverable` is hardcoded `false` at all twelve construction sites — and is left to a follow-up.

Suppressing the failure was never the safer option; it just moved the cost from an exception you can catch to a truncated answer you cannot detect.

### Event signatures

- `AgentFailedOver` gains a required `string $invocationId` as its last constructor argument. Existing listeners are unaffected; anything constructing the event by hand must pass it.
- `ToolInvoked` gains a required `float $time` as its last constructor argument. The gateway always measures it, so it is not nullable.
- `StepFailed` (new in this PR) exposes the stream error on `$event->exception->error` when the exception is a `StreamErrorException`, rather than as a separate `$event->error` property.


goatrenterguy:
This is a fantastic step to making it easier to setup analytics for AI agents. Should dramatically simplify my current setup

PR #849, opened 2026-08-05
Add Vercel protocol coverage for citation, error, and default tool error parts

dfinchenko:
This fills the last gaps in `VercelProtocolStreamTest`. Three `toVercelProtocolArray()` mappings still had no coverage anywhere in the suite:                                                                    
                                                                                                                                                                                                                                         
  - `Citation` - `source-url`, emitted by the Anthropic and OpenRouter streaming gateways when a web search result is cited mid-text.                                                                                                    
  - `Error` - `error` / `errorText`, emitted by every streaming gateway (Anthropic, Gemini, Groq, Mistral, Ollama, OpenAI, OpenAI-compatible, OpenRouter, DeepSeek, xAI) as well as the `BroadcastAgent` job.                            
  - The `errorText` fallback in `ToolResult`, used when a tool call fails without an error message. The existing unexecuted-tool-call test only covers the branch where an explicit message is present, so the `'The tool call failed.'` default was never asserted. It is reachable from `TextGenerationLoop`, which passes the tool result as the error text on the final step — that is `null` whenever the tool itself returned `null`.                                     
                                                                                                                                                                                                                                         
This adds three tests: a cited text stream emitting `source-url` between its `text-delta` and `text-end` frames, a failed stream emitting `error` with no `finish` frame (matching how the gateways yield an error and stop), and a failed tool call falling back to the default error text. All three reuse the existing `vercelProtocolParts()` helper and follow the style of the surrounding tests.                                                                    
                                                                                                                                                                                                                                         
Tests only, no source changes.

### DIFF
--- a/tests/Feature/VercelProtocolStreamTest.php
+++ b/tests/Feature/VercelProtocolStreamTest.php
@@ -4,6 +4,8 @@ use Laravel\Ai\Approvals\PendingApproval;
 use Laravel\Ai\Responses\Data;
 use Laravel\Ai\Responses\Data\Usage;
 use Laravel\Ai\Responses\StreamableAgentResponse;
+use Laravel\Ai\Streaming\Events\Citation;
+use Laravel\Ai\Streaming\Events\Error;
 use Laravel\Ai\Streaming\Events\ReasoningDelta;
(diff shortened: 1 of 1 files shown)


pushpak1300:
thanks.

PR #854, opened 2026-08-07
Reset text part state after TextEnd in Responses API streams

### LINKED ISSUE #853: OpenAI gateway: text stream does not start a new text part after TextEnd
### Summary

The OpenAI Responses gateway emits a Vercel-protocol `text-end` event for each `response.output_text.done` event. It does not reset the text-part state after that event.

If the same step receives a later `response.output_text.delta` event, the gateway emits more `text-delta` events under the old text-part id. It does not emit a new `text-start` event first.

AI SDK UI clients then reject the stream.

### Package version

- `laravel/ai` v0.10.2 (also present on `v0.10.3` and current `0.x`)
- Client: Vercel AI SDK UI message stream (`ai` package, `processUIMessageStream`)

### How we stream

We return the agent stream with `usingVercelDataProtocol()`, as the Laravel AI docs describe.

### Error from the client

```
Received text-delta for missing text part with ID "...". Ensure a "text-start" chunk is sent before any "text-delta" chunks.
```

The client removes the text part when it receives `text-end`. A later `text-delta` for the same id has no active part.

### Where it happens

`src/Gateway/OpenAi/Concerns/HandlesTextGeneration.php` in `processTextStream()`.

On `response.output_text.done`, the code emits `TextEnd` and continues. It does not set `$textStartEmitted = false`. It does not create a new `$messageId`.

Pinned copy of the block:

https://github.com/laravel/ai/blob/71b90141894a409de0929c4e1c522bbf7a706bfb/src/Gateway/OpenAi/Concerns/HandlesTextGeneration.php#L100-L107

### Contrast in the same method

For reasoning, the code resets the id after `ReasoningEnd`:

https://github.com/laravel/ai/blob/71b90141894a409de0929c4e1c522bbf7a706bfb/src/Gateway/OpenAi/Concerns/HandlesTextGeneration.php#L142-L149

```php
yield (new ReasoningEnd(
    $this->generateEventId(),
    $reasoningId,
    time(),
))->withInvocationId($invocationId);

$reasoningId = '';
```

Text handling after `TextEnd` has no equivalent reset.

### Why a second text segment can occur

The OpenAI Responses stream can emit more than one text content part in one response step. Each part ends with `response.output_text.done`.

Refs:

- OpenAI streaming events: https://developers.openai.com/api/docs/guides/streaming-responses
- Vercel text parts (`text-start` / `text-delta` / `text-end`): https://ai-sdk.dev/docs/ai-sdk-ui/stream-protocol

This is intermittent. It only appears when one step has more than one distinct text segment.

### Expected behavior

After each `TextEnd` for a text part:

1. Set `$textStartEmitted = false`.
2. Create a new `$messageId` with `generateEventId()`.

Then a later `response.output_text.delta` in the same step emits a new `TextStart` before any `TextDelta`.

### Suggested change

In the `response.output_text.done` branch, after you yield `TextEnd`:

```php
$textStartEmitted = false;
$messageId = $this->generateEventId();
```

### Notes

- Present since the OpenAI Responses gateway landed in #275. Later refactors did not change this block.
- Current OpenAI streaming tests cover one text segment only. A second `output_text.done` then a second delta sequence will catch this.
- Related protocol coverage in #845 covers a single `text-start` / `text-delta` / `text-end` sequence.

chrisnetonline:
Fixes #853

## Summary

When the OpenAI Responses gateway emits multiple text segments in one response step, later `response.output_text.delta` events must be preceded by a new `text-start` chunk for Vercel AI SDK UI clients.

After each `TextEnd`, reset the text part state so a later delta in the same step starts a fresh text part:

- set `$textStartEmitted = false`
- generate a new `$messageId`

This matches the existing reasoning reset pattern in the same method.

## Changes

- Reset text part state after `TextEnd` in `HandlesTextGeneration` (OpenAI and Azure OpenAI)
- Apply the same fix in `HandlesTextStreaming` (xAI uses the same Responses API event shape)
- Add multi-segment streaming tests for OpenAI and xAI

## Test plan

- [x] `./vendor/bin/pest tests/Feature/Providers/OpenAi/StreamingTest.php --filter="text part"`
- [x] `./vendor/bin/pest tests/Feature/Providers/Xai/StreamingTest.php --filter="text part"`
- [x] `./vendor/bin/pest tests/Feature/Providers/OpenAi/StreamingTest.php tests/Feature/Providers/Xai/StreamingTest.php tests/Feature/Providers/AzureOpenAi/StreamingTest.php`

### DIFF
--- a/src/Gateway/Anthropic/Concerns/HandlesTextStreaming.php
+++ b/src/Gateway/Anthropic/Concerns/HandlesTextStreaming.php
@@ -55,7 +55,7 @@ trait HandlesTextStreaming
         $usage = null;
         $stopReason = '';
 
-        $emitTextStart = function () use (&$textStartEmitted, $messageId, $invocationId): ?\Laravel\Ai\Streaming\Events\StreamEvent {
+        $emitTextStart = function () use (&$textStartEmitted, &$messageId, $invocationId): ?\Laravel\Ai\Streaming\Events\StreamEvent {
             if ($textStartEmitted) {
                 return null;
             }
@@ -250,6 +250,7 @@ trait HandlesTextStreaming
                     ))->withInvocationId($invocationId);
 
                     $textStartEmitted = false;
+                    $messageId = $this->generateEventId();
                 } elseif ($currentBlockType === 'thinking' && $reasoningStartEmitted) {
                     if (isset($responseContent[$currentBlockIndex])) {
                         $responseContent[$currentBlockIndex]['thinking'] = $currentThinkingText;
--- a/src/Gateway/OpenAi/Concerns/HandlesTextGeneration.php
+++ b/src/Gateway/OpenAi/Concerns/HandlesTextGeneration.php
@@ -104,6 +104,9 @@ trait HandlesTextGeneration
                     time(),
                 ))->withInvocationId($invocationId);
 
+                $textStartEmitted = false;
+                $messageId = $this->generateEventId();
+
                 continue;
             }
 
--- a/src/Gateway/Xai/Concerns/HandlesTextStreaming.php
+++ b/src/Gateway/Xai/Concerns/HandlesTextStreaming.php
@@ -106,6 +106,9 @@ trait HandlesTextStreaming
                     time(),
                 ))->withInvocationId($invocationId);
 
+                $textStartEmitted = false;
+                $messageId = $this->generateEventId();
+
                 continue;
             }
 
--- a/tests/Feature/Providers/Anthropic/StreamingTest.php
+++ b/tests/Feature/Providers/Anthropic/StreamingTest.php
@@ -41,6 +41,42 @@ describe('text streaming', function (): void {
             ->and($events[4])->toBeInstanceOf(TextEnd::class)
             ->and($events[5])->toBeInstanceOf(StreamEnd::class);
     });
+
+    test('streaming starts a new text part after each text block', function (): void {
+        Http::fake([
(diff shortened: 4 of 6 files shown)


sulimanbenhalim:
nice fix. worth flagging that anthropic has the same duplicate-id class, just from a different path.

on a single response with more than one text block (the `text -> server_tool_use (web search) -> text` flow) the anthropic gateway re-emits `TextStart` for the 2nd block but reuses the same `$messageId`, so both text parts go out with the same id. i faked a text/web_search/text stream and got 2 `TextStart` + 2 `TextEnd` all sharing one id. same thing tracked in vercel/ai#10781 for @ai-sdk/anthropic.

its a bit different from the openai case though. anthropic already resets `textStartEmitted = false` on `content_block_stop`, so it does re-emit a start and doesnt hit the "missing text part" error, the two parts just collide on id. and its own thinking blocks already regenerate their id per block, text is the odd one out.

one gotcha if the fix gets extended there: `$emitTextStart` captures `$messageId` by value, so you cant just regenerate `$messageId` in the loop or TextStart (old id) and TextDelta/TextEnd (new id) desync and you'd actually trigger the missing-part error. needs `&$messageId` by ref + regenerate after the text `content_block_stop`.

happy to send the anthropic side as a follow-up if you'd rather keep this PR focused, or you can fold it in, whatever works.

pushpak1300:
Thanks.

PR #856, opened 2026-08-08
Correlate tool invocation ids across nested generations

### LINKED ISSUE #855: ToolInvoked reports the wrong toolInvocationId when a tool invokes a sub-agent
## Summary

`GeneratesText::$currentToolInvocationId` is a single mutable property on a provider, and providers are memoized per name by `MultipleInstanceManager`. When a tool's `handle()` starts a nested generation (which `AgentTool` does by design) the nested generation overwrites that property and never restores it. The outer tool's `ToolInvoked` then carries the inner tool's id.

The result is an event that emits a pair which cannot exist: an outer generation's `invocationId` alongside a `toolInvocationId` that was only ever issued inside the inner generation.

## Reproduction

An outer agent with a sub-agent as a tool, and the sub-agent calling a leaf tool. Recording `InvokingTool` and `ToolInvoked`:

```
#  event         invocationId  toolInvocationId
0  InvokingTool  9ec210        b0fd81            <- outer tool starting
1  InvokingTool  251033        cc44fc            <- inner tool starting
2  (handle runs)
3  ToolInvoked   251033        cc44fc            <- inner tool done, correct
4  ToolInvoked   9ec210        cc44fc            <- outer tool done, wrong
```

Event 4 should read `9ec210 / b0fd81`.

Note that reaching this requires *not* using `Ai::fakeAgent()`. `AiManager::textProviderFor()` clones the provider per resolution for faked agents, so each agent gets its own instance and the shared property is never actually shared. A test written with `fakeAgent()` passes. Swapping the gateway onto the resolved provider instead reproduces production's memoization:

```php
Ai::textProvider('openai')->useTextGateway(new FakeTextGateway([
    new ToolCall(id: 'call_outer', name: 'InnerAgent', arguments: ['task' => 'go']),
    new ToolCall(id: 'call_inner', name: 'ProbeTool', arguments: []),
    'inner done',
    'outer done',
]));

(new OuterAgent)->prompt('go');
```

## Cause

In `GeneratesText::listenForToolInvocations()` (`src/Providers/Concerns/GeneratesText.php:186-201`), the two ids in the same closure pair are scoped differently:

```php
invoking: function (Tool $tool, array $arguments) use ($invocationId, $agent): void {
    $this->currentToolInvocationId = (string) Str::uuid7();
    $this->events->dispatch(new InvokingTool(
        $invocationId, $this->currentToolInvocationId, $agent, $tool, $arguments
    ));
},
invoked: function (Tool $tool, array $arguments, mixed $result) use ($invocationId, $agent): void {
    $this->events->dispatch(new ToolInvoked(
        $invocationId, $this->currentToolInvocationId, $agent, $tool, $arguments, $result
    ));
},
```

`$invocationId` is captured lexically and stays correct under nesting. `$this->currentToolInvocationId` is read from the provider and does not.

`InvokesTools` already treats re-entrancy as a real hazard one layer down — `$toolInvocationCallbackStack` with `pushToolInvocationCallbacks()` / `popToolInvocationCallbacks()` in a `try/finally` exists for exactly this reason. The id living one layer up on the provider did not get the same treatment.

## Impact

Any listener correlating `InvokingTool` to `ToolInvoked` gets mismatched pairs whenever a sub-agent is involved. Audit or evidence trails attribute a completed tool call to the wrong invocation, and span-based telemetry mis-parents.

Sibling tool calls in a single step are *not* affected  - `TextGenerationLoop` runs them sequentially through `array_map`, so only re-entrancy triggers this.

## Related

- **#618** (open) proposes event-driven telemetry with OpenTelemetry drivers built on these events. Spans derived from `InvokingTool` / `ToolInvoked` pairs would be mis-parented under any sub-agent, so it seems worth resolving before that lands.
- **#347** (completed) centralized the multi-step tool loop into the SDK core, which is where `InvokesTools` and its `pushToolInvocationCallbacks()` / `popToolInvocationCallbacks()` stack come from. That stack is the precedent the suggested fix follows; re-entrancy was already handled for the callbacks, just not for the id one layer up.

I searched open and closed issues and PRs for `toolInvocationId`, `AgentTool`, `ToolInvoked`, and nested/sub-agent correlation and did not find an existing report, but I may have missed one.

## Suggested fix

A stack rather than a scalar, mirroring the existing `InvokesTools` precedent: push in `invoking`, pop in `invoked`. Because `executeTool` brackets both callbacks in one synchronous frame, nesting is strictly LIFO and cannot drift. This stays entirely within `GeneratesText` and changes no signatures.

## Environment

- `laravel/ai` 0.10.x
- Confirmed deterministically with `FakeTextGateway`; no provider calls involved.

fissible:
Fixes #855.

`GeneratesText::$currentToolInvocationId` is a single scalar on the provider, and providers are
memoized per name by `MultipleInstanceManager`. When a tool starts a nested generation — which
`AgentTool` does by design when it prompts its sub-agent — the nested generation overwrites that
property and never restores it. The outer tool's `ToolInvoked` then carries the inner tool's id.

The two ids handed to the same constructor are scoped differently, which is the whole of the bug:

```php
invoking: function (Tool $tool, array $arguments) use ($invocationId, $agent): void {
    $this->currentToolInvocationId = (string) Str::uuid7();
    $this->events->dispatch(new InvokingTool(
        $invocationId, $this->currentToolInvocationId, $agent, $tool, $arguments
    ));
},
```

`$invocationId` is captured lexically and stays correct under nesting. `$this->currentToolInvocationId`
is read off the shared provider and does not. The result is a `ToolInvoked` carrying an outer
generation's `invocationId` beside a `toolInvocationId` that was only ever issued inside the inner
generation — a pair that cannot exist.

## The change

A stack instead of a scalar: push in `invoking`, pop in `invoked`. `InvokesTools` already treats
re-entrancy as real one layer down — `$toolInvocationCallbackStack` with
`pushToolInvocationCallbacks()` / `popToolInvocationCallbacks()` in a `try`/`finally` exists for
exactly this reason. The id living one layer up did not get the same treatment. Because
`executeTool()` brackets both callbacks in one synchronous frame, unwinding is strictly LIFO.

`$currentToolInvocationId` is kept and kept accurate — it now returns to the enclosing tool when a
nested invocation completes, rather than being left pointing at the finished inner one. No signatures
change and the fix stays inside `GeneratesText`.

## Test

`tests/Feature/NestedToolInvocationIdTest.php` drives a parent agent through a sub-agent to a leaf
tool with an `Http::fake()` sequence, following the shape of
`tests/Feature/Providers/TimeoutInToolLoopTest.php`. It asserts that each `ToolInvoked` carries the
id its own `InvokingTool` announced, and that the two invocations do not share an id.

The test fails on `0.x` and passes with the fix:

| | result |
|---|---|
| `0.x` | 1 failed, 1980 passed |
| with fix | 1981 passed |

No other test changes state. `pint` and `phpstan` are clean.

Existing coverage does not catch this. `tests/Integration/AgentTest.php:232` is the only assertion on
`toolInvocationId` in the suite and it checks `! is_null($event->toolInvocationId)` — the clobbered
value is a valid uuid, just the wrong one, so it passes. `tests/Integration/MultiAgentTest.php:56-58`
runs the exact parent → sub-agent shape and asserts on `ToolInvoked`, but never inspects the id.

Note that a test written with `Agent::fake()` cannot reproduce this: `AiManager::textProviderFor()`
clones the provider per resolution for faked agents, so each agent gets its own instance and the
shared property is never actually shared. The `Http::fake()` approach keeps both generations on the
one memoized provider that production uses.

## Impact

Any listener correlating `InvokingTool` to `ToolInvoked` gets mismatched pairs whenever a sub-agent is
involved: audit trails attribute a completed tool call to the wrong invocation, and span-based
telemetry mis-parents. Sibling tool calls in a single step are unaffected — `TextGenerationLoop` runs
them sequentially through `array_map`, so only re-entrancy triggers it.

### DIFF
--- a/src/Providers/Concerns/GeneratesText.php
+++ b/src/Providers/Concerns/GeneratesText.php
@@ -42,6 +42,17 @@ trait GeneratesText
     protected string $currentToolInvocationId;
 
     /**
+     * The invocation IDs of the tools that are currently executing.
+     *
+     * A tool may start a nested generation while it runs, which an AgentTool does by design when
+     * it prompts its sub-agent. Providers are memoized per name, so the nested generation shares
+     * this instance. Invocations always unwind in LIFO order.
+     *
+     * @var array<int, string>
+     */
+    protected array $toolInvocationIds = [];
+
+    /**
      * Invoke the given agent.
      */
     public function prompt(AgentPrompt $prompt): AgentResponse
@@ -190,15 +201,19 @@ trait GeneratesText
     {
         $this->textGenerationLoop()->onToolInvocation(
             invoking: function (Tool $tool, array $arguments) use ($invocationId, $agent): void {
-                $this->currentToolInvocationId = (string) Str::uuid7();
+                $this->toolInvocationIds[] = $this->currentToolInvocationId = (string) Str::uuid7();
 
                 $this->events->dispatch(new InvokingTool(
                     $invocationId, $this->currentToolInvocationId, $agent, $tool, $arguments
                 ));
             },
             invoked: function (Tool $tool, array $arguments, mixed $result) use ($invocationId, $agent): void {
+                $toolInvocationId = array_pop($this->toolInvocationIds) ?? $this->currentToolInvocationId;
+
+                $this->currentToolInvocationId = end($this->toolInvocationIds) ?: $toolInvocationId;
+
                 $this->events->dispatch(new ToolInvoked(
-                    $invocationId, $this->currentToolInvocationId, $agent, $tool, $arguments, $result
+                    $invocationId, $toolInvocationId, $agent, $tool, $arguments, $result
                 ));
             },
         );
--- a/tests/Feature/NestedToolInvocationIdTest.php
+++ b/tests/Feature/NestedToolInvocationIdTest.php
@@ -1,0 +1,119 @@
+<?php
+
+use GuzzleHttp\Promise\PromiseInterface;
+use Illuminate\Support\Facades\Event;
+use Illuminate\Support\Facades\Http;
+use Laravel\Ai\Contracts\Agent;
+use Laravel\Ai\Contracts\CanActAsTool;
+use Laravel\Ai\Contracts\HasTools;
+use Laravel\Ai\Events\InvokingTool;
+use Laravel\Ai\Events\ToolInvoked;
+use Laravel\Ai\Promptable;
+use Laravel\Ai\Tools\AgentTool;
+use Tests\Fixtures\Tools\FixedNumberGenerator;
+
+class NestedResearchAgent implements Agent, CanActAsTool, HasTools
+{
+    use Promptable;
+
+    public function name(): string
+    {
+        return 'research_agent';
+    }
+
+    public function description(): string
+    {
+        return 'Research a topic in depth and return a summary.';
+    }
+
+    public function instructions(): string
+    {
+        return 'You are a research agent. Summarize your findings concisely.';
+    }
+
+    public function tools(): iterable
+    {
+        return [new FixedNumberGenerator];
+    }
+}
+
+class NestedDelegatingAgent implements Agent, HasTools
+{
+    use Promptable;
+
+    public function instructions(): string
+    {
+        return 'You are a project manager that delegates research tasks to your research_agent sub-agent.';
+    }
+
+    public function tools(): iterable
+    {
+        return [new NestedResearchAgent];
+    }
+}
+
+function nestedFakeOpenAiToolCallResponse(string $callId, string $name, string $arguments): PromiseInterface
+{
+    return Http::response([
+        'id' => 'resp_tool_'.$callId,
+        'status' => 'completed',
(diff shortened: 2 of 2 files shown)


fissible:
Superseded by #848. I should have found it before opening this. It removes $currentToolInvocationId and listenForToolInvocations() entirely, scopes the invocation ID locally through ParentInvocation, and fixes the streaming path as well. That's a better solution than the stack proposed here.

One thing from this branch is still worth salvaging: the regression test (tests/Feature/NestedToolInvocationIdTest.php) fails on 0.x and passes against #848 without modification. It drives a parent -> sub-agent -> leaf tool flow through an Http::fake() sequence and asserts that each ToolInvoked event carries the invocation ID announced by its own InvokingTool (I hit this in Verdict, which records these events into an audit trail). You're welcome to take it if it's useful as additional coverage for the correlation case.

I'll leave #855 open for now as a description of the defect in 0.x. If you'd rather track everything under #848, I'm happy to close it.

PR #857, opened 2026-08-08
Add web search support to the xAI provider

sulimanbenhalim:
Adds web search (Grok Live Search) to the xAI provider, matching what OpenAI, Anthropic, Gemini, OpenRouter, and Azure already support.

On the Responses API (which the xAI gateway already uses), Live Search is just a `web_search` tool in the tools array (`{"type":"web_search"}`), the same shape as OpenAI, so it maps the same way. The response side was already in place — xAI's `ParsesTextResponses` already reads `url_citation` annotations into `UrlCitation`s — it just couldn't be triggered because `MapsTools` skipped every provider tool. So this is two small changes:

- `XaiProvider` implements `SupportsWebSearch` and forwards `allowed_domains` (plus xAI-specific options like `excluded_domains` / `enable_image_understanding` via `withProviderOptions`).
- xAI's `MapsTools` maps the `WebSearch` tool to `{"type":"web_search", ...}` instead of dropping it, mirroring the OpenAI gateway.

Tests cover the request mapping (type, allowed_domains, provider options) and citations parsing end-to-end. Docs: https://docs.x.ai/developers/tools/web-search

### DIFF
--- a/src/Gateway/Xai/Concerns/MapsTools.php
+++ b/src/Gateway/Xai/Concerns/MapsTools.php
@@ -23,9 +23,7 @@ trait MapsTools
 
         foreach ($tools as $tool) {
             if ($tool instanceof ProviderTool) {
-                $providerTool = $this->mapProviderTool($tool, $provider);
-
-                if (filled($providerTool)) {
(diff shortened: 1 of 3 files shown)


pushpak1300:
Thanks

PR #858, opened 2026-08-08
Add transcription support to the OpenAI-compatible provider

sulimanbenhalim:
The `openai-compatible` driver implements text and embeddings but not transcription, so calling `transcribe()` on it fails — even though OpenAI-compatible servers like vLLM, LiteLLM, and LocalAI expose the standard `/v1/audio/transcriptions` endpoint.

This wires the driver into the existing transcription architecture, following #841 (which did the same for embeddings):

- `OpenAiCompatibleProvider` implements `TranscriptionProvider` through the shared `GeneratesTranscriptions` and `HasTranscriptionGateway` concerns, so faking and events work unchanged. The default transcription model must be configured (like text and embeddings), since arbitrary endpoints have no known models.
- `OpenAiCompatibleGateway` gains `generateTranscription()`, which uploads the audio as standard OpenAI **multipart** (`file` + `model` + optional `language`) and reuses the gateway's client and failover handling. It mirrors the **OpenAI** gateway's transcription, not OpenRouter's — OpenRouter sends a non-standard base64 JSON body, whereas compatible servers expect the standard multipart upload. Diarization isn't supported (compatible servers run plain Whisper), so it throws a clear `LogicException`, matching OpenRouter.

```php
'my-llm' => [
    'url' => env('MY_LLM_URL'),
    'models' => ['transcription' => ['default' => 'whisper-1']],
],
```

Tests mirror the OpenAI transcription suite: multipart endpoint, model + language in the body, missing-model exception, diarize rejection, response parsing, auth header, rate-limit/overload/error handling, and faking.

### DIFF
--- a/resources/boost/skills/ai-sdk-development/SKILL.md
+++ b/resources/boost/skills/ai-sdk-development/SKILL.md
@@ -420,6 +420,7 @@ Point the SDK at any OpenAI-compatible endpoint (LM Studio, vLLM, Together, etc.
             'default' => 'some-embedding-model',
             'dimensions' => 1024, // optional; omit to use native dimensions
         ],
+        'transcription' => ['default' => 'some-transcription-model'],
     ],
 ],
 ```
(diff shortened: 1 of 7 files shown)


pushpak1300:
Thanks.

PR #866, opened 2026-08-11
Handle Mistral responses that return content as a list of blocks

### LINKED ISSUE #865: [Bug] Mistral gateway crashes when content is a list of blocks
When a Mistral answer draws on a tool result, content arrives as a list of blocks instead of a string:

```json
{"content":[{"type":"reference","reference_ids":["search_documents"]}]}
```

Both Mistral paths assume a string and die on it:

- `Gateway/Mistral/Concerns/HandlesTextStreaming.php:87` -> `$currentText .= $delta['content'];` → ErrorException: Array to string conversion, stream aborts, user gets no answer.
- `Gateway/Mistral/Concerns/ParsesTextResponses.php:46` -> `$message['content']` goes into StepResponse, whose `$text` is string → TypeError: Argument #1 ($text) must be of type string, array given.

Not an edge case: with `mistral-medium-latest`, the array arrived on 10 of 12 runs of the same tool-backed question. The Anthropic gateway already walks block content (`$data['content'] ?? []` into `buildStepResponse()`); Mistral has no equivalent, and there's no non-string guard under `src/Gateway/`.

Important note I discovered this using Claude, but apart from that I saw this back when debugging it into the logs. I wrote a local fix, but I feel this could effect more people that use their models.
If it helps I could write a PR for this, but I don't fully get what would be the best solution to fix this situation.

laravel/ai 0.x, PHP 8.5, mistral-medium-latest.

lazerg:
When a Mistral answer draws on a tool result, `content` comes back as a list of chunks rather than a string, for example `[{"type": "reference", "reference_ids": ["search_documents"]}]`. Both Mistral paths assume a string and break on it: the streaming loop concatenates the delta straight onto `$currentText` and raises `Array to string conversion`, and `parseTextResponse()` passes the array into `StepResponse::$text`, which raises a `TypeError`. Either way the caller ends up with no answer.

This flattens the content value before it is used, keeping the `text` of each text chunk and dropping the rest, the same thing the Anthropic gateway already does when it walks block content. Non-text chunks such as references produce no text delta.

Fixes #865

### DIFF
--- a/src/Gateway/Mistral/Concerns/HandlesTextStreaming.php
+++ b/src/Gateway/Mistral/Concerns/HandlesTextStreaming.php
@@ -73,7 +73,9 @@ trait HandlesTextStreaming
                 ))->withInvocationId($invocationId);
             }
 
-            if (isset($delta['content']) && $delta['content'] !== '') {
+            $content = $this->extractContentText($delta['content'] ?? '');
+
+            if ($content !== '') {
                 if (! $textStartEmitted) {
                     $textStartEmitted = true;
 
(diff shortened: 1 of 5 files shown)


pushpak1300:
Thanks

Issue #850, opened 2026-08-06
Anthropic explicit cache breakpoint


AngeloKarugo:
Hello,

I'm aware that we can use providerOptions() to enable [automatic caching](https://platform.claude.com/docs/en/build-with-claude/prompt-caching#automatic-caching) for Anthropic API. However, the top-level cache gets missed the moment a new conversation is initialised with a different initial message, as per [Anthropic's caching rules](https://platform.claude.com/docs/en/build-with-claude/prompt-caching), which result in a cache miss (and an expensive write) if anything in the cached block changes. To illustrate, the top-level caching behavior:

>(User1 starts a new conversation): [Tools, System, Usr1Msg1] (cache write), [AsstMsg1]

>(User 1 sends second message): [Tools, System, Usr1Msg1] (cache read), [AsstMsg1, Usr1Msg2] (appended to the cache block), [AsstMsg2]

Subsequent messages in User1's conversation get appended to the cache block similarly.

>(User2 starts a new conversation): [Tools, System, Usr2Msg1] (cache write, because User1 cache block contains the additional blocks [Usr1Msg1, AsstMsg1, Usr1Msg2, ...]), [AsstMsg3]

Subsequent messages in User2's conversation get appended to User2's cache block similarly as well.

As you can see, the static Tools and System blocks get written to the cache afresh for virtually every new conversation.

By setting a cache breakpoint after the system block, every user's request will read the static tools + instructions block from cache, and only append the new user's message(s). To illustrate:

>(User1 starts a new conversation): [Tools, System] (cache write via explicit breakpoint at system msg), [Usr1Msg1] (cache write via top-level cache), [AsstMsg1]

>(User1 sends second message): [Tools,System,Usr1Msg1] (cache read, as expected), [AsstMsg1, UsrMsg2] (cache write via top-level cache), [AsstMsg2]

When User2 starts a new conversation:

>(User2 sends first message): [Tools, System] (cache read from User1's cache write at the explicit breakpoint), [Usr2Msg1] (cache write via top-level cache), [AsstMsg3]

Subsequent User2 messages get appended to the User2 cache block.

Instead of every conversation building from 0 cached tokens, each finds the Tools and Instructions blocks already cached, and starts cache-writing from the first user message. This is particularly useful for applications with large system prompts, many tools and/or large volumes of new conversations.

This can be done in the provider config or per Agent using providerOptions() as @pushpak1300 suggested in https://github.com/laravel/ai/pull/365#issuecomment-4212262810.

[...]

PR #869, opened 2026-08-12
Add explicit prompt cache breakpoints for Anthropic and Bedrock

pushpak1300:
Passing `'cache_control' => ['type' => 'ephemeral']` through `providerOptions()` already turns on Anthropic's *automatic* caching, but automatic caching puts the breakpoint after the *last message*. So every fresh conversation re-writes the whole static prefix, the tools and the system instructions, and pays full price for it (#850). The top-level `array_merge($body, $providerOptions)` can't express per-block `cache_control`, and Bedrock has no top-level switch at all.

This PR adds a `prompt_cache` option that marks the static prefix. On Anthropic it emits the system prompt in block form with `cache_control` and stamps the last tool (a breakpoint caches everything before it, so the last tool covers all of them). On Bedrock it appends a `cachePoint` block to `system` and to `toolConfig.tools`. Raw strings work alongside the enum, and a typo throws via `PromptCacheTarget::from()` rather than silently doing nothing. Composes with the existing top-level automatic caching, which still passes through untouched, and the option is never forwarded to providers that don't support it.

Supersedes #663 and #672. Refs #850, #119.

## Documentation

> ### Prompt Caching
>
> Most providers cache repeated prompt prefixes automatically and bill the cached portion at a large discount. OpenAI, Gemini, Groq, DeepSeek, and xAI require no configuration at all, and you may inspect the savings via `$response->usage->cacheReadInputTokens`.
>
> Anthropic and Bedrock do not cache unless you ask them to. Anthropic offers two mechanisms and they may be combined.
>
> **Automatic caching.** A top-level `cache_control` option places a single breakpoint after the last block of the request. As a conversation grows the breakpoint moves forward with it, so each turn reads the previous turns from the cache:
>
> ```php
> public function providerOptions(Lab|string $provider): array
> {
>     return match ($provider) {
>         Lab::Anthropic => ['cache_control' => ['type' => 'ephemeral']],
>         default => [],
>     };
> }
> ```
>
> **Explicit breakpoints.** Automatic caching follows the conversation, so a *new* conversation begins with a cache miss: your tool definitions and instructions are written to the cache again for every fresh conversation. If your agent has a large set of tools or lengthy instructions, place an explicit breakpoint at the end of that static prefix with the `prompt_cache` option. Every conversation then reads the prefix from the cache and only pays to cache its own messages:
>
> ```php
> use Laravel\Ai\Enums\PromptCacheTarget;
>
> public function providerOptions(Lab|string $provider): array
> {
>     return ['prompt_cache' => [PromptCacheTarget::System]];
> }
> ```
>
> The option is supported by the `anthropic` and `bedrock` providers, and it is never forwarded to providers that do not support it, so an agent may safely declare it alongside options for other providers. Two targets are available:
>
> - `PromptCacheTarget::System` caches the tool definitions **and** the instructions. A breakpoint caches everything before it, and providers build the prompt in the order tools, instructions, messages.
> - `PromptCacheTarget::Tools` caches only the tool definitions. Reach for this when your instructions change on every request, for example when they embed the current date or the authenticated user's name. Caching a prefix that changes every request only pays the cache write cost and never reads from it.
>
> Pass both to cache the tools and the instructions independently, which is useful when the two change at different rates.
>
> **Cache duration.** Cached prefixes are retained for five minutes by default. Anthropic can also retain them for an hour, which you may request per target by giving the target a `5m` or `1h` TTL instead of listing it:
>
> ```php
> return [
>     'prompt_cache' => [
>         'tools' => '1h',
>         'system' => '5m',
>     ],
> ];
> ```
>
> The one hour cache costs more to write and the same to read, so it pays off when your agent is called regularly but with gaps longer than five minutes between calls. Bedrock cache points have no TTL control, so a TTL is ignored there.
>
> Targets are given as array keys here, so a TTL is paired with the target's string value rather than the enum case itself (`PromptCacheTarget::Tools->value => '1h'`). The list and map forms may be mixed freely.
>
> Anthropic only caches prompts above a model specific minimum length, ranging from 512 to 4,096 tokens. Shorter prompts are processed normally and no error is returned. Anthropic also limits a request to four cache breakpoints in total.


taylorotwell:
I pushed an alternative API that makes prompt caching a first-class agent configuration rather than a reserved prompt_cache provider option.

```php
use Laravel\Ai\Attributes\CacheInstructions;
use Laravel\Ai\Attributes\CacheToolDefinitions;

#[CacheInstructions(ttl: '1h')]
#[CacheToolDefinitions(ttl: '1h')]
class Assistant implements Agent
{
    // ...
}
```

The changes:

- Replace PromptCacheTarget and the prompt_cache key with two expressive attributes.
- Keep providerOptions() limited to provider-native options.
- Resolve cache configuration once through TextGenerationOptions.
- Preserve caching configuration across multi-step tool calls.
- Translate the attributes into Anthropic cache_control blocks and Bedrock cachePoint blocks.
- Pass supported TTLs through to Bedrock rather than ignoring them.
- Reject invalid TTL ordering before sending the request.
- Avoid calling an agent’s providerOptions() twice per generation step.

PR #878, opened 2026-08-12
Reconstruct S3 document attachments in File::fromArray

sulimanbenhalim:
`S3Document::toArray()` writes `type => 's3-document'` but `File::fromArray()` has no arm for it, so it hits `default => null`. every other attachment type has a case, s3-document is the only one that doesnt.

so if you attach an S3 document and the conversation gets persisted then reloaded (`RemembersConversations` / `DatabaseConversationStore`), the attachment rehydrates to null and gets filtered out. if it was the only attachment on that message, the message drops down to a plain no-attachment message so the doc is gone on every replayed turn. this bites Bedrock since thats the provider that takes S3 document refs.

fix is the one missing match arm. also added `s3-document` to the `FileFromArrayTest` round-trip and malformed datasets, it was the only type not covered there either.


pushpak1300:
Thanks

PR #879, opened 2026-08-13
Strip markdown code fences from OpenAI-compatible structured output

sulimanbenhalim:
the openai-compatible gateway decodes structured output with a raw `json_decode`, its the only text gateway that doesnt go through the `DecodesStructuredOutput` trait (10 others do). so if the backend wraps its json in a markdown code fence the decode returns null and the structured output silently comes back as `[]`.

local backends (vLLM / LM Studio / llama.cpp etc) do this a lot, a bunch of them dont actually honor `response_format: json_schema` and just emit fenced json in the content.

fix routes it through the same `decodeStructuredOutput` helper every sibling uses (strips the fence then decodes). added a structured-response parse test for the compat provider since it only had a request-format one before, plus a fenced variant that was red before the change.

### DIFF
--- a/src/Gateway/Concerns/DecodesStructuredOutput.php
+++ b/src/Gateway/Concerns/DecodesStructuredOutput.php
@@ -9,28 +9,30 @@ trait DecodesStructuredOutput
      */
     protected function decodeStructuredOutput(?string $text): array
     {
-        if ($text === null || trim($text) === '') {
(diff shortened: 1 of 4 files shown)


pushpak1300:
Thanks.

pushpak1300:
Thanks.

PR #882, opened 2026-08-14
Preserve multi-round tool call boundaries in stored conversations

### LINKED ISSUE #402: Multi-round tool call loops lose round boundaries when stored in DatabaseConversationStore
When an agent executes multiple rounds of tool calls in a single conversation turn, all tool calls and results are stored as flat lists in the database. When the conversation history is reconstructed on the next user message, all rounds are collapsed into a single AssistantMessage + ToolResultMessage pair, losing the original round-by-round sequence.

Providers like Anthropic require the interleaved structure to be preserved:

```
AssistantMessage [tool_call round 1]
ToolResultMessage [result round 1]
AssistantMessage [tool_call round 2]
ToolResultMessage [result round 2]
```
Sending the flattened version causes the provider to reject the history on the next user message.

Expected behavior:
The round-by-round sequence is preserved when storing and reconstructing conversation history.

Affected:
All providers that require interleaved tool call/result sequences (Anthropic, Groq).

para1992:
Fixes #402.

`DatabaseConversationStore` previously flattened tool calls and results from multiple rounds into a single assistant/tool-result pair. This produced invalid conversation history for providers that require interleaved tool messages.

This change stores the tool-call IDs from each completed step in the existing message metadata and uses that grouping when reconstructing the conversation.

Legacy or malformed records fall back to the existing behavior. Approval-paused turns remain unchanged.

No database migration or public API change is required.

### DIFF
--- a/src/Storage/DatabaseConversationStore.php
+++ b/src/Storage/DatabaseConversationStore.php
@@ -213,13 +213,17 @@ class DatabaseConversationStore implements ConversationStore
         if (! $response->hasPendingApprovals()) {
             $toolCallRounds = $response->steps
                 ->filter(fn ($step): bool => $step instanceof Step && filled($step->toolCalls))
-                ->map(fn (Step $step): array => array_values(array_map(
-                    fn (ToolCall $toolCall): string => $toolCall->id,
-                    $step->toolCalls,
-                )))
+                ->map(fn (Step $step): array => [
+                    'text' => $step->text,
+                    'tool_call_ids' => array_values(array_map(
+                        fn (ToolCall $toolCall): string => $toolCall->id,
+                        $step->toolCalls,
+                    )),
+                ])
                 ->values();
 
-            if ($toolCallRounds->count() > 1) {
+            if ($toolCallRounds->count() > 1
+                || $toolCallRounds->contains(fn (array $round): bool => filled($round['text']))) {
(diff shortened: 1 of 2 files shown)


barryvdh:
And what if the message contains text first?

>  User: Can you check the status 
>  Assistant: Sure, I'll check
>  Tool 1: Lookup order
>  Tool 2: Lookup carrier
>  Assistant: Here you go, status is ..


Right now it would still be flatted to 1 text message with

> Tool 1 
> Tool 2
> Assistant: Sure, I'll check
>      Here you go, status is ..

It would be preferable to store the original sequence, also for the UI showing this.

para1992:
Thanks, good catch. I’ve updated the implementation to preserve the assistant text attached to each tool-call round here https://github.com/laravel/ai/pull/882/changes/6a8e00ec81683431d21054d24d9eadbeda98ff0a

sulimanbenhalim:
this is the #402 meta-grouping approach and it looks right. pulled the branch, all 31 storage tests green, and mutation-checked the two new ones (forced `reconstructToolCallRounds` to bail, both go red) so theyre real guards. nice that it also recovers the pre-tool-call text, not just the round boundaries. lgtm.

pushpak1300:
Tracking different approach in #1007

PR #884, opened 2026-08-14
Fail over on Anthropic transient gateway and Cloudflare errors

sulimanbenhalim:
#810 widened the base `overloadedStatusCodes` to `[502,503,504,520,522,524]` so transient gateway + cloudflare errors fail over. but `AnthropicGateway` (and `AnthropicFileGateway`) override that method and return `[529]`, which replaces the base list instead of extending it, so anthropic never picked up the new codes.

so every other provider fails over on a 520/522/524 (cloudflare) or 502/504 (gateway), but anthropic doesnt, even though api.anthropic.com is cloudflare fronted. a failover setup with anthropic as primary just hard fails on those.

fix returns `[529, 502, 503, 504, 520, 522, 524]` in both anthropic gateways (529 is anthropics own overloaded code, the rest are the shared transient set). added the same parametrized test the gemini side already has, it was red for anthropic before this.

### DIFF
--- a/src/Gateway/Anthropic/AnthropicFileGateway.php
+++ b/src/Gateway/Anthropic/AnthropicFileGateway.php
@@ -91,6 +91,7 @@ class AnthropicFileGateway implements FileGateway
      */
     protected function overloadedStatusCodes(): array
     {
-        return [529];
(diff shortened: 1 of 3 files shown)


pushpak1300:
Thanks

PR #885, opened 2026-08-14
Allow the caller to supply an agent invocation id

timmcleod:
`Promptable::prompt()` generates the invocation id internally and only hands it back on the response. When a run throws there is no response, so the caller never learns the id of the run it started, even though each event it dispatched, including `AgentFailed`, carries it.

`AgentPrompt` already accepts an `invocationId`, so this is only the entry point. This adds `withInvocationId()` to `Promptable`, following the fluent style of `continue()` and `forParticipant()`. The id is consumed by one run, so a reused agent instance can't stamp a later run with an earlier id. `Contracts\Agent` is unchanged.

Use case: a queued job that prompts an agent and keeps its own record of the work can pass that record's id in, so the run's events arrive already labelled with it instead of being reconciled after the fact.

### DIFF
--- a/src/Promptable.php
+++ b/src/Promptable.php
@@ -40,6 +40,38 @@ trait Promptable
     use SerializesModels;
 
     /**
+     * The invocation id supplied by the caller for the agent's next run, if any.
+     */
+    protected ?string $pendingInvocationId = null;
+
+    /**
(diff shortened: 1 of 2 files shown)


pushpak1300:
Thanks for your pull request!

PR #890, opened 2026-08-17
Add MiniMax text-to-speech support

octo-patch:
Reason: The audio generation pipeline had no MiniMax text-to-speech support, so MiniMax speech models could not be used through `Audio::of(...)`.

## Summary

Adds text-to-speech support to the existing audio generation pipeline via a new `minimax` provider. This is purely additive — no existing provider, contract, or public API changes behaviour.

The provider implements `AudioProvider` only (text-to-speech), so it plugs into `Audio::of(...)->generate(...)`, failover, faking, and queueing without any pipeline changes.

## Usage

```php
use Laravel\Ai\Audio;

// Default model and voice.
$response = Audio::of('Hello there!')->generate(provider: 'minimax');

// Explicit model, or a specific voice id.
$response = Audio::of('Hello there!')
    ->voice('English_expressive_narrator')
    ->generate(provider: 'minimax', model: 'speech-2.8-turbo');

$response->store('audio');
```

## Changes

- `src/Providers/MiniMaxProvider.php` — implements `AudioProvider`, memoizes its audio gateway, and defaults to the `speech-2.8-hd` model (overridable via `models.audio.default`).
- `src/Gateway/MiniMaxGateway.php` — implements `AudioGateway`:
  - `POST t2a_v2` with `model`, `text`, `stream`, `output_format`, `voice_setting.voice_id` and `audio_setting.format`.
  - `default-female` / `default-male` are mapped to voice ids; any other value is passed through as a voice id unchanged.
  - Response parsing: the returned audio is hex encoded, so it is decoded and re-encoded as base64 for `AudioResponse`, with an `audio/mpeg` MIME type.
  - Error handling: a non-zero `base_resp.status_code` is surfaced as an exception even when the HTTP status is 200, and malformed or missing audio payloads are rejected rather than returned as empty audio. HTTP-level failures continue to flow through the shared failover error handling.
  - Authorization uses a bearer token, and the base URL is configurable so both the global and China endpoints are supported.
- `config/ai.php` — adds the `minimax` connection with `MINIMAX_API_KEY` and a `MINIMAX_URL` override defaulting to the global endpoint. Set `MINIMAX_URL=https://api.minimaxi.com/v1` to use the China endpoint.
- `src/Enums/Lab.php`, `src/AiManager.php` — registers the `minimax` driver.
- `resources/boost/skills/ai-sdk-development/SKILL.md` — adds the provider to the TTS row of the provider support table.

## Tests

Added `tests/Feature/Providers/MiniMax/`, mirroring the existing dedicated audio provider tests:

- `AudioTest.php` — request mapping, both voice aliases, custom voice pass-through, bearer header, pass-through of each supported speech model, the default model, hex to base64 decoding of the response, and the failure paths (failing `base_resp.status_code`, missing audio, non-hex audio, error/rate limit/overloaded HTTP responses).
- `BaseUrlTest.php` — the global endpoint fallback, the China endpoint, and a custom base URL.
- `GatewayCachingTest.php` — audio gateway memoization.

## Checks

- `php -l` on every changed PHP file — no syntax errors.
- Exercised `MiniMaxGateway::generateAudio()` directly against stubbed HTTP responses (24 assertions) covering request mapping, voice resolution, every supported speech model, hex decoding, base URL resolution for both endpoints, and each error branch.

Note: `composer test`, `pint` and `phpstan` were not run locally because a PHP 8.3 toolchain was not available in my environment, so the new Pest tests have not been executed here — please let CI (or a maintainer) confirm them.

### DIFF
--- a/config/ai.php
+++ b/config/ai.php
@@ -114,6 +114,12 @@ return [
             'key' => env('JINA_API_KEY'),
         ],
 
+        'minimax' => [
+            'driver' => 'minimax',
+            'key' => env('MINIMAX_API_KEY'),
+            'url' => env('MINIMAX_URL', 'https://api.minimax.io/v1'),
+        ],
+
         'mistral' => [
             'driver' => 'mistral',
             'key' => env('MISTRAL_API_KEY'),
--- a/resources/boost/skills/ai-sdk-development/SKILL.md
+++ b/resources/boost/skills/ai-sdk-development/SKILL.md
@@ -487,7 +487,7 @@ Calling a capability not supported by a provider throws a `LogicException`. Refe
 | ---------- | --------------------------------------------------------------- |
 | Text       | OpenAI, Anthropic, Gemini, Azure, Groq, xAI, DeepSeek, Mistral, Ollama, OpenRouter, OpenAI-compatible |
 | Images     | OpenAI, Gemini, xAI                                            |
-| TTS        | OpenAI, ElevenLabs                                              |
+| TTS        | OpenAI, ElevenLabs, MiniMax                                     |
 | STT        | OpenAI, ElevenLabs, Mistral, Groq, OpenAI-compatible            |
 | Embeddings | OpenAI, OpenAI-compatible, Gemini, Azure, Cohere, Mistral, Jina, VoyageAI |
 | Reranking  | Cohere, Jina                                                    |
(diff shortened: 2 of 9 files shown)


pushpak1300:
Thanks for your pull request!

PR #891, opened 2026-08-17
Added assertPromptedTimes test method

F1nnG:
I have added the `assertPromptedTimes` test method to the `Promptable` trait, similar to how `Bus::assertDispatchedTimes` and `Notification::assertSentTimes` work.

## Example
```php
SalesCoach::assertPromptedTimes(3);
```


pushpak1300:
Thanks.

PR #892, opened 2026-08-17
Surface Anthropic web fetch citations on the response

sulimanbenhalim:
enabling citations on Anthropic's web_fetch tool (`->withProviderOptions(['citations' => ['enabled' => true]])`) returns an empty `$response->meta->citations`. `extractCitations` only handles `web_search_tool_result` blocks, so `web_fetch_tool_result` gets dropped.

this adds a `web_fetch_tool_result` branch mirroring the web_search one, pushing the fetched `content.url` + title as a `UrlCitation` (guarded on `web_fetch_result`, so error blocks are skipped). non-streaming only: web_fetch's streaming citations come as `char_location`/`page_location` deltas with no url to map onto `UrlCitation`.

test included, red before the branch and green after.


pushpak1300:
Thanks

PR #904, opened 2026-08-19
Add an ai_sdk_extra_headers provider option for custom HTTP headers

pushpak1300:
## Why

Today you can set custom HTTP headers for a provider in `config/ai.php`, but those apply to every request that provider makes. There is no way to set a header for a single request.

That matters for anything that routes traffic through a proxy or gateway, because those services identify and tag individual requests with headers. [Cloudflare AI Gateway](https://developers.cloudflare.com/ai-gateway/) is the clearest example: it authenticates with `cf-aig-authorization` and attaches per-request cost and usage metadata with `cf-aig-metadata`, so the value has to change from one call to the next. The same is true of per-tenant billing attribution (`OpenAI-Project`) and of opting a single agent into a provider beta such as `anthropic-beta` without waiting for an SDK release.

## Usage

Pass an `ai_sdk_extra_headers` key to the provider options you already use. It applies everywhere provider options are accepted.

Agents:

```php
class SalesCoach implements Agent, HasProviderOptions
{
    use Promptable;

    {
        return 'You are a sales coach.';
    }

    {
        return [
            'ai_sdk_extra_headers' => ['cf-aig-metadata' => json_encode(['tenant' => $this->team->id])],
        ];
    }
}
```

Embeddings, images, audio, transcription and reranking:

```php
Embeddings::for($chunks)
    ->withProviderOptions(['ai_sdk_extra_headers' => ['cf-aig-metadata' => $metadata]])
    ->generate();

Image::of('A red apple')
    ->withProviderOptions(['ai_sdk_extra_headers' => ['cf-aig-metadata' => $metadata]])
    ->generate();

Audio::of('Hello world')
    ->withProviderOptions(['ai_sdk_extra_headers' => ['cf-aig-metadata' => $metadata]])
    ->generate();

Transcription::fromPath('/path/to/audio.mp3')
    ->withProviderOptions(['ai_sdk_extra_headers' => ['cf-aig-metadata' => $metadata]])
    ->generate();

Reranking::of($documents)
    ->withProviderOptions(['ai_sdk_extra_headers' => ['cf-aig-metadata' => $metadata]])
    ->rerank('What is Laravel?');
```

Documents and files:

```php
Document::fromPath('/path/to/document.pdf')
    ->withProviderOptions(['ai_sdk_extra_headers' => ['cf-aig-metadata' => $metadata]])
    ->put();
```

Closures work as they do for any other provider option, so headers can be resolved per provider during failover:

```php
Embeddings::for($chunks)
    ->withProviderOptions(fn (Provider $provider) => [
        'ai_sdk_extra_headers' => ['cf-aig-metadata' => json_encode(['driver' => $provider->driver()])],
    ])
    ->generate();
```

Configured headers keep working, and a request header replaces one of the same name, matched without regard to casing.

## Why it's done this way

Headers have to work for agents, five generation builders and file uploads. A `withHeaders()` method means one new method per capability, and another one every time a capability is added. Provider options already reach all of those places and already resolve per provider, so headers travel as a reserved key in that bag instead. Nothing new appears on the public API.

The trade-off is that `ai_sdk_extra_headers` looks like a request body field. It is defined once as `HasProviderOptions::HEADERS` and removed from the options before they are sent, so it never reaches the body. The `ai_sdk_` prefix keeps it out of the namespace any provider uses for body parameters, and gives future SDK level options a home: anything the SDK reserves, including experimental features we want to ship without committing to a public method, goes under the same prefix.

Headers are excluded from the embeddings cache key, since they don't change the vectors.

Depends on #899, #900 and #901 for the image, audio and reranking builders.


taylorotwell:
Added a public withHeaders() convenience method to all generation builders and file uploads, allowing request-specific HTTP headers without manually placing values under the internal ai_sdk_extra_headers provider option:

```php
Embeddings::for($chunks)
    ->withHeaders([
        'cf-aig-metadata' => json_encode(['tenant' => $tenantId]),
    ])
    ->withProviderOptions(['dimensions' => 1024])
    ->generate();
```

Header arrays and closures follow the same provider-resolution behavior as withProviderOptions(), compose safely with normal provider options, survive serialization for queued requests, and are excluded from provider request bodies and embedding cache keys. File upload option resolution was also consolidated so dynamic header closures run only once per upload.

PR #912, opened 2026-08-20
Emit citation events for Gemini grounding results while streaming

sulimanbenhalim:
gemini's google search grounding citations (groundingMetadata) are surfaced on the non-streaming path via extractCitations, but the streaming handler never emitted them, so streaming loses all the web sources.

this runs the same extractCitations on the final chunk and yields a CitationEvent per grounding citation, matching anthropic (#892) and openrouter. red/green streaming test added.

### DIFF
--- a/src/Gateway/Gemini/Concerns/HandlesTextStreaming.php
+++ b/src/Gateway/Gemini/Concerns/HandlesTextStreaming.php
@@ -43,6 +43,7 @@ trait HandlesTextStreaming
         $modelParts = [];
         $usage = null;
         $data = [];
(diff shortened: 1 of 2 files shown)


pushpak1300:
Gemini doesn’t guarantee grounding metadata lands on the last chunk. It can send a trailing usage-only chunk with no candidates which would drop all citations again.  To fix this we remember the last chunk that actually contained grounding or citation metadata and extract from that instead. I’ve also added a test with a candidate-less trailing chunk to ensure it works as expected.

PR #914, opened 2026-08-20
Read Cohere Bedrock embeddings input token count from response header

serpentblade:
Cohere's Bedrock `invoke_model` response body carries no usage data for embeddings, so the gateway currently always reports 0 input tokens for Cohere embeddings on Bedrock. AWS does report the count, in the `x-amzn-bedrock-input-token-count` response header, so this reads it from there instead of hardcoding 0.

Claude assisted with this change; I reviewed, tested, and understand it.


serpentblade:
What more information or help can I provide?  Probably isn't sufficient that I'm already using this in production via composer patches so I can properly record my AI usage for embeddings

pushpak1300:
@serpentblade is there any documentation where we can check this ?

PR #918, opened 2026-08-21
Convert Vercel AI SDK UI messages into agent chat input

pushpak1300:
Builds on the previous PR in the stack.

[`useChat`](https://ai-sdk.dev/docs/ai-sdk-ui/chatbot) POSTs [UI messages](https://ai-sdk.dev/docs/reference/ai-sdk-core/ui-message): arrays of typed parts where the text lives in `parts`, files arrive as `data:` URLs or remote URLs inside [`file` parts](https://ai-sdk.dev/docs/reference/ai-sdk-core/ui-message#filepart), and the full history rides along on every request. None of that maps to what `stream()` accepts, so every app starts with the same parsing chore:

```php
Route::post('/chat', function (Request $request) {
    $messages = $request->input('messages'); // Vercel UI messages

    // 😩 dig the prompt out of the parts array
    $text = collect(end($messages)['parts'])
        ->where('type', 'text')->pluck('text')->implode("\n");

    // 😩 decode data URLs, guess mime types, build attachments by hand
    // 😩 prior turns are silently dropped, the agent forgets the conversation

    return (new ChatAgent)->stream($text)->usingVercelDataProtocol();
});
```

After this PR, one parse of the request gives you the whole turn:

```php
$chat = Vercel::chat($request); // newest message, attachments, approvals, history
```

### Stateless

The client owns the history and sends the full `messages` array each turn:

```php
use Laravel\Ai\Vercel\Vercel;

Route::post('/chat', function (Request $request) {
    $chat = Vercel::chat($request);

    return (new ChatAgent)
        ->withMessages($chat->history())   // every turn before this one
        ->stream($chat)                    // newest user message, file parts become attachments
        ->usingVercelDataProtocol();
});
```

### Stateful

A `Conversational` agent owns the history in the database, so the client only needs to send the newest message back:

```php
Route::post('/chat/{conversation}', function (Request $request, Conversation $conversation) {
    return (new ChatAgent)
        ->continue($conversation->id, as: $request->user()) // stored history replays automatically
        ->stream(Vercel::chat($request))
        ->usingVercelDataProtocol();
});
```

### Human in the loop, same route

When a stream pauses on a tool approval, the client answers with `addToolApprovalResponse` and re-submits. The approval responses ride back on the trailing assistant message, and `Vercel::chat()` resolves them into `Decisions`, so the exact same route handles both a fresh prompt and a resume:

```php
$chat = Vercel::chat($request);

$chat->message();   // newest UserMessage, or null on a resume turn
$chat->decisions(); // Decisions from the approval responses, or null on a prompt turn

(new ChatAgent)->stream($chat); // picks whichever the turn contains
```

`Chat` implements the `ChatInput` contract added in the previous PR, which is what makes that work.

### The building blocks

Everything `Vercel::chat()` composes is public, for when you need the pieces individually:

```php
Vercel::messagesFrom($request->input('messages')); // list<UserMessage|AssistantMessage>
Vercel::messageFrom($uiMessage);                   // one UI message array to a Message
```

`withMessages()` also normalizes raw UI message arrays, so a mixed history works:

```php
$agent->withMessages([
    ['role' => 'user', 'parts' => [['type' => 'text', 'text' => 'Hi']]], // UI message array
    new AssistantMessage('Hello!'),                                      // already a message
]);
```

File parts convert without ceremony: `data:` URLs become `Base64Image` / `Base64Audio` / `Base64Video` / `Base64Document` by mime type, remote URLs become their `Remote*` counterparts, and part filenames are preserved. Malformed parts and unknown roles are skipped rather than failing the request, since the payload is client-supplied.

Additive throughout — no breaking changes.

### DIFF
--- a/src/Promptable.php
+++ b/src/Promptable.php
@@ -37,6 +37,7 @@ use Laravel\Ai\Responses\QueuedAgentResponse;
 use Laravel\Ai\Responses\StreamableAgentResponse;
 use Laravel\Ai\Responses\StreamedAgentResponse;
 use Laravel\Ai\Streaming\Events\StreamEvent;
+use Laravel\Ai\Vercel\Vercel;
 use LogicException;
 use ReflectionClass;
 use RuntimeException;
@@ -276,7 +277,9 @@ trait Promptable
         }
 
         $this->adHocMessages = Collection::make($messages)
-            ->map(fn ($message) => Message::tryFrom($message))
+            ->flatMap(fn ($message) => is_array($message) && isset($message['parts'])
+                ? Vercel::messagesFrom([$message])
+                : [Message::tryFrom($message)])
             ->all();
 
         return $this;
--- a/src/Vercel/Chat.php
+++ b/src/Vercel/Chat.php
@@ -3,11 +3,11 @@
 namespace Laravel\Ai\Vercel;
 
 use Laravel\Ai\Approvals\Decisions;
-use Laravel\Ai\Contracts\ChatInput;
+use Laravel\Ai\Contracts\AgentInput;
 use Laravel\Ai\Messages\Message;
 use Laravel\Ai\Messages\UserMessage;
 
-class Chat implements ChatInput
+class Chat implements AgentInput
 {
     /**
      * @param  array<int, array<string, mixed>>  $messages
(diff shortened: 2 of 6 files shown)


vinitkadam03:
why were these reverted? I was planning to remove our custom converter and switch to this built in one and then saw that this was reverted later :sweat_smile: . do we have plans to add it again?

PR #919, opened 2026-08-21
Hydrate useChat from stored conversation messages

pushpak1300:
Builds on the previous PRs in the stack, which convert a `useChat` request into agent input. This PR adds the return trip: turning stored messages back into UI messages so `useChat` can be hydrated with an existing conversation.

```php
Route::get('/chat/{conversation}', fn (Conversation $conversation) => view('chat', [
    'conversation' => $conversation,
    'messages' => Vercel::uiMessagesFrom($conversation->messages),
]));
```

`uiMessagesFrom()` accepts both `ConversationMessage` models and plain `Message` objects, and renders a turn the way the client left it:

- Text becomes a `text` part; attachments become `file` parts, with stored files inlined as `data:` URLs and provider-hosted files skipped.
- Tool calls become `tool-{name}` parts. A call whose result is known settles as `output-available`, a denied one as `output-denied`, and a call still awaiting a decision renders as `approval-requested` with its reason attached — so a paused conversation reloads into the same approval prompt the user walked away from.
- Tool results are matched back to their call wherever it was rendered, whether they arrived on the same `ConversationMessage` row or in a separate `ToolResultMessage`.
- `ConversationMessage` rows keep their stored id, so client-side keys stay stable across a reload.

Also adds property annotations to `ConversationMessage` for static analysis of the attributes this reads.

Additive throughout — no breaking changes.

### DIFF
--- a/src/Models/ConversationMessage.php
+++ b/src/Models/ConversationMessage.php
@@ -6,6 +6,15 @@ use Illuminate\Database\Eloquent\Attributes\WithoutIncrementing;
 use Illuminate\Database\Eloquent\Model;
 use Illuminate\Database\Eloquent\Relations\BelongsTo;
 
+/**
+ * @property string $id
+ * @property string $role
+ * @property ?string $content
+ * @property ?array $attachments
+ * @property ?array $tool_calls
+ * @property ?array $tool_results
+ * @property ?array $approval_state
+ */
 #[WithoutIncrementing]
 class ConversationMessage extends Model
 {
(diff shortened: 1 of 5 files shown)


taylorotwell:
Renamed the Vercel message conversion methods to make the direction of each conversion clearer and keep the public API consistent:

- uiMessagesFrom() → toUiMessages()
- messagesFrom() → fromUiMessages()
- messageFrom() → fromUiMessage()

Issue #506, opened 2026-05-05
Tool schemas without ->required() properties fail OpenAI strict mode validation


paulkirow:
When defining tool schemas, any property without ->required() is excluded from the JSON Schema required array. 
Since the SDK always sends `'strict' => true` to OpenAI, every property in properties must also appear in required.  
Missing ->required() on any parameter produces an invalid schema that OpenAI rejects with a 400 invalid_function_parameters at runtime.

The key issue here is that ->required() must always be present.  There's no SDK-level validation to catch this early, and nothing in the docs mentions that ->required() is needed on every parameter.

Alternatively, it could always auto-populate required since `strict: true` is always set, just as `additionalProperties: false` is always set. Which would remove the need for developers to manually call ->required() on every field just to satisfy a constraint that's implicit in strict mode.

Unless I'm missing something here...

## Steps to reproduce:
1. Define a tool with a schema field that doesn't have ->required()
2. Call an agent with this tool attached using OpenAI as a provider

## Example Request/Response
Tool Schema
```
    public function schema(JsonSchema $schema): array
    {
        return [
            'query' => $schema->string(),
        ];
    }
```

```
{
  "model": "gpt-5.4-nano",
  "input": [
    {
      "role": "system",
      "content": "..."
    },
    {
      "role": "user",
      "content": [
        {
          "type": "input_text",
          "text": "..."
        }
      ]
    }
  ],
  "tool_choice": "auto",
  "tools": [
    {
      "type": "function",
      "name": "...",
      "description": "...",
      "strict": true,
      "parameters": {
        "type": "object",
        "properties": {
          "query": {
            "type": "string"
          }
        },
        "required": [],
        "additionalProperties": false
      }
    }
  ]
}
```

```
"response_status":400,"response_body":"{
  \"error\": {
    \"message\": \"Invalid schema for function '...': In context=(), 'required' is required to be supplied and to be an array including every key in properties. Missing 'query'.\",
    \"type\": \"invalid_request_error\",
    \"param\": \"tools[1].parameters\",
    \"code\": \"invalid_function_parameters\"
  }
}
```

Package & version: laravel/ai v0.6.6
PHP version: PHP 8.4

Highlights a similar issue mentioned in this comment: https://github.com/laravel/ai/issues/313#issuecomment-4169755019

PR #923, opened 2026-08-21
Honor the Strict attribute for structured output on the Groq, Mistral, OpenAI-compatible, and OpenRouter providers

sulimanbenhalim:
#530 wired the `#[Strict]` opt-in into the OpenAI provider, but the other providers that send a structured-output `strict` flag still hardcode `strict: true`, so they ignore the attribute.

same thing #506 was about: with `strict: true`, a schema that has any field without `->required()` gets a 400, since strict mode wants every property listed in `required`. so a user who never added `#[Strict]` still gets it forced on.

this threads `Strict::isAppliedTo($options?->agent)` into the response format for Groq, Mistral, OpenAI-compatible, and OpenRouter, same as the OpenAI fix. default is `strict: false` now (best-effort), `#[Strict]` turns it back on. added a without-attribute test per provider asserting strict false, the existing `#[Strict]` tests still assert true.

### DIFF
--- a/src/Gateway/Groq/Concerns/BuildsTextRequests.php
+++ b/src/Gateway/Groq/Concerns/BuildsTextRequests.php
@@ -3,6 +3,7 @@
 namespace Laravel\Ai\Gateway\Groq\Concerns;
 
 use Illuminate\Support\Arr;
+use Laravel\Ai\Attributes\Strict;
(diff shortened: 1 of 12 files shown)


pushpak1300:
Thanks

PR #924, opened 2026-08-21
Exclude cached tokens from the reported prompt token count on the OpenRouter provider

sulimanbenhalim:
OpenRouter drops raw `prompt_tokens` into `promptTokens` but also reports the cached portion in `cacheReadInputTokens`, so cache-read tokens get counted twice. Every other OpenAI-family provider already excludes them (OpenAI, DeepSeek in #896, Gemini, and Groq + OpenAI-compatible in #909), so OpenRouter is the last one still double-counting.

checked it against real OpenRouter responses: `cached_tokens` is a subset of `prompt_tokens` (e.g. `prompt_tokens` 10339 with `cached_tokens` 10318, and `total_tokens` = prompt + completion). cache write tokens are separate and not part of `prompt_tokens`, so this subtracts only `cached_tokens` and leaves `cacheWriteInputTokens` alone.

one line in the shared `extractUsage`, so streaming gets it too. flipped the existing usage assertion to the right value and added a streaming test that carries cache tokens.

### DIFF
--- a/src/Gateway/OpenRouter/Concerns/ParsesTextResponses.php
+++ b/src/Gateway/OpenRouter/Concerns/ParsesTextResponses.php
@@ -94,12 +94,15 @@ trait ParsesTextResponses
     protected function extractUsage(array $data): Usage
     {
         $usage = $data['usage'] ?? [];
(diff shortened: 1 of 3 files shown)


pushpak1300:
Thanks

PR #925, opened 2026-08-21
Fail over on Groq's flex tier capacity exceeded status code

sulimanbenhalim:
Groq's `overloadedStatusCodes()` doesn't include `498`, its own "flex tier capacity exceeded" status. so when a flex-tier request hits a capacity spike, the classifier falls through to a raw `RequestException` (not a `FailoverableException`), `withModelFailover` never catches it, and failover silently doesn't happen. the caller gets a raw Guzzle exception instead.

Groq documents 498 as transient ("the flex tier is at capacity and the request won't be processed. You can try again later."), same category as the 503 it already handles. this adds an `overloadedStatusCodes()` override mirroring the Anthropic 529 one from #884, so 498 maps to `ProviderOverloadedException` and fails over. added a 498 test next to the existing 503 one.

### DIFF
--- a/src/Gateway/Groq/GroqGateway.php
+++ b/src/Gateway/Groq/GroqGateway.php
@@ -42,8 +42,8 @@ class GroqGateway implements StepTextGateway, TranscriptionGateway
      */
     protected function overloadedStatusCodes(): array
     {
-        // 498 is Groq's "flex tier capacity exceeded" status, plus the shared transient gateway and Cloudflare codes.
-        return [498, 502, 503, 504, 520, 522, 524];
+        // The status codes Groq documents as transient: 498 is "flex tier capacity exceeded", alongside 502 and 503.
+        return [498, 502, 503];
     }
 
     /**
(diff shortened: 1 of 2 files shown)


src/Gateway/Groq/GroqGateway.php
```diff
@@ -35,6 +35,17 @@ class GroqGateway implements StepTextGateway, TranscriptionGateway
...
 
+    /**
+     * The status codes that indicate Groq is transiently unavailable and the request should fail over.
+     *
+     * @return list<int>
+     */
+    protected function overloadedStatusCodes(): array
+    {
+        // 498 is Groq's "flex tier capacity exceeded" status, plus the shared transient gateway and Cloudflare codes.
+        return [498, 502, 503, 504, 520, 522, 524];
```
pushpak1300: https://console.groq.com/docs/errors there is no documented 504, 522, 524

pushpak1300 (commented):
Lets support documented only. https://console.groq.com/docs/errors

src/Gateway/Groq/GroqGateway.php
sulimanbenhalim: done, trimmed to just the documented ones (498, 502, 503).

pushpak1300:
Thanks

PR #926, opened 2026-08-21
Add audio (TTS) support to the Mistral provider

SanderMuller:
Mistral released Voxtral TTS in March 2026 (`voxtral-mini-tts-2603`), but the SDK's Mistral provider only had text, embeddings, and transcription wired up — so Voxtral speech could not be used through `Audio::of(...)`. This adds audio (TTS) support to the existing provider. Purely additive: no new driver, no contract or config changes, and the existing `mistral` connection just gains the capability.

The endpoint is `POST /v1/audio/speech` with a JSON body, and the response returns the audio base64-encoded in an `audio_data` field (unlike the raw bytes OpenAI and ElevenLabs return), so the gateway passes it straight through to `AudioResponse`.

## Usage

```php
use Laravel\Ai\Audio;

// Default model and voice.
$response = Audio::of('Hello there!')->generate(provider: 'mistral');

// Explicit model, or a specific preset/custom voice id.
$response = Audio::of('Bonjour!')
    ->voice('fr_marie_neutral')
    ->generate(provider: 'mistral', model: 'voxtral-mini-tts-2603');

$response->store('audio');
```

## What's in it

- `MistralProvider` implements `AudioProvider`; the audio gateway reuses the shared, memoized `MistralGateway` instance like the other capabilities. Default model is `voxtral-mini-tts-2603` (overridable via `models.audio.default`) — Mistral documents no `-latest` alias for the TTS model, so the pinned name is used deliberately, unlike the aliased text/STT defaults.
- `MistralGateway` implements `AudioGateway` with `generateAudio()`:
  - `POST audio/speech` with `model`, `input`, `voice_id`, and `response_format: mp3` (returned as `audio/mpeg`).
  - `default-female` / `default-male` map to Mistral's `gb_jane_neutral` / `en_paul_neutral` presets (verified against the live voices endpoint); any other value passes through as a `voice_id` unchanged, so all 30 presets and custom (cloned) voices work.
  - A missing or empty `audio_data` throws instead of returning empty audio, mirroring the Gemini gateway. HTTP failures flow through the shared failover error handling.
  - `instructions` is deliberately not sent — the Mistral API has no such field (same as ElevenLabs); a test documents the omission.
- `resources/boost/skills/ai-sdk-development/SKILL.md` — adds Mistral to the TTS row of the provider support table.

Docs: https://docs.mistral.ai/capabilities/audio/text_to_speech

## Tests

12 cases in `tests/Feature/Providers/Mistral/AudioTest.php` covering request mapping, both default-voice aliases, custom voice pass-through, bearer auth, base64 pass-through + MIME/meta, the default model, the omitted-instructions contract, missing `audio_data`, and error/rate limit/overloaded responses — plus an audio case in the existing `BaseUrlTest.php` and a gateway-sharing assertion. Full suite green, Pint and PHPStan clean.

Also smoke-tested against the live API: both default presets return HTTP 200 with valid mp3 audio for exactly the request this gateway sends.

### DIFF
--- a/resources/boost/skills/ai-sdk-development/SKILL.md
+++ b/resources/boost/skills/ai-sdk-development/SKILL.md
@@ -487,7 +487,7 @@ Calling a capability not supported by a provider throws a `LogicException`. Refe
 | ---------- | --------------------------------------------------------------- |
 | Text       | OpenAI, Anthropic, Gemini, Azure, Groq, xAI, DeepSeek, Mistral, Ollama, OpenRouter, OpenAI-compatible |
 | Images     | OpenAI, Gemini, xAI                                            |
-| TTS        | OpenAI, ElevenLabs                                              |
+| TTS        | OpenAI, ElevenLabs, Mistral                                     |
 | STT        | OpenAI, ElevenLabs, Mistral, Groq, OpenAI-compatible            |
 | Embeddings | OpenAI, OpenAI-compatible, Gemini, Azure, Cohere, Mistral, Jina, VoyageAI |
 | Reranking  | Cohere, Jina                                                    |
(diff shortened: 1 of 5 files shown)


pushpak1300:
Thanks.

PR #927, opened 2026-08-21
Fetch and inline remote image, document, and audio attachments for the Gemini provider

sulimanbenhalim:
Gemini maps `Image::fromUrl()` / `Document::fromUrl()` / `Audio::fromUrl()` to a `fileData.fileUri` pointing at the raw public URL. but Gemini's `fileUri` only accepts a Files API URI (`files/...`) or a YouTube URL for video, so a public content URL just gets rejected. every other provider handles these `fromUrl` attachments fine.

this fetches the bytes and inlines them as base64 for the remote-URL image, document, and audio cases, matching what the Local/Stored arms already do and what the Anthropic gateway does for remote docs. remote video stays on `fileUri` since YouTube is the one case Gemini accepts there. added a per-type test.

### DIFF
--- a/src/Files/Concerns/HasRemoteContent.php
+++ b/src/Files/Concerns/HasRemoteContent.php
@@ -49,7 +49,7 @@ trait HasRemoteContent
      */
     protected function response(): Response
     {
-        return $this->response ??= Http::get($this->url);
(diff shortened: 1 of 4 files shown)


pushpak1300:
Thanks.

PR #929, opened 2026-08-21
Add a voices listing API for audio generation

SanderMuller:
The SDK can generate speech with a chosen `voice`, but it gives applications no way to discover which voices exist. Anyone building a voice picker, or validating a stored voice id, has to hand-roll HTTP calls per provider today. This adds a small listing API for the two supported providers with a dynamic voice catalogue: ElevenLabs (`GET /v1/voices`) and Mistral (`GET /v1/audio/voices`).

Every returned `Voice->id` can be passed directly to the audio builder's `voice()` method. That holds for ElevenLabs voice ids and Mistral voice slugs, both checked against the live APIs.

## Usage

```php
use Laravel\Ai\Audio;

$voices = Audio::voices(provider: 'eleven');

foreach ($voices as $voice) {
    $voice->id;        // 'gb_jane_neutral', pass straight to ->voice(...)
    $voice->name;      // 'Jane - Neutral'
    $voice->gender;    // 'female', or null when the provider does not expose it
    $voice->languages; // ['en_gb']
}

$voices->find('gb_jane_neutral');

// Without an explicit provider, the default audio provider is used:
$voices = Audio::voices();
```

## Design notes

`VoiceProvider` is a new opt-in contract rather than an addition to `AudioProvider`. Extending `AudioProvider` would break OpenAI, Gemini, and OpenRouter, which have no voices endpoint, and providers with a fixed documented voice list gain nothing from an HTTP round-trip.

The plumbing mirrors the Files feature: a `VoiceGateway` contract, `HasVoiceGateway` and `ListsVoices` provider concerns, `AiManager::voiceProvider()` / `fakeableVoiceProvider()` with the same `LogicException` guard, and shared gateway instances (Mistral reuses its memoized gateway across all capabilities).

Two things worth calling out:

- Mistral caps `limit` at 100 and paginates via `offset`. The documented `page` parameter is ignored server-side (confirmed with live requests: `page=2` returns the first page, `offset=10` returns the second). The gateway follows `offset` until the reported `total` is collected, and if `total` is ever absent it continues while pages come back full.
- `VoicesResponse` follows the `RerankingResponse` idiom (iterable, countable, `first()`, `find()`, `collect()`). Its `Meta` carries the provider and a `null` model, since a listing has no model — `Meta` already supports this. And `Audio::voices()` takes the provider as a direct argument because a listing has no builder step; without one it resolves `config('ai.default_for_audio')`, matching `Audio::of(...)->generate()`.

## Testing

```php
Audio::fakeVoices(); // two default voices, zero-config

Audio::fakeVoices([new Voice('custom-voice', 'Custom Voice', 'female', ['nl'])]);

Audio::fakeVoices(fn ($provider) => [...]); // per-provider

Audio::assertVoicesListed('eleven');
Audio::assertVoicesNotListed('mistral');
Audio::assertNoVoicesListed();
```

`assertVoicesListed` and `assertVoicesNotListed` accept a provider-name string as shorthand for the closure form, following `assertFileDeleted`.

18 tests cover request mapping and auth per provider, null-safe field mapping for voices without labels or languages, `find()`, error propagation, a 150-voice two-request pagination case, default-provider resolution, the fake layer (defaults, explicit voices, closure, an explicitly empty catalogue), and the `LogicException` for providers without voice support. Full suite green, Pint and PHPStan clean. The `resources/boost/skills` doc is updated.

### DIFF
--- a/resources/boost/skills/ai-sdk-development/SKILL.md
+++ b/resources/boost/skills/ai-sdk-development/SKILL.md
@@ -199,6 +199,11 @@ $audio = Audio::of('Hello from Laravel.')
     ->generate();
 
 $path = $audio->store();
+
+// List the voices available for audio generation (ElevenLabs, Mistral)
+$voices = Audio::voices(provider: 'eleven');
+
+$voices->first()->id; // Pass directly to ->voice(...)
 ```
 
 ### Transcription
@@ -384,6 +389,11 @@ Image::assertNothingGenerated();
 // Audio
 Audio::fake();
 Audio::assertGenerated(fn ($prompt) => $prompt->contains('Hello'));
+
+// Voices
+Audio::fakeVoices();
+Audio::assertVoicesListed('eleven');
+Audio::assertNoVoicesListed();
 
 // Transcription
 Transcription::fake(['Transcribed text.']);
--- a/src/AiManager.php
+++ b/src/AiManager.php
@@ -14,6 +14,7 @@ use Laravel\Ai\Contracts\Providers\RerankingProvider;
 use Laravel\Ai\Contracts\Providers\StoreProvider;
 use Laravel\Ai\Contracts\Providers\TextProvider;
 use Laravel\Ai\Contracts\Providers\TranscriptionProvider;
+use Laravel\Ai\Contracts\Providers\VoiceProvider;
 use Laravel\Ai\Enums\Lab;
 use Laravel\Ai\Gateway\Anthropic\AnthropicGateway;
 use Laravel\Ai\Gateway\Gemini\GeminiGateway;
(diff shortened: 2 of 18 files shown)


SanderMuller:
Related: #926 adds Mistral TTS generation; both touch `MistralProvider`/`MistralGateway`, so whichever lands second will have a small conflict. Happy to rebase either.

pushpak1300:
Can you fix conflicts ? 

I'm not sure if this needs to merged yet and how much difficult it will be to maintain it but i will review once it's conflict free.

SanderMuller:
Conflicts resolved & merged `0.x` in (the overlap was with #926, which touched the same Mistral provider/gateway). The Mistral provider now implements both audio generation and voice listing on its shared gateway.

pushpak1300:
@SanderMuller discussed this with team we are deferring this for now.

PR #932, opened 2026-08-23
Validate an approval resumption before executing the approved tools

### LINKED ISSUE #931: Approval resume executes tools before checking the paused turn belongs to the user, leaving an executed-but-unrecorded result
## Summary

When a paused tool approval is resumed by an agent whose user doesn't match the one that paused the conversation, the approved tools run first and the mismatch is only caught afterwards, when the results are written to the conversation store. The end state is that the tool has executed, but the conversation still shows the turn as waiting for approval — and approving it again runs the tool a second time.

## Versions

laravel/ai v0.11.0; the same code is on main.

## What happens

1. An agent using RemembersConversations is started with forUser($userA) and pauses on a tool that requires approval. The paused assistant message is stored with participant_type / participant_id for user A.
2. Later, the conversation is resumed with continue($conversationId) (no user) or continue($conversationId, $userB), and the approval decision is passed to prompt().
3. RemembersConversations::messages() loads the conversation by id only, so the paused message and its pending tool calls come back as normal. validateApproval() is satisfied because the tool call ids match.
4. TextGenerationLoop calls resumeFromApproval(), which executes the approved tools.
5. Only then does it call the approval-results recorder. DatabaseConversationStore::storeApprovalResults() looks the paused message up with the participant filter, finds no match, and throws ApprovalMismatchException. Its transaction rolls back, so nothing is written.

So after the exception:

- the tool has run (in my case a consequential one — the whole reason it needed approval),
- the assistant message still has approval_state pending,
- when the conversation is loaded again, the same approval request is presented again, and approving it runs the tool again.

## Why I think this is a bug rather than just a misconfiguration

Resuming as the wrong user is a host mistake, and I'm fine with it being rejected. The problem is that the rejection happens after the side effect. The write side (storeApprovalResults) enforces "this participant owns the paused message", but nothing on the read side or in the loop checks that before execution, and the store contract doesn't offer a way to. The check that catches the mistake is cheap and could run first.

## What I'd expect

A mismatch should throw ApprovalMismatchException before any tool executes, leaving the message pending and the conversation consistent — the same I'm opening a PR for this alongside the issue. The approach: validate the resumption before executing, reusing the same lookup that rejects it afterwards.

- Add ConversationStore::hasPausedTurn(string $conversationId, ?string $participantType, string|int|null $participantId, array $toolCallIds): bool.
- In DatabaseConversationStore, extract the existing lookup from storeApprovalResults() into a shared private method; hasPausedTurn() uses it without the lock, storeApprovalResults() keeps using it under lockForUpdate() as today.
- In ResumesToolApprovals, add a validator closure built the same way as the existing results recorder (same gating on RemembersConversations + a current conversation + a real gateway).
- TextGenerationLoop::generate() and the streaming path accept that closure and call it before resumeFromApproval(), passing the pending tool-call ids validateApproval() already has.

Hosts that never hit a mismatch see no change. Custom ConversationStore implementations need to implement the one new method — that's the real cost, and I'm open to shaping it differently if you'd prefer.

The PR includes two tests: pause under user A, resume under B, assert ApprovalMismatchException with the tool executed zero times and the message still pending; and the positive twin under A, executing once.

One thing this deliberately doesn't address: two simultaneous resumes by the correct user can still both execute. That needs a claim-before-execute design with crash recovery, which is much larger than this issue.

fissible:
Fixes #931.

When a paused approval is resumed by an agent whose user doesn't own the paused turn, the approved tools ran first and the mismatch was only caught when storeApprovalResults() tried to write the results. Its transaction rolled back, so the tool had executed but the turn still showed as pending — and approving it again ran the tool a second time.

This checks the resumption against the conversation store before anything executes, using the same lookup that rejects it afterwards:

- ConversationStore::hasPausedTurn() is added; DatabaseConversationStore shares the query between it and storeApprovalResults() (the latter still under lockForUpdate()).
- ResumesToolApprovals::ensureApprovalResumptionIsRecordable() sits next to the existing results recorder and uses the same participant resolution.
- Both the text and streaming paths call it right after the eager validateApproval(). generate() now accepts the validated approval like stream() already does, so a gated tool's shouldRequestApproval() is still called exactly once.

A mismatch throws the same ApprovalMismatchException, just before any tool runs, leaving the turn pending for the right user.

Tests: resume as a different user, and resume without a user on a user-owned pause — both assert the exception, zero tool invocations, the turn still pending, and that the owner can then resume and execute once. Plus a store-level test for hasPausedTurn(). Custom ConversationStore implementations need to add the one method (the three anonymous stores in the test suite did).

Not addressed here, deliberately: two simultaneous resumes by the correct user can still both execute — that needs a claim-before-execute design and is out of scope.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,49 @@
 # Release Notes
 
-## [Unreleased](https://github.com/laravel/ai/compare/v0.11.0...0.x)
+## [Unreleased](https://github.com/laravel/ai/compare/v0.11.2...0.x)
+
+## [v0.11.2](https://github.com/laravel/ai/compare/v0.11.1...v0.11.2) - 2026-09-03
+
+### What's Changed
+
+* Copy uploaded files to a managed temp path so their contents survive the request by [@sulimanbenhalim](https://github.com/sulimanbenhalim) in https://github.com/laravel/ai/pull/959
+* Preserve tool result failure status across conversations by [@drewmt](https://github.com/drewmt) in https://github.com/laravel/ai/pull/812
+* Fix invalid Mistral cheapest text model default by [@kachelle](https://github.com/kachelle) in https://github.com/laravel/ai/pull/980
+
+### New Contributors
+
+* [@drewmt](https://github.com/drewmt) made their first contribution in https://github.com/laravel/ai/pull/812
+
+**Full Changelog**: https://github.com/laravel/ai/compare/v0.11.1...v0.11.2
+
+## [v0.11.1](https://github.com/laravel/ai/compare/v0.11.0...v0.11.1) - 2026-09-02
+
+### What's Changed
+
+* Align the Vercel UI message stream with the data stream protocol by [@pushpak1300](https://github.com/pushpak1300) in https://github.com/laravel/ai/pull/906
+* Allow caching embeddings individually by [@timacdonald](https://github.com/timacdonald) in https://github.com/laravel/ai/pull/905
+* Add v0.11 upgrade guide by [@pushpak1300](https://github.com/pushpak1300) in https://github.com/laravel/ai/pull/897
+* Fix Vercel provider tool events and finish reasons by [@pushpak1300](https://github.com/pushpak1300) in https://github.com/laravel/ai/pull/921
+* Emit citation events for Gemini grounding results while streaming by [@sulimanbenhalim](https://github.com/sulimanbenhalim) in https://github.com/laravel/ai/pull/912
+* Emit citation events for xAI web search results while streaming by [@sulimanbenhalim](https://github.com/sulimanbenhalim) in https://github.com/laravel/ai/pull/911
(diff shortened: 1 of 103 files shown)


fissible:
Following up on this one, still very much wanted on my end. In practice, I maintain Verdict, an approval-gate layer on top of laravel/ai, and this is exactly the kind of failure it has to defend against: a queued approval resuming against the participant's most recently updated conversation instead of the one that was actually paused. Under concurrency, that can execute the approved tool in the wrong context.

I'm shipping a defensive reference pattern on the Verdict side that resumes the specific paused conversation ID, but validating the resumption in the framework is still the right place for the guarantee.

Happy to rebase or address any review notes if that helps move it forward.

pushpak1300:
Thanks for the thorough writeup — I went a different way in #1046 and removed the participant match from `storeApprovalResults` entirely, since the filter never guarded anything the read path or the loop respected and only fired after the tool had already run. Authorising who may approve a paused turn belongs in the application, alongside the check that already authorises the chat route.

PR #935, opened 2026-08-24
Report the transcribed audio duration on transcription usage

JulBeg:
Transcription models are commonly billed per minute of audio, and `whisper-1` reports no token counts at all, so there is currently nothing for an application to record its spend against.

Transcription responses now carry an `AudioUsage`, a `Usage` subclass that adds the duration:

```php
$response = Transcription::fromPath('meeting.mp3')->generate(provider: 'openai');

// The length of the audio, in seconds.
$response->usage->durationSeconds; // 2.11

// Duration billed models bill whole seconds.
ceil($response->usage->durationSeconds); // 3
```

`TranscriptionResponse::$usage` is typed `Usage`, so text generation, embeddings, streams, and persisted conversation messages are untouched.

It is populated wherever the provider documents a duration: `duration` falling back to `usage.seconds` for OpenAI, OpenAI-compatible and OpenRouter, `usage.prompt_audio_seconds` for Mistral, `audio_duration_secs` for ElevenLabs, and `duration` from Groq's `verbose_json`. Everything else reports `0.0`. `Transcription::fake()` reports the length of the segment it invents.

### DIFF
--- a/src/Gateway/ElevenLabsGateway.php
+++ b/src/Gateway/ElevenLabsGateway.php
@@ -11,9 +11,9 @@ use Laravel\Ai\Contracts\Providers\AudioProvider;
 use Laravel\Ai\Contracts\Providers\TranscriptionProvider;
 use Laravel\Ai\Files\Audio;
 use Laravel\Ai\Responses\AudioResponse;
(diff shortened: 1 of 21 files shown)


pushpak1300:
closing because of #1025

PR #945, opened 2026-08-26
Store uploaded files from their path instead of base64 encoding them

### LINKED ISSUE #943: Files::put() base64-encodes an UploadedFile that already has a real path, tripling peak memory
### Description

`Files::put()` converts an `UploadedFile` to a `Base64Document` even though an `UploadedFile` always has a readable path on disk. The base64 round-trip is pure overhead: it raises peak memory to **2.67× the file size**, where reading the path directly costs 1.00×.

Verified on `main` (also present in v0.11.0).

### Current behaviour

```php
// src/Files.php
$file = match (true) {
    is_string($file) => new Base64Document(base64_encode($file), $mimeType),
    $file instanceof UploadedFile => Base64Document::fromUpload($file)->as($file->getClientOriginalName()),
    default => $file,
};
```

```php
// src/Files/Base64Document.php
public static function fromUpload(UploadedFile $file, ?string $mimeType = null): self
{
    return new self(base64_encode($file->getContent()), ...);   // full file -> base64 in memory
}

public function content(): string
{
    return base64_decode($this->base64);                        // decoded back for the upload
}
```

So the bytes are read in full, expanded to base64 (~1.33×) and held on the object, then decoded back to a second full copy when `prepareStorableFile()` calls `content()`. `Document::fromUpload()` has the same body.

### Measurement

PHP 8.4, measured with `memory_reset_peak_usage()` / `memory_get_peak_usage(true)` around each path, up to the `content()` call that `PreparesStorableFiles::prepareStorableFile()` makes:

```
file size          : 33554432 bytes (32.0 MiB)
Base64Document     : peak +89481216 bytes (2.67x)
LocalDocument      : peak +33566720 bytes (1.00x)
reduction          : 2.67x
```

Same 2.67× / 1.00× ratio at 8 MiB, so the overhead is proportional to file size, not a constant.

Reproduction script:

```php
$path = '/tmp/memtest.bin';
file_put_contents($path, random_bytes(32 * 1024 * 1024));

$measure = function (callable $fn) {
    gc_collect_cycles();
    $before = memory_get_usage(true);
    memory_reset_peak_usage();
    $fn();
    return memory_get_peak_usage(true) - $before;
};

// current path
$a = $measure(function () use ($path) {
    $u = new UploadedFile($path, 'x.bin', 'application/octet-stream', null, true);
    return Base64Document::fromUpload($u)->as('x.bin')->content();
});

// path-based equivalent
$b = $measure(fn () => (new LocalDocument($path, 'application/octet-stream'))->as('x.bin')->content());
```

### Impact

On a shared FPM pool this is the difference between one upload costing 32 MiB and costing 85 MiB of the worker's `memory_limit`. A handful of concurrent uploads can push unrelated requests in the same pool into an OOM that has nothing to do with the file being uploaded.

### Suggested fix

`LocalDocument` already implements `StorableFile`, takes an optional MIME type, and supports `->as()`, so the mapping can preserve both the client MIME type and the original filename:

```php
$file instanceof UploadedFile => (new LocalDocument($file->getRealPath(), $file->getClientMimeType()))
    ->as($file->getClientOriginalName()),
```

`content()` then becomes a single `file_get_contents()`. One caveat worth handling: `getRealPath()` returns `string|false`, and `false` would reach `LocalDocument`'s `blank($path)` guard as an empty string, so the failure message would be about an empty path rather than a missing upload.

This does not make uploads streaming — `HasContent::content(): string` still forces one full copy into memory — but it removes the two extra copies without any contract change. I have opened a separate issue for the streaming side so the two can be judged independently.

Happy to send a PR if the direction looks right.

lazerg:
`Files::put()` turns an `UploadedFile` into a `Base64Document`, but an uploaded file always sits on disk with a readable path already, so the base64 round-trip buys nothing. The bytes get read in full, expanded ~1.33x into base64 and held on the object, then decoded back into a second full copy when the gateway calls `content()`.

Measured on a 32 MiB upload up to that `content()` call: `Base64Document` peaks at 2.67x the file size, `LocalDocument` at 1.00x. Same ratio at 8 MiB, so it scales with the file rather than being a fixed cost. On a shared FPM pool that turns a 32 MiB upload into 85 MiB of the worker's `memory_limit`, which is enough to push unrelated requests in the same pool into an OOM.

`LocalDocument` already implements `StorableFile`, accepts a MIME type and supports `->as()`, so the client MIME type and original filename both survive the swap and `content()` becomes a single `file_get_contents()`. `getClientMimeType()` is non-nullable, so `LocalDocument` never falls back to sniffing the file, and `->as()` always sets the name, so it never falls back to `basename()`. All three gateways call `content()` synchronously inside `putFile()`, so nothing outlives the request's temp file.

The path comes from `getPathname()` rather than `getRealPath()` on purpose. When an upload fails with something like `UPLOAD_ERR_NO_TMP_DIR`, PHP leaves `tmp_name` empty, and `realpath('')` resolves to the current working directory instead of returning `false`, which would have quietly stored an empty file under the client's filename. `getPathname()` stays empty and hits `LocalDocument`'s blank-path guard, so a failed upload still throws the way it does today.

This doesn't make uploads streaming, `content(): string` still forces one full copy, but it drops the two extra ones without changing any contract.

Fixes #943

### DIFF
--- a/src/Files.php
+++ b/src/Files.php
@@ -33,7 +33,10 @@ class Files
     {
         $file = match (true) {
             is_string($file) => new Base64Document(base64_encode($file), $mimeType),
-            $file instanceof UploadedFile => (new LocalDocument($file->getRealPath() ?: $file->getPathname(), $file->getClientMimeType()))->as($file->getClientOriginalName()),
+            $file instanceof UploadedFile => (new LocalDocument(
+                $file->getPathname(),
+                $file->getClientMimeType(),
+            ))->as($file->getClientOriginalName()),
             default => $file,
         };
 
--- a/src/Files.php
+++ b/src/Files.php
@@ -7,6 +7,7 @@ use Illuminate\Http\UploadedFile;
 use Laravel\Ai\Contracts\Files\StorableFile;
 use Laravel\Ai\Files\Base64Document;
 use Laravel\Ai\Files\Document;
+use Laravel\Ai\Files\LocalDocument;
 use Laravel\Ai\Gateway\FakeFileGateway;
 use Laravel\Ai\Responses\FileResponse;
 use Laravel\Ai\Responses\StoredFileResponse;
@@ -32,7 +33,7 @@ class Files
     {
         $file = match (true) {
             is_string($file) => new Base64Document(base64_encode($file), $mimeType),
-            $file instanceof UploadedFile => Base64Document::fromUpload($file)->as($file->getClientOriginalName()),
+            $file instanceof UploadedFile => (new LocalDocument($file->getRealPath() ?: $file->getPathname(), $file->getClientMimeType()))->as($file->getClientOriginalName()),
             default => $file,
         };
 
--- a/src/Store.php
+++ b/src/Store.php
@@ -8,7 +8,7 @@ use Laravel\Ai\Contracts\Files\HasProviderId;
 use Laravel\Ai\Contracts\Files\StorableFile;
 use Laravel\Ai\Contracts\Providers\FileProvider;
 use Laravel\Ai\Contracts\Providers\StoreProvider;
-use Laravel\Ai\Files\Base64Document;
+use Laravel\Ai\Files\LocalDocument;
 use Laravel\Ai\Files\ProviderDocument;
 use Laravel\Ai\Responses\AddedDocumentResponse;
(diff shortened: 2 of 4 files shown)


pushpak1300:
Thanks.

PR #947, opened 2026-08-26
Add tools that ask the caller for the values they need

### LINKED ISSUE #770: 🖱️Feature Request: Add support for client-executed tools (browser-side tools)

eznix86:
Some tool arguments only the user can supply. The address to ship to, the file to pick, the location of the browser. Today the model must guess them, or the app must collect them before the turn starts.

A tool now declares what it needs. The run pauses until the caller submits it.

```php
class PickAddress implements NeedsInput, Tool
{
    public function needsInput(JsonSchema $schema, Request $request): array
    {
        return ['address_id' => $schema->integer()->enum(Auth::user()->addresses->pluck('id')->all())->required()];
    }

    {
        return new Response("Shipping to {$request['address_id']}.", data: ['chosen' => $request['address_id']]);
    }
}
```

The tool runs without a pause when the model already supplies the value.

## Read the pause

```php
$response = $agent->forUser($user)->prompt('Ship my order.');

if ($response->hasPendingApprovals()) {
    return $response->pendingApprovals; // id, tool, arguments, reason, schema, data
}
```

The `schema` is JSON Schema, so a client can render the form from it. The `data` is anything else the client needs to draw the card.

Resume with the values. The keys merge into the call the model made.

```php
$agent->continue($response->conversationId, $user)->prompt(
    Decisions::from(['call_2' => Decision::submit(['address_id' => 3])])
);
```

Reload a paused conversation later with `pendingApprovalsFor($conversationId)` on the conversation store.

## Read the pause on the UI

A stream emits `Streaming\Events\ToolApprovalRequest`, which carries the same pending approvals.

```php
return $agent->forUser($user)->stream('Ship my order.')->usingVercelDataProtocol();
```

The Vercel protocol sends a `tool-approval-request` part per paused call. It carries `approvalDescriptor` (the `data`) and `inputSchema` (the `schema`).

> [!NOTE]
> The AI SDK `addToolApprovalResponse` only approves or denies. Submit values from your own endpoint.

Any tool can also return a `Tools\Response` with `data`. The client reads it, and the model never does.

> [!WARNING]
> `InvokesTools::executeTool()` returns `Tools\Response` instead of `string`. `Response` is `Stringable`, so callers that cast keep working.

### DIFF
--- a/src/Approvals/Approval.php
+++ b/src/Approvals/Approval.php
@@ -4,13 +4,37 @@ namespace Laravel\Ai\Approvals;
 
 class Approval
 {
-    public function __construct(public readonly ?string $reason = null) {}
+    /**
+     * @param  array<string, mixed>|null  $ui
+     */
+    public function __construct(
+        public readonly ?string $reason = null,
+        public readonly ?array $ui = null,
+    ) {}
 
     /**
-     * Create a required approval.
+     * Create a required approval, answered with an approval or a rejection.
      */
     public static function required(?string $reason = null): self
     {
         return new self($reason);
+    }
+
+    /**
+     * Create an approval carrying a payload for the client, answered with a submission.
+     *
+     * @param  array<string, mixed>  $ui
+     */
+    public static function input(array $ui, ?string $reason = null): self
+    {
+        return new self($reason, $ui);
+    }
+
+    /**
+     * Determine whether the pause is waiting on a client submission.
+     */
+    public function isInteractive(): bool
+    {
+        return $this->ui !== null;
     }
 }
--- a/src/Approvals/Decision.php
+++ b/src/Approvals/Decision.php
@@ -32,11 +32,23 @@ class Decision
     }
 
     /**
+     * Approve the pending tool call with its arguments replaced wholesale.
+     *
      * @param  array<string, mixed>  $arguments
      */
     public static function edit(array $arguments): self
     {
         return new self('edit', arguments: $arguments);
+    }
+
+    /**
+     * Approve the pending tool call with the given arguments merged into its own.
+     *
+     * @param  array<string, mixed>  $arguments
+     */
+    public static function submit(array $arguments): self
+    {
+        return new self('submit', arguments: $arguments);
     }
 
     /**
@@ -81,6 +93,10 @@ class Decision
                 throw new InvalidArgumentException('The wildcard decision may not use the edit action.');
             }
 
+            if ($id === '*' && $decision->isSubmitted()) {
+                throw new InvalidArgumentException('The wildcard decision may not use the submit action.');
+            }
+
             $normalized[$id] = $decision;
         }
 
@@ -109,5 +125,13 @@ class Decision
     public function isEdited(): bool
     {
         return $this->action === 'edit';
+    }
+
+    /**
+     * Determine whether the tool call was answered with a client submission.
+     */
+    public function isSubmitted(): bool
+    {
+        return $this->action === 'submit';
     }
 }
--- a/src/Approvals/PendingApproval.php
+++ b/src/Approvals/PendingApproval.php
@@ -8,12 +8,14 @@ class PendingApproval implements Arrayable
 {
     /**
      * @param  array<string, mixed>  $arguments
+     * @param  array<string, mixed>|null  $ui
      */
     public function __construct(
         public readonly string $id,
         public readonly string $tool,
         public readonly array $arguments,
         public readonly ?string $reason = null,
+        public readonly ?array $ui = null,
     ) {}
 
     /**
@@ -26,6 +28,7 @@ class PendingApproval implements Arrayable
             'tool' => $this->tool,
             'arguments' => $this->arguments,
             'reason' => $this->reason,
+            ...($this->ui === null ? [] : ['ui' => $this->ui]),
         ];
     }
 }
--- a/src/Contracts/Interactive.php
+++ b/src/Contracts/Interactive.php
@@ -1,0 +1,15 @@
+<?php
+
+namespace Laravel\Ai\Contracts;
+
+use Laravel\Ai\Tools\Request;
+
+interface Interactive
+{
+    /**
+     * Get the payload the client renders, or null to run handle() without a round-trip.
+     *
+     * @return array<string, mixed>|null
+     */
+    public function ask(Request $request): ?array;
+}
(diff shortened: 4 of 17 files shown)


pushpak1300:
Thanks for this — the problem is real (it's the same one as #770) 

I want to land it, but with a smaller surface and actually wired into the loop. As it stands `present()` is never called anywhere in `src/`, so no surface reaches a pending approval, a stream event, a tool result, or the Vercel protocol; only `Decision::merge()` is live.

### What to keep

- `Decision::merge()` semantics. Rename to `Decision::submit()` — it's what a user does to a form, and `edit` keeps meaning "replace wholesale".
- The idea that a paused call and a finished call can both carry something for the client.

### What to drop

- `Surface` interface, `Presentable`, `Surface::actions()`. The client already owns its components and keys them by tool name (that's how the AI SDK, CopilotKit, Apps SDK all work). A class hierarchy plus an actions allow-list is ceremony the client never reads, and the model never sees it either.

### Proposed shape

Three kinds of tool:

| Kind | Declares | Behaviour |
|---|---|---|
| Backend | `Tool` | `handle()` runs immediately |
| Backend, gated | `Tool` + `Approvable` | yes/no pause, then `handle()` (today) |
| Frontend | `Tool` + `Interactive` | pause with a client payload, client submits input, then `handle()` |

Display-only UI is not a kind — any tool can return a `Response` with a `ui` payload.

```php
// Display only
{
    $product = Product::findOrFail($request['id']);

    return new Response(
        text: "{$product->name}, {$product->price_label}",   // what the model reads
        ui: $product->only('name', 'price', 'image_url'),    // what the client renders, never sent to the model
    );
}
```

```php
interface Interactive
{
    /** Payload for the client to render, or null to run handle() without a round-trip. */
    public function ask(Request $request): ?array;
}

class PickAddress implements Tool, Interactive
{
    public function ask(Request $request): ?array
    {
        return ['options' => $request->user()->addresses->map->only('id', 'label')->all()];
    }

    {
        $data = $request->validate(['address_id' => 'required|integer']);

        return new Response("User chose address {$data['address_id']}", ui: ['chosen' => $data['address_id']]);
    }
}

// Browser-executed (geolocation) is the same contract with an empty schema; handle() validates what came back.
```

```php
// Resume
$agent->continue($id, $user)->stream(
    Decisions::from(['call_2' => Decision::submit(['address_id' => 3])])
);
```

Rules that fall out of this:

- `schema()` is what the model sends. `ask()` sees those args and decides what to show. `handle()` gets model args merged with the client's submission and validates the client half. With strict mode the model can't fabricate `address_id`.
- The client submits **arguments, never results**. `handle()` always runs — it's the trust boundary. A raw client result would be prompt injection with tool-result authority.
- An `Interactive` pause refuses a bare or wildcard `approve`; you have to submit.
- `Interactive` compiles to the existing pause: `approvalForTool()` returns `Approval::input($ui)`, which flows into `PendingApproval->ui` with no other loop change. If a tool implements both contracts, `Interactive` wins.

### Wire

Native SSE keeps every event type; `tool_result` gains `ui`, each item in `tool_approval_request` gains `ui`. Vercel protocol fills the slots the AI SDK already has:

```json
{"type":"tool-approval-request","toolCallId":"call_2","approvalId":"call_2","approvalDescriptor":{"options":[...]}}
{"type":"tool-output-available","toolCallId":"call_1","output":{"name":"Widget","price":1900}}
```

Happy to pair on it or take the rework myself if you'd rather — say which.

eznix86:
Thank you @pushpak1300 for the review. I'll refactor it!

eznix86:
Refactored it to the asked shape.

`Presentable` are gone, `Decision::submit()` replaces merge`()`, and `Interactive::ask()` compiles to `Approval::input($ui)` so a pause reaches `PendingApproval->ui`. `Response(text:, ui:)` covers display-only, ui never reaches the model, and `handle()` always runs. 


Wired into tool_result, tool_approval_request and the Vercel approvalDescriptor/output slots.

Still happy to pair on it if you'd rather go through it together.

pushpak1300:
Can you please update the final pr description ?

eznix86:
> Can you please update the final pr description ?

done!

oliverbj:
Following this...

PR #963, opened 2026-08-31
Simplify provider options and stream protocol plumbing

pushpak1300:
This is a cleanup with no behaviour change. It removes code that was repeated in several places and keeps one copy.

### DIFF
--- a/src/Streaming/Protocols/StreamProtocol.php
+++ b/src/Streaming/Protocols/StreamProtocol.php
@@ -31,7 +31,13 @@ abstract class StreamProtocol
      *
      * @return array<string, string>
      */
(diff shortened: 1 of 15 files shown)


src/Streaming/Protocols/StreamProtocol.php
```diff
@@ -29,7 +29,13 @@ abstract protected function maskedErrorParts(): Generator;
      *
      * @return array<string, string>
      */
-    abstract protected function headers(): array;
+    protected function headers(): array
```
pascalbaljet, line 34: Can you add `'X-Accel-Buffering' => 'no'` here too, same as #960?

PR #913, opened 2026-08-20
Add AG-UI protocol streaming

pushpak1300:
Currently the only wire protocol an agent stream can speak is Vercel's, so there isn't a way to serve an AG-UI client (CopilotKit, the AG-UI SDKs) without hand translating events in the application:

```php
// The AG-UI client receives Laravel AI's own event shapes and can't parse them...
return $agent->stream($request->string('prompt'));

// Or you translate every event yourself, per app...
return response()->stream(function () use ($agent, $request) {
    foreach ($agent->stream($request->string('prompt')) as $event) {
        yield 'data: '.json_encode(translate($event))."\n\n";
    }
}, headers: ['Content-Type' => 'text/event-stream']);
```

This PR lets you write:

```php
return $agent->stream($request->string('prompt'))
    ->usingProtocol(new \Laravel\Ai\Streaming\Protocols\AgUiProtocol(
        threadId: $request->input('threadId'),
        runId: $request->input('runId'),
    ));
```

Both arguments are optional. `threadId` falls back to the conversation ID and then to a generated ULID, and `runId` falls back to the invocation ID:

```php
return $agent->forUser($request->user())
    ->stream($request->string('prompt'))
    ->usingProtocol(new \Laravel\Ai\Streaming\Protocols\AgUiProtocol()); // threadId is the conversation, runId the invocation
```

The response is a plain AG-UI event stream, so `HttpAgent` consumes it directly:

```
Content-Type: text/event-stream

data: {"type":"RUN_STARTED","threadId":"thread-1","runId":"run-1"}
data: {"type":"STEP_STARTED","stepName":"1"}
data: {"type":"TOOL_CALL_START","toolCallId":"call-1","toolCallName":"GetWeather"}
data: {"type":"TOOL_CALL_ARGS","toolCallId":"call-1","delta":"{\"city\":\"Lisbon\"}"}
data: {"type":"TOOL_CALL_END","toolCallId":"call-1"}
data: {"type":"TOOL_CALL_RESULT","messageId":"result-1","toolCallId":"call-1","content":"sunny","role":"tool"}
data: {"type":"STEP_FINISHED","stepName":"1"}
data: {"type":"STEP_STARTED","stepName":"2"}
data: {"type":"TEXT_MESSAGE_START","messageId":"msg-2","role":"assistant"}
data: {"type":"TEXT_MESSAGE_CONTENT","messageId":"msg-2","delta":"Sunny."}
[...]

PR #965, opened 2026-08-31
Add AG-UI approval interrupts and snapshot hydration to streams

pushpak1300:
Third of a four-PR stack that splits #939. Base: #964.

Builds on the AG-UI protocol streaming added in #913.

**Approval interrupts carry their tool.** `interrupts()` previously built the interrupt payload inline and emitted only an id and a message. It now delegates to `AgentUserInteraction::interrupt()` (added in #964), so a paused run tells the client *which* tool wants to run and *with what arguments* — enough to render a real approval prompt instead of an opaque "something is waiting".

The interrupt reports `reason: 'approval_required'`. `@ag-ui/core` types `Interrupt.reason` as a free-form `z.string()`, and `approval_required` is what TanStack AI's interrupt manager matches on to bind Approve/Reject controls. CopilotKit reads neither field and is unaffected.

**`snapshotResponse()`** streams a stored thread as a run without calling a model:

```
RUN_STARTED -> MESSAGES_SNAPSHOT -> RUN_FINISHED
```

with `outcome.interrupts` on the finish event when an approval is still pending, so a client reopening a thread lands in the same paused state it left. `MESSAGES_SNAPSHOT` is the one hydration shape both AG-UI clients already handle — `@ag-ui/client` replaces `agent.messages` from it, and TanStack's stream processor has a case for it.

Also drops the protocol's `headers()` override, now redundant against the base-class default from #963.

### DIFF
--- a/src/AgentUserInteraction/AgentUserInteraction.php
+++ b/src/AgentUserInteraction/AgentUserInteraction.php
@@ -265,13 +265,11 @@ class AgentUserInteraction
     {
         return [
             'id' => $id,
-            // TanStack only resolves approvals reasoned this way, so it wins over AG-UI's core "tool_call"...
             'reason' => 'approval_required',
             ...(filled($reason) ? ['message' => $reason] : []),
             'toolCallId' => $id,
             'metadata' => [
                 'kind' => 'approval',
-                // A missing tool name would leave the interrupt unmatched by TanStack, and so unapprovable...
                 'toolName' => (string) $tool,
                 'input' => (object) $arguments,
             ],
@@ -306,7 +304,6 @@ class AgentUserInteraction
             'toolCallId' => $toolResult['id'],
             'content' => $content,
             ...($denied || $failed ? ['error' => $content] : []),
-            // The AG-UI tool message has an error field but no denied field, so refusal rides metadata as it does live...
             ...($denied ? ['metadata' => ['denied' => true]] : []),
         ];
     }
--- a/src/AgentUserInteraction/AgentUserInteraction.php
+++ b/src/AgentUserInteraction/AgentUserInteraction.php
@@ -265,12 +265,14 @@ class AgentUserInteraction
     {
         return [
             'id' => $id,
+            // TanStack only resolves approvals reasoned this way, so it wins over AG-UI's core "tool_call"...
             'reason' => 'approval_required',
             ...(filled($reason) ? ['message' => $reason] : []),
             'toolCallId' => $id,
             'metadata' => [
                 'kind' => 'approval',
-                ...(filled($tool) ? ['toolName' => $tool] : []),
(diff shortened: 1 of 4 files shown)


pascalbaljet:
PR #964 gives a hydrated tool result somewhere to say it was refused:

```php
'content' => $content,
...($denied ? ['error' => $content] : []),
```

The live event has no equivalent. `AgentUserInteractionProtocol` maps it as:

```php
$event instanceof ToolResult => [[
    'type' => 'TOOL_CALL_RESULT',
    'messageId' => $event->toolResult->resultId ?? $event->id,
    'toolCallId' => $event->toolResult->id,
    'content' => $this->toolResultContent($event),
    'role' => 'tool',
]],
```

and `toolResultContent()` flattens the outcome into that one string:

```php
protected function toolResultContent(ToolResult $event): string
{
    if (! $event->successful) {
        return $event->error ?? 'The tool call failed.';
    }

    return is_string($event->toolResult->result)
        ? $event->toolResult->result
        : $this->json($event->toolResult->result);
}
```

So a failure, a refusal and an answer all arrive as `content`, and the same run reads differently streamed vs reopened. The stream event has the facts to distinguish them (`successful`, `error`, `denied`), they just have nowhere to go.

Could we put the same two flags on the live frame? Both additive, both absent on success:

```php
$event instanceof ToolResult => [[
    'type' => 'TOOL_CALL_RESULT',
    'messageId' => $event->toolResult->resultId ?? $event->id,
    'toolCallId' => $event->toolResult->id,
    'content' => $this->toolResultContent($event),
    'role' => 'tool',
    ...(! $event->successful ? ['error' => $event->error ?? 'The tool call failed.'] : []),
    ...($event->denied ? ['denied' => true] : []),
]],
```

`denied` rather than folding refusal into `error`, because otherwise the two paths mean different things by the same key: on the hydrated row `error` is set *only* for denials, so a client reading `error` live would infer "refused" for an ordinary failure. `denied` is also the name already used on `Responses\Data\ToolResult` and in the persisted payload, so it stays one vocabulary. Absent both keys means success, so no existing client changes.

pascalbaljet:
`uiMessagesFrom()` hydrates from the stored row:

```php
$id = $message instanceof ConversationMessage ? $message->id : (string) Str::ulid();
```

The live frames use the stream event's id instead:

```php
'messageId' => $event->messageId,
```

Those ids come from different places, so a client that rendered the turn as it streamed shares no id with the row that comes back from `toClientState()`. It can't tell they're the same message, so both copies render.

Could we report the persisted id on `RUN_FINISHED`? `StreamableAgentResponse::$assistantMessageId` already holds it by the time the run completes, and `runFinishedPart()` spreads attributes, so it's one key:

```php
$this->runFinishedPart([
    ...$this->completionAttributes(...),
    'messageId' => $response->assistantMessageId,
]);
```

PR #969, opened 2026-08-31
Preserve structured output format when merging provider options (Anthropic)

andyglim:
### Problem

`Gateway/Anthropic/Concerns/BuildsTextRequests::buildTextRequestBody()` places the native structured-output schema under `output_config.format`:

```php
$body['output_config'] = [
    'format' => ['type' => 'json_schema', 'schema' => /* … */],
];
```

and then finishes with:

```php
return $this->applyPromptCacheBreakpoints(array_merge($body, $providerOptions), $options);
```

`array_merge` is **shallow** on string keys, so any agent whose `providerOptions()` also returns an `output_config` key replaces the entire value and silently drops `format`. For example an agent that lowers the effort level:

```php
{
    return ['output_config' => ['effort' => 'medium']];
}
```

produces a request where `output_config` is `['effort' => 'medium']` — the structured-output `format` is gone. The request is sent without its schema, the model no longer returns the structured payload, and the parsed response is missing the expected keys. **No error is raised** — the caller just gets an array without its structured keys (e.g. `Undefined array key "…"` downstream).

This makes `output_config`-based provider options (such as `effort`) mutually exclusive with structured output, even though the two settings are meant to coexist in a single `output_config` object.

### Fix

Merge `output_config` one level deeper so provider-option keys coexist with the builder-supplied `format`, while every other provider option keeps its existing shallow-merge (override) behaviour:

```php
$merged = array_merge($body, $providerOptions);

if (isset($body['output_config'], $providerOptions['output_config'])
    && is_array($body['output_config'])
    && is_array($providerOptions['output_config'])) {
    $merged['output_config'] = array_merge(
        $body['output_config'],
        $providerOptions['output_config'],
    );
}
```

Leaf keys in `providerOptions` still win (so an explicitly provided `format` would still override), but sibling keys like `format` are no longer discarded.

### Tests

Adds a regression test (`tests/Feature/Providers/Anthropic/ProviderOptionsTest.php`) and a fixture agent (`StructuredWithOutputConfigOptionsAgent`) that combines structured output with an `output_config` provider option, asserting the sent request body carries **both** `output_config.format` (`json_schema`) and `output_config.effort` (`medium`).

> Note: I was unable to run the full Pest suite locally (dependency install kept timing out on network in my environment), so I'm relying on CI to run the added test. The merge behaviour itself was verified in isolation. Happy to adjust the approach — e.g. a more general recursive merge — if you'd prefer.

### DIFF
--- a/src/Gateway/Anthropic/Concerns/BuildsTextRequests.php
+++ b/src/Gateway/Anthropic/Concerns/BuildsTextRequests.php
@@ -70,7 +70,38 @@ trait BuildsTextRequests
             'top_p' => $options?->topP,
         ]));
 
-        return $this->applyPromptCacheBreakpoints(array_merge($body, $providerOptions), $options);
+        return $this->applyPromptCacheBreakpoints(
+            $this->mergeProviderOptions($body, $providerOptions),
+            $options,
+        );
+    }
+
+    /**
+     * Merge provider options into the request body.
+     *
+     * Provider options are shallow-merged, except for the "output_config" key,
+     * which is merged one level deeper so that options sharing it (e.g. an
+     * "effort" level) do not discard the structured-output "format" the request
+     * builder places there.
+     *
+     * @param  array<string, mixed>  $body
+     * @param  array<string, mixed>  $providerOptions
+     * @return array<string, mixed>
+     */
+    protected function mergeProviderOptions(array $body, array $providerOptions): array
+    {
+        $merged = array_merge($body, $providerOptions);
+
+        if (isset($body['output_config'], $providerOptions['output_config'])
+            && is_array($body['output_config'])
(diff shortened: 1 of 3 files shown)


pushpak1300:
Thanks for your pull request to Laravel!

I appreciate you taking the time to submit this; however, it appears this contribution may have been primarily AI-generated without careful human review and consideration.

We’ve found that AI-generated code often doesn’t align well with Laravel’s conventions, architectural decisions, and the specific context of what we’re trying to accomplish with our framework and packages. Quality contributions require thoughtful human insight into the codebase.

If you’re interested in contributing to Laravel, I’d encourage you to familiarize yourself with the existing codebase, engage with the community, and submit PRs that reflect your own understanding and careful consideration of the problem you’re solving.

PR #970, opened 2026-09-01
Support hosted code execution as a provider tool

pushpak1300:
Anthropic, OpenAI, Gemini and xAI all host a sandboxed code interpreter that runs Python server side, but the SDK had no way to reach it. The only option was `ProviderTool` with a hand written payload per provider.

This adds `CodeExecution` alongside the existing provider tools:

```php
use Laravel\Ai\Providers\Tools\CodeExecution;

// The model writes and runs Python in the provider's sandbox.
agent(tools: [new CodeExecution])->prompt('What is the SHA-256 of "laravel"?');
```

Each provider gets the wire format its API expects, so the same tool works everywhere:

| Provider | Sent as |
|---|---|
| Anthropic | `{"type": "code_execution_20250825", "name": "code_execution"}` |
| OpenAI, Azure | `{"type": "code_interpreter", "container": {"type": "auto"}}` |
| Gemini | `{"code_execution": {}}` |
| xAI | `{"type": "code_interpreter"}` |

Provider options work the way they do on `WebSearch`, and on OpenAI and Azure they take precedence over the default container, which is how you attach files:

```php
(new CodeExecution)->withProviderOptions([
    'container' => ['type' => 'auto', 'file_ids' => ['file_123']],
]);
```

### DIFF
--- a/src/Contracts/Providers/SupportsCodeExecution.php
+++ b/src/Contracts/Providers/SupportsCodeExecution.php
@@ -1,0 +1,13 @@
+<?php
+
+namespace Laravel\Ai\Contracts\Providers;
+
+use Laravel\Ai\Providers\Tools\CodeExecution;
+
+interface SupportsCodeExecution
+{
+    /**
+     * Get the code execution tool options for the provider.
+     */
+    public function codeExecutionToolOptions(CodeExecution $codeExecution): array;
+}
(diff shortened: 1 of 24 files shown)


goatrenterguy:
You are crushing it with these new additions. I seem to be in a cycle of "I need it to do XXXX" I check open/closed prs and you guys are actively working on it! I might need to take a week or two off so its "shipped" rather than "in-progress" but love to see the progress

PR #974, opened 2026-09-01
Disable tools on the final OpenAI-compatible step

### LINKED ISSUE #973: OpenAI-compatible final step can request a tool after the step budget is exhausted
### Description

`TextGenerationLoop` marks the last configured request as `isFinalStep`. The OpenAI-compatible gateway still sends the configured tools on that request with `tool_choice: auto`.

The model can therefore request another tool even though the loop has no remaining step in which to run it. The loop refuses the call:

> The agent reached its maximum number of steps without running this tool call.

No request remains for the model to produce a response from the results already collected. If the provider response contains only the tool call, the agent response has no text.

I reproduced this with `laravel/ai` v0.11.0 and an OpenAI-compatible provider. The same path is present on the current `0.x` branch. Increasing `maxSteps` from 5 to 10 only moved the boundary.

### Expected behavior

The final OpenAI-compatible request should keep the tool definitions required by its message history, but set `tool_choice` to `none`. This lets the model produce a response from the results already collected without requesting a tool that cannot run.

This keeps `maxSteps` as the total number of model requests. For example, `maxSteps: 5` still allows at most five requests: up to four tool-capable requests followed by one response request. With `maxSteps: 1`, tools cannot be called.

### Related

- #639 reports a related empty-response and unbalanced-history symptom against the previous loop implementation.
- #699 and #702 proposed allowing `maxSteps` tool executions followed by another response request.

This proposal differs from #699 and #702 because it performs no additional request.

[gpt-5.6-sol(xhigh) agent, claude-opus-5(xhigh) reviewer, AnnoyingTechnology(human) reviewer]

AnnoyingTechnology:
## Summary

- set `tool_choice` to `none` on the final OpenAI-compatible generation step
- keep tool definitions so previous tool calls and results remain valid
- preserve `maxSteps` as the total model-request limit
- cover synchronous and streaming generation

When tools remain callable on the final request, a model can request one even though `TextGenerationLoop` will not execute it. The loop then ends without a final text response.

The final request now requires a response while retaining tool definitions and their prompt-cache prefix. With `maxSteps: 1`, tools cannot be called; agents that need one must configure at least two steps. The upgrade guide documents this behavior.

Fixes #973

## Tests

- `composer test` — 2318 passed, 260 skipped, 4915 assertions; 2 existing notices
- `composer test:lint` — passed
- `composer test:types` — passed

[gpt-5.6-sol(xhigh) agent, claude-opus-5(xhigh) reviewer, AnnoyingTechnology(human) reviewer]

### DIFF
--- a/UPGRADE.md
+++ b/UPGRADE.md
@@ -1,5 +1,15 @@
 # Upgrade Guide
 
+## Upgrading To 0.12 From 0.11
+
+### Final OpenAI-Compatible Step Disables Tools
+
+**Likelihood Of Impact: Low**
+
+OpenAI-compatible text generation now sets `tool_choice` to `none` on the final configured model request. Tool definitions remain in the request so previous tool calls and results stay valid, but the model must produce a response instead of requesting a tool that the loop cannot run.
+
+`maxSteps` still sets the total number of model requests. For example, `maxSteps: 5` allows up to four tool-capable requests followed by one response request. With `maxSteps: 1`, tools cannot be called; configure at least two steps when an agent needs to call one.
+
 ## Upgrading To 0.11 From 0.10
 
 ### Connection Failures Throw `ProviderConnectionException`
--- a/src/Gateway/OpenAiCompatible/Concerns/BuildsTextRequests.php
+++ b/src/Gateway/OpenAiCompatible/Concerns/BuildsTextRequests.php
@@ -56,7 +56,11 @@ trait BuildsTextRequests
         $providerOptions = $options?->providerOptions($provider->name());
 
         if (filled($providerOptions)) {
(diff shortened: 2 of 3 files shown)


AnnoyingTechnology:
@pushpak1300 If you have a moment to take a look.

pushpak1300:
I don't think this is the correct approach to solve it. I'll look into more detail i feel providing middleware on per step level is better than doing this.

AnnoyingTechnology:
> I don't think this is the correct approach to solve it. I'll look into more detail i feel providing middleware on per step level is better than doing this.

Right, I've just seen https://github.com/laravel/ai/pull/761 makes sense.

PR #978, opened 2026-09-02
[1.x] Report why a resume was rejected on RUN_ERROR

pascalbaljet:
Resuming a paused run with a decision the run loop rejects throws `ApprovalMismatchException`. On a normal request that renders as a 409 carrying the message and the approvals still pending. On a streamed run it reaches the protocol's error handling and comes back as `RUN_ERROR` with "An error occurred.", which is also what a provider timeout looks like. A client that could recover, by hydrating the thread through `snapshotResponse()` in #965 and landing back on the paused turn, has no way to know that's the right move.

`RunError` takes an optional code, so the mismatch can name itself:

```json
{"type":"RUN_ERROR","message":"Approval decisions do not match the pending tool calls.","code":"approval_mismatch"}
```

The message is the one the 409 already returns, so nothing is exposed here that a non-streaming client couldn't already see. `maskedErrorParts()` receives the exception to tell the cases apart. A mismatch that still has approvals pending is a race between clients rather than a fault, so it's no longer reported, while the variants thrown with nothing left to approve stay reported.

Targeting 1.x only, since that parameter changes an abstract method signature and would break any custom `StreamProtocol` implementation.

### DIFF
--- a/src/Exceptions/ApprovalMismatchException.php
+++ b/src/Exceptions/ApprovalMismatchException.php
@@ -20,14 +20,6 @@ class ApprovalMismatchException extends AiException
     }
 
     /**
-     * Determine if the run is still paused on approvals that may be decided again.
-     */
-    public function isRecoverable(): bool
-    {
-        return $this->pendingApprovals->isNotEmpty();
-    }
-
-    /**
      * Render the exception as an HTTP response.
      */
     public function render(Request $request): Response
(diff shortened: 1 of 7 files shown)


pascalbaljet:
Rewrote this after checking the AG-UI interrupts contract. The original version finished the run with an interrupt outcome, which is against the spec: It requires `RunError` for a resume that can't be correlated or doesn't cover every open interrupt. This keeps `RunError` and adds a `code` instead.

PR #982, opened 2026-09-04
fix: preserve ResourceLink appResourceUri in McpServerTool output

### LINKED ISSUE #933: MCP Apps: preserve ResourceLink (appResourceUri) in McpServerTool output
# MCP Apps: preserve ResourceLink (appResourceUri) in McpServerTool output

## Problem

`McpServerTool::finalResponse()` extracts only text from MCP responses. When an MCP tool returns a `ResourceLink` content alongside text (e.g. a tool annotated with `#[RendersApp]`), the app URI is discarded. The AI agent and the frontend never receive the app resource information needed to render interactive HTML iframes (MCP Apps).

## Current behavior

```
MCP tool returns: [Text("dashboard loaded."), ResourceLink("ui://resources/weather-dashboard-app")]
                              ↓
McpServerTool::finalResponse() → "dashboard loaded."  // ResourceLink discarded
```

`src/Tools/McpServerTool.php` (current `finalResponse`):

```php
protected function finalResponse(array $responses): string
{
    $final = collect($responses)->last(fn (object $response): bool => ! $response->isNotification());

    if ($final === null) {
        return '';
    }

    $text = (string) $final->content();

    return $final->isError()
        ? $this->errorMessage($text)
        : $text;
}
```

## Proposed change

Detect `ResourceLink` content (from `Laravel\Mcp\Server\Content\ResourceLink`) in MCP responses and include the URI in the tool output:

```
McpServerTool::finalResponse() → {"text":"dashboard loaded.","appResourceUri":"ui://resources/weather-dashboard-app"}
```

## File to modify

`src/Tools/McpServerTool.php` — method `finalResponse()`

## Implementation

```php
protected function finalResponse(array $responses): string
{
    $final = collect($responses)->last(fn (object $response): bool => ! $response->isNotification());

    if ($final === null) {
        return '';
    }

    $text = (string) $final->content();
    $appUri = null;

    foreach ($responses as $response) {
        if ($response->isNotification()) {
            continue;
        }

        $content = $response->content();

        if ($content instanceof ResourceLink && str_starts_with((string) $content, 'ui://')) {
            $appUri = (string) $content;
        }
    }

    if ($appUri !== null) {
        return json_encode([
            'text' => $text,
            'appResourceUri' => $appUri,
        ], JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES);
    }

    return $final->isError()
        ? $this->errorMessage($text)
        : $text;
}
```

Add import:

```php
use Laravel\Mcp\Server\Content\ResourceLink;
```

## Why JSON output

- MCP Apps are a host-level feature — the host (IDE, chat UI) must detect the URI and render an iframe
- JSON is parseable by any frontend without custom parsing
- Backward compatible: if no `ResourceLink`, returns plain text as before

## Frontend contract

When tool output is JSON with `appResourceUri`:

1. Fetch HTML via `resources/read` JSON-RPC to the MCP endpoint
2. Render in iframe via `srcdoc`
3. Implement MCP host via `postMessage` to proxy `tools/call` back to the MCP server

## Upstream alternative

Instead of JSON in text output, could emit a separate streaming event (e.g. `app_resource`) — but that requires AI SDK protocol changes too.

skie:
Fixes #933.

## Problem
`McpServerTool::finalResponse()` extracts only text from MCP responses. When an MCP tool returns a `ResourceLink` alongside text (e.g. a tool annotated with `#[RendersApp]`), the app URI is discarded. The agent and frontend never receive the app resource info needed to render interactive HTML iframes (MCP Apps).

## Change
- Detect `ResourceLink` content with `ui://` URI in MCP responses and include it as `appResourceUri` in a JSON payload: `{"text":"...","appResourceUri":"ui://..."}`
- Backward compatible: no `ResourceLink` returns plain text as before
- Ignores notifications, order-independent, preserves MCP tool error prefix

## Frontend contract
When tool output is JSON with `appResourceUri`:
1. Fetch HTML via `resources/read` JSON-RPC to the MCP endpoint
2. Render in iframe via `srcdoc`
3. Implement MCP host via `postMessage` to proxy `tools/call` back to the MCP server

## Tests
- it includes app resource uri when tool returns text and ui resource link
- it resolves app resource uri regardless of content order
- it ignores notifications when resolving app resource uri
- it ignores non-ui resource links for app rendering
- it preserves error prefix when app resource is present

All 14 tests in `McpServerToolTest` pass.

### DIFF
--- a/src/Tools/McpServerTool.php
+++ b/src/Tools/McpServerTool.php
@@ -10,6 +10,7 @@ use Laravel\Ai\Contracts\Tool;
 use Laravel\Ai\Tools\Concerns\NormalizesMcpResult;
 use Laravel\Mcp\Response;
 use Laravel\Mcp\ResponseFactory;
+use Laravel\Mcp\Server\Content\ResourceLink;
 
 class McpServerTool implements Tool
 {
@@ -136,22 +137,60 @@ class McpServerTool implements Tool
     }
 
     /**
-     * Reduce a list of responses to the last non-notification response's text.
+     * Reduce a list of responses to tool output, preserving MCP App resource links.
+     *
+     * When an MCP tool returns a `ResourceLink` with a `ui://` URI alongside
+     * text, the URI is included as `appResourceUri` in a JSON payload so the
+     * host can fetch the HTML via `resources/read` and render it in an iframe.
+     * Without a `ResourceLink`, returns plain text as before.
      *
      * @param  array<int, object>  $responses
      */
     protected function finalResponse(array $responses): string
     {
-        $final = collect($responses)->last(fn (object $response): bool => ! $response->isNotification());
+        $text = '';
+        $appUri = null;
+        $isError = false;
 
-        if ($final === null) {
+        foreach (array_reverse($responses) as $response) {
+            if ($response->isNotification()) {
+                continue;
+            }
+
+            $content = $response->content();
+
+            if ($content instanceof ResourceLink && str_starts_with((string) $content, 'ui://')) {
+                $appUri ??= (string) $content;
+                $isError = $isError || $response->isError();
+
+                continue;
+            }
+
+            if ($text === '') {
+                $text = (string) $content;
+                $isError = $response->isError();
+            }
+
+            if ($text !== '' && $appUri !== null) {
+                break;
+            }
+        }
+
+        if ($text === '' && $appUri === null) {
             return '';
         }
 
-        $text = (string) $final->content();
+        if ($isError) {
+            $text = $this->errorMessage($text);
+        }
 
-        return $final->isError()
-            ? $this->errorMessage($text)
-            : $text;
+        if ($appUri !== null) {
+            return json_encode([
+                'text' => $text,
+                'appResourceUri' => $appUri,
+            ], JSON_UNESCAPED_UNICODE | JSON_UNESCAPED_SLASHES) ?: '';
+        }
+
+        return $text;
     }
 }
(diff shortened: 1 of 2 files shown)


pushpak1300:
is this tested by you ?

skie:
Yes, I built demo app that render mcp dashboard with fake sql queires problems, and agent call tools based on button pressed inside MCP app, and asking approval then create index.

pushpak1300:
@codex review

pushpak1300:
Thanks for the detailed writeup, it made this easy to trace.

Closing this one, because the URI does not come from the tool result.

### Where the URI actually comes from

It comes from the **tool definition**, not the response. `#[RendersApp]` puts `_meta.ui.resourceUri` into `Tool::toArray()`:

- laravel/mcp `src/Server/Tool.php:84`
- [MCP Apps spec](https://modelcontextprotocol.io/docs/extensions/apps): *"The tool description includes a `_meta.ui.resourceUri` field... The host can preload this resource before the tool is even called."*

A `#[RendersApp]` tool just returns `Response::text(...)`. There is no `ResourceLink` in the result, so `finalResponse()` is not losing anything.

### Why not put it in the string anyway

That returned string is what the **model** reads. Folding render info into it pushes host metadata into the prompt. It belongs out of band, on the tool call.

### Where the real gap is

Not in `McpServerTool`. It is on the client path, when you connect to an outside MCP server:

1. `McpTool` drops the `_meta` that `Client\Primitives\Tool` already exposes.
2. laravel/mcp's client sends `capabilities: {}` on initialize (`src/Client/Methods/Initialize.php:34`), so a strict server never advertises the UI extension at all.

I'll open issues for both. The laravel/ai one is smaller if you'd like to pick it up. 🙏

PR #984, opened 2026-09-06
Allow creating a conversation without prompting

### LINKED ISSUE #983: Create a conversation before the first prompt
# Create a conversation before the first prompt

## Problem

The only supported way to create a conversation is inside a prompt:

```php
$response = (new SalesCoach)->forUser($user)->prompt('Hello!');

$conversationId = $response->conversationId;
```

Chat UIs need an empty conversation before the first message: a row to list, a title to rename, a place to navigate to. Today they either reach into `DatabaseConversationStore::storeConversation()` or insert the row by hand with `$user->conversations()->create([...])`. Both depend on internals such as the unguarded model, the string primary key, and the morph participant, which can change without notice.

## Proposed API

`RemembersConversations` is the trait that opts an agent into the package's conversation persistence, so conversation-creation belongs there, next to `forUser()`, `continue()`, and `continueLastConversation()`:

```php
$agent = (new SalesCoach)->forUser($user)->startConversation(title: 'New chat');

$conversationId = $agent->currentConversation();
```

`startConversation()` stores a conversation for the current participant and returns `static`, so the conversation ID is read with `currentConversation()`, like `continueLastConversation()`. `RememberConversation::handle()` already reuses an existing conversation when the agent has one, so the first prompt after creating continues this conversation instead of creating another.

## Behavior

- Same columns as prompting: `participant_type`, `participant_id`, `title`, timestamps.
- No messages are created.
- A later `continue($conversationId)` continues this conversation instead of creating another.

SantosVilanculos:
# Create a conversation before the first prompt

Adds `startConversation()` to the `RemembersConversations` trait so chat UIs can create an empty conversation before the first prompt.

```php
$agent = (new SalesCoach)->forUser($user)->startConversation(title: 'New chat');

$conversationId = $agent->currentConversation();
```

The method stores a conversation for the current participant through the `ConversationStore` and returns `static`, so the ID is read with `currentConversation()`, like `continueLastConversation()`. The first prompt after creating continues the same conversation, because `RememberConversation::handle()` only stores a conversation when the agent has none.

Closes #983

### DIFF
--- a/src/Concerns/RemembersConversations.php
+++ b/src/Concerns/RemembersConversations.php
@@ -31,6 +31,24 @@ trait RemembersConversations
     }
 
     /**
+     * Create a new conversation for the current participant without prompting.
+     */
+    public function startConversation(string $title = 'New conversation'): static
+    {
+        $participant = $this->conversationUser;
+        $participantType = $participant === null ? null : Conversation::participantType($participant);
+        $participantId = $participant === null ? null : Conversation::participantKey($participant);
+
+        $this->conversationId = resolve(ConversationStore::class)->storeConversation(
+            $participantType,
+            $participantId,
+            $title,
+        );
+
+        return $this;
+    }
+
+    /**
      * Continue an existing conversation, optionally as the given user.
      */
     public function continue(string $conversationId, ?object $as = null): static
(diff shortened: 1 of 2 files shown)


pushpak1300:
Thanks for your pull request!

PR #986, opened 2026-09-06
Allow custom columns on conversations

### LINKED ISSUE #985: Set custom columns on conversations
Set custom columns on conversations
---
# Set custom columns on conversations

## Problem

`DatabaseConversationStore::storeConversation()` writes only `id`, `participant_type`, `participant_id`, `title`, and timestamps. Any extra column on the `agent_conversations` table stays null when the package creates a conversation.

A common case: a conversation belongs to a user inside an organization. The `participant` relation is a single `morphTo`, so it points at one model, the user or the organization, not both. Apps that need both scopes add a column such as `organization_id`.

## Current workaround

Pre-create the conversation through the app's own code and set the column by hand:

```php
$conversation = $request->user()->conversations()->create([
    'id' => Str::uuid7(),
    'organization_id' => $organization->id,
    'title' => 'New chat',
]);
```

### Tradeoffs of the workaround

- Covers only the pre-created path. A conversation created by the first `prompt()` alone goes through `storeConversation()`, which writes no `organization_id`, so that row stays null and the scope is lost.
- Depends on internals: the empty `$guarded` array, the hand-supplied string primary key, and the morph participant. The line stores today; any of those change and the app breaks.
- The app must route every creation through its own controller to keep scoping consistent. Any prompting code that skips pre-creating violates the scope silently, because the package does not enforce it.

## Proposed API

Add a fluent setter and a plain getter to `RemembersConversations`, following the package's existing `withMiddleware()`/`middleware()` pattern:

```php
public function withConversationAttributes(array $attributes): static
{
    $this->conversationAttributes = array_merge($this->conversationAttributes, $attributes);

    return $this;
}

public function conversationAttributes(): array
{
    return $this->conversationAttributes;
}
```

The getter name matches the trait's other noun getters, such as `conversationParticipant()`.

Attributes apply to both creation paths, because they are read where the conversation is stored:

```php
// Pre-creating without a prompt...
$agent = (new SalesCoach)
    ->forUser($user)
    ->withConversationAttributes(['organization_id' => $organization->id])
    ->startConversation(title: 'New chat');

// Or the first prompt...
(new SalesCoach)
    ->forUser($user)
    ->withConversationAttributes(['organization_id' => $organization->id])
    ->prompt('Hello');
```

## Store changes

`storeConversation()` gains an `$attributes` parameter on the interface and every implementation:

```php
public function storeConversation(?string $participantType, string|int|null $participantId, string $title, array $attributes = []): string
```

`DatabaseConversationStore` merges `$attributes` into its insert, so custom columns land on the row. `RememberConversation::handle()` and `startConversation()` pass the agent's attributes through.

Changing the interface signature breaks custom `ConversationStore` implementations on 0.x; the in-repo fixtures update with it.

## Decisions

- Unknown attribute keys fail the insert with a `QueryException`. That is intentional: the app owns the column and adds it to its own table.
- A model-event-based approach is out of scope; the calling code already knows its columns, so `withConversationAttributes()` covers the need without changing how messages are stored.

SantosVilanculos:
# Allow custom columns on conversations

## What

`RemembersConversations` gains `withConversationAttributes()` to set extra columns before a conversation is created, and `conversationAttributes()` to read them. `storeConversation()` on the `ConversationStore` contract and every implementation takes an `array $attributes` (defaulting to `[]`) and `DatabaseConversationStore` merges them into the insert.

```php
(new SalesCoach)
    ->forUser($user)
    ->withConversationAttributes(['organization_id' => $organization->id])
    ->prompt('Hello');

// or pre-creating without a prompt...
$agent = (new SalesCoach)
    ->forUser($user)
    ->withConversationAttributes(['organization_id' => $organization->id])
    ->startConversation(title: 'New chat');
```

This lets apps scope conversations to an organization (or any custom column) on both creation paths, instead of hand-inserting rows through the unguarded model.

## Tests

- Trait setter merges and the getter returns accumulated attributes.
- Attributes reach the store when the first `prompt()` creates the conversation (via `InMemoryConversationStore`).
- `DatabaseConversationStore` persists the attributes on the conversation row.

The interface signature change is reflected across the in-repo fixtures and anonymous test stores.

Closes #985

### DIFF
--- a/src/Concerns/RemembersConversations.php
+++ b/src/Concerns/RemembersConversations.php
@@ -12,6 +12,11 @@ trait RemembersConversations
     protected ?object $conversationUser = null;
 
     /**
+     * @var array<string, mixed>
+     */
+    protected array $conversationAttributes = [];
+
+    /**
      * Start a new conversation for the given participant.
      */
     public function forParticipant(object $participant): static
@@ -52,6 +57,28 @@ trait RemembersConversations
             ->latestConversationId(Conversation::participantType($as), Conversation::participantKey($as));
 
         return $this;
+    }
+
+    /**
+     * Set the extra columns to persist when a new conversation is created.
+     *
+     * @param  array<string, mixed>  $attributes
+     */
+    public function withConversationAttributes(array $attributes): static
+    {
+        $this->conversationAttributes = array_merge($this->conversationAttributes, $attributes);
+
+        return $this;
+    }
+
+    /**
+     * Get the extra columns to persist when a new conversation is created.
+     *
+     * @return array<string, mixed>
+     */
+    public function conversationAttributes(): array
+    {
+        return $this->conversationAttributes;
     }
 
     /**
--- a/src/Contracts/ConversationStore.php
+++ b/src/Contracts/ConversationStore.php
@@ -18,8 +18,10 @@ interface ConversationStore
 
     /**
      * Store a new conversation and return its ID.
+     *
+     * @param  array<string, mixed>  $attributes  Additional conversation columns to persist.
      */
-    public function storeConversation(?string $participantType, string|int|null $participantId, string $title): string;
+    public function storeConversation(?string $participantType, string|int|null $participantId, string $title, array $attributes = []): string;
 
     /**
      * Store a new user message for the given conversation and return its ID.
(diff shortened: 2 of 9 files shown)


pushpak1300:
Thanks for your pull request!

PR #1004, opened 2026-09-16
Add SummarizeContext middleware for long conversations

pushpak1300:
This adds a `SummarizeContext` agent middleware. When a conversation grows past a number of turns, the older turns are replaced with a short summary before the request goes to the model. The latest turns are always sent as they are.

```php
use Laravel\Ai\Middleware\SummarizeContext;

public function middleware(): array
{
    return [new SummarizeContext(turns: 30)];
}
```

`turns` is how many of the most recent user turns stay in full. 

Summaries are not stored in the database. They live in the cache, so nothing is added to the conversation tables and there is no migration. Each new summary builds on the previous one, so only the turns that just fell out of the window are summarized.

> [!NOTE]
> The cache store comes from the new `ai.caching.summaries.store` config key, which defaults to `CACHE_STORE`. Add it to your published `config/ai.php` if you want a different store.


taylorotwell:
I think I want to leave this for user-land for now.

PR #1007, opened 2026-09-16
Store conversation turns as steps

### LINKED ISSUE #402: Multi-round tool call loops lose round boundaries when stored in DatabaseConversationStore

### LINKED ISSUE #809: Approval resume fails with a 400 when a multi-step turn paused on a later step
When an agent turn runs several tool-calling steps and then pauses for approval on a later step (for example, two read tools execute automatically across earlier steps, then a gated tool returns `Approval::required()`), resuming the approval fails with a 400 from Anthropic:

```
messages.N.content.0: unexpected tool_use_id found in tool_result blocks: toolu_xxx.
Each tool_result block must have a corresponding tool_use block in the previous message.
```

Single-step pauses resume fine. The failure only happens when the turn already produced tool results in earlier steps before pausing.

I hit this on Anthropic, but the Gemini, DeepSeek and Bedrock mappers replay `providerContentBlocks` verbatim in the same way, so I'd expect the same mismatch there. Both transports are affected since the streamed and queued paths rehydrate through the same store.

- laravel/ai: v0.10.1
- laravel/framework: v13
- PHP: 8.4
- Provider: Anthropic (claude-sonnet-5), conversation storage enabled

## Root cause

A paused multi-step turn is persisted as one assistant row holding the whole turn's `tool_calls` and `tool_results`, but `meta.provider_content_blocks` only contains the final step's blocks. `pausedProviderContentBlocks()` reads the last assistant message of the turn, so earlier steps' `tool_use` blocks are not in it.

On resume, `DatabaseConversationStore::reconstructToolTurn()` takes the pause branch: it replays the assistant message as those raw blocks, then emits every already-resolved result of the turn in a `ToolResultMessage` directly after. The Anthropic mapper replays raw blocks verbatim in place of the message's `toolCalls` (`MapsMessages::mapAssistantMessage()`), so the earlier steps' results reference `tool_use` ids the provider was never shown, and the request is rejected.

The raw blocks can't simply be dropped to fall back to the generic path, because they carry the signed `thinking` block that Anthropic requires back unmodified during tool use.

## Reproduction

Deterministic, no API call needed. The seeded row below mirrors a row captured from my production database after a real pause (structure for structure, with a placeholder signature since nothing validates it at the store layer), and it's the shape `storeAssistantMessage()` and `pausedProviderContentBlocks()` write by construction: two calls, one resolved, one pending, raw blocks containing only the pending call. Seeding rows directly is the same approach `DatabaseConversationStoreTest` already uses.

```php
use Illuminate\Support\Facades\DB;
use Illuminate\Support\Str;
use Laravel\Ai\Messages\AssistantMessage;
use Laravel\Ai\Messages\ToolResultMessage;
use Laravel\Ai\Storage\DatabaseConversationStore;

it('replays earlier-step results with no matching tool_use in the raw blocks', function () {
    $conversationId = (string) Str::uuid7();

    DB::table('agent_conversations')->insert([
        'id' => $conversationId, 'title' => '', 'created_at' => now(), 'updated_at' => now(),
    ]);

    DB::table('agent_conversation_messages')->insert([
        [
            'id' => (string) Str::uuid7(), 'conversation_id' => $conversationId,
            'agent' => 'demo', 'role' => 'user',
            'content' => 'Research my data, then run the gated tool.',
            'attachments' => '[]', 'tool_calls' => '[]', 'tool_results' => '[]',
            'usage' => '[]', 'meta' => '[]', 'approval_state' => null,
            'created_at' => now(), 'updated_at' => now(),
        ],
        [
            'id' => (string) Str::uuid7(), 'conversation_id' => $conversationId,
            'agent' => 'demo', 'role' => 'assistant',
            'content' => 'Checked the data, now running the gated tool.',
            'attachments' => '[]',
            'tool_calls' => json_encode([
                ['id' => 'toolu_1', 'name' => 'search_data', 'arguments' => []],
                ['id' => 'toolu_2', 'name' => 'gated_tool', 'arguments' => []],
            ]),
            'tool_results' => json_encode([
                ['id' => 'toolu_1', 'name' => 'search_data', 'arguments' => [], 'result' => 'rows'],
            ]),
            'usage' => '[]',
            'meta' => json_encode([
                'provider' => 'anthropic',
                'provider_content_blocks' => [
                    ['type' => 'thinking', 'thinking' => '', 'signature' => 'sig'],
                    ['type' => 'text', 'text' => 'Checked the data, now running the gated tool.'],
                    ['type' => 'tool_use', 'id' => 'toolu_2', 'name' => 'gated_tool', 'input' => []],
                ],
            ]),
            'approval_state' => json_encode(['pending' => ['toolu_2' => 'Needs approval']]),
            'created_at' => now()->addSecond(), 'updated_at' => now()->addSecond(),
        ],
    ]);

    $messages = (new DatabaseConversationStore)->getLatestConversationMessages($conversationId, 50)->values();

    $pauseIndex = $messages->search(
        fn ($message) => $message instanceof AssistantMessage && filled($message->providerContentBlocks)
    );

    $next = $messages[$pauseIndex + 1];

    // The raw blocks only contain tool_use toolu_2, yet the very next message
    // carries toolu_1's result. This is the shape the provider rejects on resume.
    expect($next)->toBeInstanceOf(ToolResultMessage::class)
        ->and(collect($messages[$pauseIndex]->providerContentBlocks)->where('type', 'tool_use')->pluck('id')->all())->toBe(['toolu_2'])
        ->and($next->toolResults->pluck('id')->all())->toBe(['toolu_1']);
});
```

The test passes on v0.10.1, meaning the broken shape is produced. On resume, `appendApprovalResults()` merges the newly approved result into that trailing `ToolResultMessage`, so the request sent to Anthropic is:

```
assistant: [thinking, text, tool_use toolu_2]         <- raw blocks, toolu_1 missing
user:      [tool_result toolu_1, tool_result toolu_2] <- toolu_1 rejected
```

## Expected behaviour

The earlier steps should replay ahead of the raw-blocks message, the way the non-blocks path in `reconstructToolTurn()` already handles it:

```
assistant: [tool_use toolu_1]
user:      [tool_result toolu_1]
assistant: [thinking, text, tool_use toolu_2]  <- raw blocks, signature intact
user:      [tool_result toolu_2]               <- appended by the resume
```

## Suggested fix

Partition the turn's resolved results by whether their id appears as a `tool_use` in the raw blocks. Earlier-step results get a synthetic call/result pair in front of the raw-blocks message; same-step results (an auto-run call sharing the paused step) stay behind it, where the resume merges the approved result.

I'd suggest doing this at rehydration rather than changing what gets persisted: fixing it read-side heals every already-stored paused conversation with no migration, and because the fix lands in the shared store it covers all the raw-block providers at once rather than just Anthropic. I'm running exactly this as a store override in production and it resolves the resume. Happy to submit it as a PR with the test above if useful.

Possibly related, though the mechanism differs (those are orphaned calls, this is orphaned results): #639, #758.

pushpak1300:
The 1.x release is close. The conversation schema is the hardest part of the package to change so this PR tries to settles it now.

Today `agent_conversation_messages` stores one row per assistant turn. All tool calls and tool results of the turn are flattened into `tool_calls` and `tool_results`, so the step boundaries the provider needs are lost on write and guessed back on read.

**Before and after**

```mermaid
flowchart LR
    subgraph before ["Today: one flat row"]
        R["content: 'Read a, now deleting b'<br/>tool_calls: [toolu_1, toolu_2, toolu_3]<br/>tool_results: [toolu_1]<br/>meta.provider_content_blocks: [...]"]
    end
    R -->|"reader guesses:<br/>which call was run?<br/>which step made it?<br/>whose blocks are these?"| G["Assistant [toolu_1, toolu_2, toolu_3]<br/>ToolResult [toolu_1]<br/>toolu_2 never ran<br/>toolu_3 paused, same message"]
```

```mermaid
flowchart LR
    subgraph after ["After: one row, one entry per step"]
        S1["step 1<br/>calls: [toolu_1 + its result]<br/>reasoning: ''<br/>replay_blocks: [...]"]
        S2["step 2<br/>calls: [toolu_2, toolu_3]<br/>reasoning: '...'<br/>replay_blocks: [...]"]
    end
    S1 -->|"replayed as is"| M1["Assistant [toolu_1] + blocks<br/>ToolResult [toolu_1]"]
    S2 -->|"toolu_3 pending, kept<br/>toolu_2 unanswered, dropped"| M2["Assistant [toolu_3] + blocks"]
```

This change replaces `tool_calls` and `tool_results` with a `steps` column. Each entry is one model round-trip: its tool calls, each carrying the result that answered it, the reasoning the model emitted, and the raw provider blocks needed to replay it.

```json
[
  {"tool_calls": [{"id": "toolu_1", "name": "read_file", "arguments": {"path": "a"}, "result": "..."}], "reasoning": "", "replay_blocks": [{"type": "tool_use", "id": "toolu_1"}]},
  {"tool_calls": [{"id": "toolu_2", "name": "delete_file", "arguments": {"path": "b"}}], "reasoning": "Read a, deleting b next.", "replay_blocks": [{"type": "text", "text": "Read a, now deleting b"}, {"type": "tool_use", "id": "toolu_2"}]}
]
```

A step carries no text of its own. The turn's `content` column stays the single home for assistant text, and the reader gives it to the turn's last step, so a step's text is not duplicated between the two. The cost is that text a model emits *alongside* a tool call, which Anthropic does routinely, is not replayed where it was written: it does not vanish, it moves. Text the model wrote at step 1 comes back attached to the last step, so on reload it reads as though it followed step 1's tool result rather than preceding it. A `content` key can be added to the step shape later without a migration if that turns out to matter; it is left out for now rather than shipped as a field only one provider fills.

A paused turn is no longer a special case. It is just a turn whose last step has a tool call with no result yet. When the user approves or rejects, the result is written into that same step.

Every step keeps the raw blocks the provider returned, not only the paused one. They are replayed only for the turn currently awaiting a decision, back to the user message that started it, so a turn that pauses more than once sends every one of its steps back exactly as the provider produced them. Once a turn completes it replays through the generic mapping. Not because the provider would reject the blocks — that was the original reasoning here and it does not hold. [Anthropic](https://platform.claude.com/docs/en/build-with-claude/thinking-tool-workflows) requires a turn's thinking back only while that turn is still open, echoed verbatim beside the `tool_result` answering its `tool_use`; across a finished turn, Opus 4.5 and models numbered 4.6 and higher keep prior thinking in context and bill it as input, while Sonnet 4.5, Haiku 4.5 and earlier strip it. [OpenAI](https://developers.openai.com/api/docs/guides/reasoning) is likewise graceful: persisted reasoning is reusable only within one model family, and on a switch the API "omits incompatible reasoning from the model's context". Neither errors.

What replaying a completed turn actually costs is input tokens for blocks the provider may discard anyway, plus state that cannot survive a provider switch. That is why it is skipped. The rule Anthropic does enforce cuts the other way, and this PR follows it: a completed turn whose thinking we no longer hold now replays with no thinking block at all, where before `Gateway\Anthropic\Concerns\MapsMessages` rebuilt one from another provider's `reasoning_summary` — a block with no signature, which is exactly the "rebuilding the message" case Anthropic documents as a 400.

OpenAI's reasoning state follows the same rule. `replay_blocks` is now filled on OpenAI regardless of the `store` setting, so what is retained locally no longer depends on the request shape, and the step's raw `reasoning` item is the one home for `reasoning_id` and `reasoning_encrypted_content`. The store stops copying those fields onto every `steps[].tool_calls[]` entry; the `ToolCall` DTO keeps them for the in-run loop and the AG-UI round-trip, and rows written with the old copies still hydrate. A completed OpenAI turn therefore replays without its reasoning, as Anthropic turns already do. Gemini's `thought_signature` stays on the tool call, since #837 replays it across completed turns on purpose.

Replay state (`provider_content_blocks`, `provider_steps`) leaves the `meta` column, which goes back to holding response metadata such as the model, citations, and reasoning. `$message->tool_calls` and `$message->tool_results` still work on `ConversationMessage`, so existing UI code does not change.

The cost is a breaking change to the table. Existing installs need to run the backfill migration below once, which converts each stored row into steps and removes the two old columns.

> [!WARNING]
> Upgrade note for `UPGRADE.md`: `agent_conversation_messages` replaces `tool_calls` and `tool_results` with `steps`. Raw SQL against the old columns must move to `steps`. The backfill migration below ships in the `UPGRADE.md` entry rather than as a package migration, since only installs that ran the pre-1.0 schema need it; copy it into `database/migrations` and run it before deploying the upgraded application code. It gathers tool results across every row of a conversation, so a result that an older release recorded on the resume's row lands on the step that made the call. The same code is kept under `tests/Fixtures/Migrations/BackfillConversationSteps.php` and covered by `tests/Feature/Storage/ConversationStepsBackfillTest.php`.

> [!WARNING]
> Upgrade note for `UPGRADE.md`: `AgentResponse::pausedProviderContentBlocks()`, `pausedSteps()`, and their `StreamedAgentResponse` equivalents are removed. Use `$response->steps`. `ToolApprovalRequest::$providerContentBlocks` is replaced by `$steps`. `StoredMessage::$toolCalls` and `$toolResults` become `toolCalls()` and `toolResults()`. `StoredMessage::toArray()`, and so the pages returned by `paginateConversationMessages()`, emit `steps` in place of `tool_calls` and `tool_results`. `ConversationMessage` appends `tool_calls` and `tool_results` so serialized models keep both keys, though `tool_results` is now narrowed to `id`, `name`, `arguments`, `result`, `result_id`, `denied` and `failed`, so provider replay state such as `reasoning_encrypted_content` no longer reaches a serialized message. `steps[].tool_calls[]` no longer carries `reasoning_id`, `reasoning_summary` or `reasoning_encrypted_content` either; that state lives in the step's `replay_blocks`. A turn's reasoning also moves from `meta.reasoning` to `steps[].reasoning`, one entry per round-trip; UI code reading `meta['reasoning']` must read the steps instead.

<details>
<summary>Backfill migration</summary>

```php
<?php

use Illuminate\Database\Query\Builder;
use Illuminate\Database\Schema\Blueprint;
use Illuminate\Support\Collection;
use Illuminate\Support\Facades\DB;
use Illuminate\Support\Facades\Schema;
use Laravel\Ai\Migrations\AiMigration;

return new class extends AiMigration
{
    /**
     * Run the migrations.
     */
    public function up(): void
    {
        $table = config('ai.conversations.tables.messages', 'agent_conversation_messages');

        Schema::connection($this->getConnection())->table($table, function (Blueprint $blueprint) {
            $blueprint->longText('steps')->nullable();
        });

        $this->query($table)->where('role', 'user')->update(['steps' => '[]']);

        $this->query($table)
            ->select('conversation_id')
            ->distinct()
            ->orderBy('conversation_id')
            ->chunk(100, function (Collection $conversations) use ($table) {
                foreach ($conversations as $conversation) {
                    $this->backfillConversation($table, $conversation->conversation_id);
                }
            });

        Schema::connection($this->getConnection())->table($table, function (Blueprint $blueprint) {
            $blueprint->longText('steps')->nullable(false)->change();
            $blueprint->dropColumn(['tool_calls', 'tool_results']);
        });
    }

    /**
     * Rewrite every assistant row of one conversation as steps, each result landing on the invocation that made its call.
     */
    protected function backfillConversation(string $table, string $conversationId): void
    {
        $rows = $this->query($table)
            ->where('conversation_id', $conversationId)
            ->where('role', 'assistant')
            ->orderBy('id')
            ->get();

        // Results are gathered across the conversation because an approval resolved on a later request used to be recorded on that request's row...
        $results = [];
        $pending = [];

        foreach ($rows as $row) {
            foreach ($this->decoded($row->tool_results) as $result) {
                if (isset($result['id'])) {
                    $results[$result['id']] ??= $result;
                }
            }

            $pending = [...$pending, ...array_keys($this->decoded($row->approval_state)['pending'] ?? [])];
        }

        foreach ($rows as $row) {
            [$steps, $meta] = $this->stepsFrom($row);

            $steps = array_map(function (array $step) use ($results, $pending): array {
                $toolCalls = [];

                foreach ($step['tool_calls'] as $toolCall) {
                    $result = $results[$toolCall['id'] ?? ''] ?? null;

                    if ($result === null && ! in_array($toolCall['id'] ?? null, $pending, true)) {
                        continue;
                    }

                    $toolCalls[] = $result === null ? $toolCall : [
                        ...$toolCall,
                        'result' => $result['result'] ?? null,
                        ...array_filter(['denied' => $result['denied'] ?? false, 'failed' => $result['failed'] ?? false]),
                    ];
                }

                $step['tool_calls'] = $toolCalls;

                return $step;
            }, $steps);

            $this->query($table)->where('id', $row->id)->update([
                'steps' => json_encode($steps),
                'meta' => json_encode($meta),
            ]);
        }
    }

    /**
     * Split a flat assistant row into steps of unanswered tool calls, moving any replay and reasoning state out of its meta.
     *
     * @return array{0: list<array<string, mixed>>, 1: array<string, mixed>}
     */
    protected function stepsFrom(object $row): array
    {
        $meta = $this->decoded($row->meta);
        $calls = array_values($this->decoded($row->tool_calls));

        $providerSteps = $meta['provider_steps'] ?? null;

        if (is_array($providerSteps) && $providerSteps !== []) {
            $steps = [];

            foreach ($providerSteps as $providerStep) {
                $ids = $providerStep['tool_call_ids'] ?? [];

                $steps[] = [
                    'tool_calls' => array_values(array_filter($calls, fn (array $call) => in_array($call['id'] ?? null, $ids, true))),
                    'reasoning' => '',
                    'replay_blocks' => $providerStep['blocks'] ?? [],
                ];
            }
        } else {
            $steps = [[
                'tool_calls' => $calls,
                'reasoning' => '',
                'replay_blocks' => $meta['provider_content_blocks'] ?? [],
            ]];
        }

        $steps[array_key_last($steps)]['reasoning'] = (string) ($meta['reasoning'] ?? '');

        unset($meta['provider_steps'], $meta['provider_content_blocks'], $meta['reasoning']);

        return [$steps, $meta];
    }

    /**
     * @return array<string, mixed>
     */
    protected function decoded(?string $json): array
    {
        return is_array($decoded = json_decode($json ?? '', true)) ? $decoded : [];
    }

    protected function query(string $table): Builder
    {
        return DB::connection($this->getConnection())->table($table);
    }
};
```

</details>

Fixes #809. Fixes #402. Fixes the storage half of #639.


pushpak1300:
trackign this in #1035

PR #1011, opened 2026-09-17
Merge url citations by url and track every range they support

pushpak1300:
`UrlCitation` carried a single `startIndex`/`endIndex` pair, and the OpenRouter and xAI extractors pushed one object per annotation. A model that cites the same source in three places produced three `UrlCitation` objects for one URL, so `meta->citations` could not answer "which sources did this answer use?" without the caller deduplicating first.

```php
// annotations: example.com/one @ 0-10, example.com/two @ 11-25, example.com/one @ 26-40
$response->meta->citations;
// before: 3 citations, example.com/one listed twice
// after:  2 citations, example.com/one carrying [0-10] and [26-40]
```

`UrlCitation` gains a `ranges` collection and `addRange()`; `startIndex`/`endIndex` stay as they were and now report the first range, so existing callers are unaffected. `toArray()` and the `citation` stream event payload both expose `ranges`.

The dedupe-then-add-range step was already open-coded in the OpenAI extractor; it moves to `Gateway\\Concerns\\MergesCitations` and OpenAI, OpenRouter and xAI all use it. Providers whose APIs report spans per mention rather than per source all need it.

First of three — #1012 and #1013 build on `addRange()`.

Providers touched here are non-streaming only; streaming is #1012.


pushpak1300:
Closing — relaying every annotation as the provider sent it is strictly more information, and callers who want one row per source can groupBy('url') themselves.

PR #1012, opened 2026-09-17
Emit streamed citations once per url with every range they support

pushpak1300:
Builds on #1011.

OpenAI streams one `response.output_text.annotation.added` event per cited span, and OpenRouter repeats a source's annotation on every chunk that mentions it. Both gateways turned each of those into its own `Citation` event, so a stream and a non-streamed run over the identical response disagreed:

```php
// one source cited twice in the same answer
$stream->citations;               // before: 2 entries for the same url, 1 range each
$prompted->meta->citations;       // 1 entry, 2 ranges
```

`Citation::combine()` does not deduplicate, so `$stream->citations` inherited the duplicates verbatim.

Both gateways now accumulate citations while the stream runs and emit them once the response completes, going through `mergeCitation()` so each URL is emitted once with every range it supports. For OpenAI that means reading annotations off the completed response rather than the per-annotation events, which is the same data the non-streamed path parses. A test asserts the two paths now produce identical output for the same response body.

Citation events consequently arrive after `text.end` rather than interleaved with the deltas. They were never ordered relative to the text they annotate, and a range is only meaningful against the finished text, so nothing that consumed them could have relied on the earlier position.

`meta->citations` on a streamed `StepResponse` is deliberately left alone: `StreamedAgentResponse` overwrites it with `Citation::combine($events)`, so populating it in the gateway would have no observable effect.


pushpak1300:
Closing — relaying every annotation as the provider sent it is strictly more information, and callers who want one row per source can groupBy('url') themselves.

PR #1041, opened 2026-09-20
Fix: make conversation title generation opt-in and visible

AmadulHaque:
## Summery
RememberConversation::generateTitle() fires a paid title-generation LLM call for every new remembered conversation by default (config('ai.conversations.generate_title', true)).
- Key ai.conversations.generate_title absent from shipped **config/ai.php**, so publishing config never surfaces it.
- Failure path catches Throwable silently and falls back to truncated prompt text — broken provider shows as garbage titles, not an error.
Severity: Medium — unexpected cost + silent failure.

## Change

- `config/ai.php`: add documented conversations.generate_title section, default false via AI_CONVERSATIONS_GENERATE_TITLE.

## Behavior after
- No extra LLM call unless user explicitly enables AI_CONVERSATIONS_GENERATE_TITLE=true.
- Title failures still degrade gracefully but are now reported to logs / exception tracker.
- Disabled path unchanged: Str::limit($prompt, 50).


pushpak1300:
this is experimental feature that why we didn't document this.

PR #1042, opened 2026-09-20
Note that Boolean is TypeSafe's noul primitive

zoispag:
Docblocks only — no behaviour change.

While adopting the classification API, the one thing that cost me time was connecting `Classification\Boolean` to what the vendor documentation describes. TypeSafe only ever calls this primitive **`noul`** ([primitives/noul](https://docs.typesafe.ai/primitives/noul.md)), and their Claude skill points at those docs as the source of truth — but the string `noul` appears nowhere in this SDK outside `TypeSafeGateway`, where it is a wire detail. So reading the docs gives you no way to find the class, and reading the class gives you no way to know what you are sending.

To be clear, I think `Boolean` is the better name here — it reads like Laravel, and `noul` is jargon. This just adds the missing signpost between the two.

The second hunk records why `BooleanAnswer` has no `confidence` while `ChoiceAnswer` and `ScoreAnswer` do. That is correct per the spec — with two outcomes the single probability describes the distribution completely — but as written it reads like an omission, and it is the kind of thing that gets filed as a bug.

```diff
      * Create a new yes / no question whose answer is the probability of "true".
      *
+     * This is TypeSafe's "noul" primitive.
+     *
```

---

Unrelated to this change, and happy to open something separate if it is worth looking at: `Boolean::toArray()` emits `'type' => 'boolean'`, which `/systemone` does not accept — the API documents only `noul`, `choice` and `score`. That never fires for the three built-ins, since `TypeSafeGateway::mapQuestion()` maps them explicitly, but the same method falls through to `default => $question->toArray()`, so a custom `Question` implementation has its provider-neutral array sent as the wire payload. I could not tell whether the neutral shape is meant to be sendable or whether the fallback is just a convenience.


pushpak1300:
Thanks for your pull request!

PR #1049, opened 2026-09-21
Add a decide string macro for boolean classification

pushpak1300:
```php
Str::of($message)->decide('Is this spam?');

Str::decide($message, 'Is this spam?', criteria: ['true' => 'Unsolicited bulk mail.'], threshold: 0.9);
```


peterjthomson:
So classify() is the larger concept and the wider array or enum classifier (that works like switch or match). And decide() is the boolean one that works more like if() ? Awesome!

tpetry:
Just my opinion: Its missig a default value that is used when the request fails, there's a timeout etc. Currently you would have to gate the call with try-catch. A lot of boilerplate. But in many cases falling back to a default decision, like saying this is not spam, is a safe thing to do. Better than having many mails not delivered when the classificiation AI provider has an outage.

